PR #28, opened 2015-01-15
Make the tests runnable with Pkg.test()

ViralBShah:
I brought this up in #28, but started implementing myself. It is nice to be able to run the tests with Pkg.test(). Currently, this requires

mpirun -np <np> julia runtests.jl

Eventually, we can move to the MPIManager so that mpirun is also not necessary.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,0 +1,24 @@
+using Base.Test
+using MPI
+
+import MPI
+
+MPI.Init()
+
+comm = MPI.COMM_WORLD
+size = MPI.Comm_size(comm)
+rank = MPI.Comm_rank(comm)
+
+rank == 0 && println("Testing bcast...")
+include("test_bcast.jl")
+test_bcast()
+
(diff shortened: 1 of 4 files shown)


ViralBShah:
@lcw I suspect this does not work well with the tests that are part of build.jl. Would love to hear thoughts on how to streamline this, so that tests can be decoupled from the build.

lcw:
With this patch how do you run the tests on say 5 MPI processes?
Are you thinking something like:
```
mpirun -np 5 julia -e Pkg.test("MPI")
```

This patch currently kills the testing done in CMake so that will have to be updated.

ViralBShah:
I will update to fix the CMake tests too. I think this patch really has to use the MPIManager, so that tests can be carried out with just `Pkg.test()`.

jakebolewski:
This has been implemented a bit differently, I think it is best to test the bindings in isolation separate from the ClusterManager interface for now to wait on that maturing a bit.

PR #35, opened 2015-02-25
added Comm_split, Comm_split_type

steven-varga:
Hello,
an experienced c/c++ R software writer would like to add functionality to this great MPI implementation. This is 
my initial pull request; and it seems after commit, jlmpi_f2c.h is not updated.  see .gitignore.


amitmurthy:
cc @lcw

steven-varga:
Valid point; though the webpage  doesn't say the MPI version; and recent versions of openMPI, MPICH implements 3.0.  

The contribution is a reconnaissance action. I understand that it is a community work; and without the support of major contributors the small changes which I need for MPI 3.0 compliant work least likely will go through.  Nevertheless MPI3.0 does have useful features and I don't mind maintain my own branch until Julia MPI catches up.

PR #64, opened 2015-09-18
Fix deprecation warnings and tests.

andreasnoack:
This changes `runtests.jl` to use the full path to the test files. They often failed for me because of this.


lcw:
Thanks!

PR #83, opened 2015-10-15
@mpi_do prints remote exceptions if present

### LINKED ISSUE #82: RemoteExceptions when using MPIManager
With the new way of handling remote exceptions it can be difficult to know when you make an error while running code with `@mpi_do` because they are not printed. E.g.
```
julia> @mpi_do man error("")

julia> 
```
I'm not sure how to get around this, but it would be useful to see an error.
cc: @amitmurthy

amitmurthy:
Closes #82


andreasnoack:
Thanks Amit. This is really useful.

PR #84, opened 2015-10-21
Update montecarlo.jl

mcreel:
This fixes a crash which will occur if the number of replications per node is not an even multiple of the number of replications to accumulate before displaying intermediate results.

### DIFF
--- a/examples/montecarlo.jl
+++ b/examples/montecarlo.jl
@@ -24,7 +24,8 @@ function montecarlo(mc_eval::Function, mc_monitor::Function, reps, n_returns, po
     end
 
     if mod(pernode,pooled) != 0
-        pooled = div(pernode, commsize-1)
+        pooled = Int64(div(pernode, commsize-1))
+        contrib = zeros(pooled, n_returns)
     end    
 
     # workers' code


examples/montecarlo.jl
```diff
@@ -24,7 +24,8 @@ function montecarlo(mc_eval::Function, mc_monitor::Function, reps, n_returns, po
     end
 
     if mod(pernode,pooled) != 0
-        pooled = div(pernode, commsize-1)
+        pooled = Int64(div(pernode, commsize-1))
```
andreasnoack: Thanks. Is the `Int64` necessary here?

mcreel: Hi,
Looking at it again, I would say no, but I'm pretty sure that I added it
because in a usage case (a poorly chosen configuration) the argument was a
Float64. Perhaps I'm mistaken, though. At any rate, I don't think it will
do any harm.
Cheers,
M.

eschnett: At this point, `pernode` is a `Float64`, since it is obtained via a division `/`.

The code would be cleaner if the calculation of `pernode` was changed to explicitly round up and be integer.

Also, some of the arguments to `montecarlo` itself should probably be restricted to `Integer`.

mcreel: Hi Erik,
Rather than round pernode up, I prefer to warn the user that the number of
MPI ranks should be changed, so that the user will think about how to
spread the work in an optimal way. I agree that the types of the arguments
could be declared. Perhaps that would help users to understand the function
better, too. I can make this change, or feel free to do it yourself, too.
Cheers,
M.

eschnett: What I wanted to suggest is to calculate `pernode` differently so that its type remains integer. It's strange to calculate it as floating-point quantity.

I was just trying to figure out why the `Int64` conversion is necessary. Since it does seem necessary, and not through your changes, I don't want to request you to change things.

mcreel: What happens is that pernode is a Float64, because it is computed using /.
There very well may be some changes that would be worth making, I'm quite a
novice with Julia. Any suggestions are welcome.
Cheers,
M.

PR #85, opened 2015-10-21
Support for in place reduce and some async comms

leonardt:
Added reduce! which uses MPI_IN_PLACE to perform the reduction into sendbuf
Added ireduce and ibcast using Request


leonardt:
Looks like openmpi on travis doesn't support ibcast yet.

Probably best to separate the reduce! addition to a separate pull request and figure out an elegant way to have optional mpi 3 support

PR #90, opened 2015-11-12
enable precompilation

### LINKED ISSUE #63: Precompilation
I dont' know if this works

stevengj:
This fixes #63.

The MPI.jl module is small enough that precompilation doesn't help load times much.  However, it will make a much bigger difference for things like PETSc.jl, since you can't precompile PETSc unless MPI.jl can be precompiled.

cc: @JaredCrean2


stevengj:
Not sure why Coverage decreased, but I assume the 1.8% change is not significant.

jakebolewski:
Cool! Thanks.

PR #91, opened 2015-11-12
fix conflict between #89 and #90

stevengj:
This fixes a conflict between #89 and #90.


stevengj:
Thanks; you might think about tagging a release at some point soon.

PR #118, opened 2016-03-14
Try using trusty

andreasnoack:
The link to the MPICH 3 package is dead and I couldn't find a new for Precise so I decided to try out Travis' Trusty version.


andreasnoack:
All green here. I'll merge to get the status back to green.

PR #119, opened 2016-03-14
Try not to build OpenMPI on OS X.

andreasnoack:


andreasnoack:
The Mac builds work fine without building OpenMPI from source. This saves us more than 20 minutes on the test on OS X with OpenMPI. I'll merge and rebase the PR that fixes the MPICH build on Linux.

PR #130, opened 2016-04-13
Waitany fix

JaredCrean2:
I didn't expect to submit another PR so soon, but I discovered the `Waitany!`function is broken.  This fixes it and adds a test.


JaredCrean2:
Bump @eschnett

PR #131, opened 2016-04-13
Adding MPIBuffertype and allow for Ref{T}

jkozdon:
Allow the user to pass a Ref along with Ptr and Array to routines for instance
```
sreq = MPI.Isend(Ref(x, 4), cnt, dest, tag, comm)
```
which is safer than
```
sreq = MPI.Isend(pointer(x, 4), cnt, dest, tag, comm)
```

Since the buffer type was getting long (and possibly hard to maintain for future types), I aliased the buffer type as well.


lcw:
Thanks!

PR #136, opened 2016-04-23
Add tests for isend and some wait/test methods

kshyatt:


kshyatt:
Spectacular. The only tests that succeed are on OSX! I'll try to fix this soon.

kshyatt:
I found a bug in `Send` for one-element values that I fixed and added tests for too.

kshyatt:
I think for `Testany!` on MPICH we were missing a case where `flag` can be true, but the `index` can be `MPI_UNDEFINED`. Seems to be working locally on OSX with MPICH for me now.

kshyatt:
The test failure in `bcast` on nightly has nothing to do with this PR - @andreasnoack and @eschnett is this ok to merge?

andreasnoack:
Are you able to reproduce the test error locally? E.g. with `coverage=true` which causes `--inline=no`,

kshyatt:
@andreasnoack yes, the problem is that `Base.copy` can't copy the function `g` in the `bcast` tests.

andreasnoack:
Is it possible to fix that?

kshyatt:
It probably is, but I'm not certain how.

kshyatt:
Success! Ok to merge?

andreasnoack:
Yes. When lights are green.

kshyatt:
Sorry! The AV failure here was actually really important - I fixed a bad test! And I added one last one.

kshyatt:
3/4 AV tests passed (YAY) and the *nix tests are doing well too, so I'm going to merge.

PR #137, opened 2016-04-23
Fix nasty depwarn on nightly

kshyatt:


kshyatt:
This also only failed on AV, so I'm going to merge since it's been open for two days.

PR #142, opened 2016-04-29
Fix bug and add test for Cancel

kshyatt:
The Cancel function was broken in multiple ways. It's fixed now,
with a small test.


kshyatt:
Going to merge since this has been open for two days and passed CI.

PR #157, opened 2016-06-03
Alltoall methods with user given receiving buffer

MiroK:
Hi, this PR adds `Alltoall!` and `Alltoallv!` methods which allow the user to pass receiving buffer.


barche:
Sorry, seems we missed this, closed in favor of #232

PR #161, opened 2016-06-06
Add a note to README about Windows wrapping.

kshyatt:
Since this seems to be tripping many people up.


kshyatt:
@eschnett @davidanthoff @stevengj this look ok?

PR #170, opened 2016-08-12
Pass opfunc correctly to Reduce

tkonolige:
This fixes a bug where `opfunc` is incorrectly passed to `Reduce`.


andreasnoack:
Thanks. It would be great if you could also add a test.

andreasnoack:
@tkonolige Thanks again for this. It is easy to miss an update of a PR so another time, feel free to leave a comment when you have updated the PR.

PR #179, opened 2017-03-16
Update README.md to avoid mentioning Julia 0.4.

andreasnoack:
Enable testing on newer versions of Julia

Also update for 0.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ os:
   - osx
 julia:
   - 0.4
+  - 0.5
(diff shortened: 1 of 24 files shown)


andreasnoack:
@amitmurthy The termination of the MPI workers seems to go wrong somehow. See e.g. https://travis-ci.org/JuliaParallel/MPI.jl/jobs/211593017. It would be great if you could take a look. Please just go ahead and push to this branch if you come up with a fix.

PR #181, opened 2017-05-05
Update build script to fix Mac CI

jgoldfar:
These seemed to work in my tests

### DIFF
--- a/conf/travis-install-mpi.sh
+++ b/conf/travis-install-mpi.sh
@@ -12,7 +12,6 @@ case "$os" in
     Darwin)
         brew update
         brew upgrade cmake
-	    brew upgrade gcc
         case "$MPI_IMPL" in
             mpich|mpich3)
                 brew install mpich


jgoldfar:
By the way, is there a reason the README states to build OpenMPI on Mac from source, but CI doesn't since https://github.com/JuliaParallel/MPI.jl/commit/6b2950f6054fb6317f2099a3b749c4389e7ac7b5?

JaredCrean2:
I think the other people who manage this repo are occupied with other things, so I'm going to merge this.  Thanks for the contribution @jgoldfar

JaredCrean2:
This failed CI after merging.  Reverting until I can figure out why.

jgoldfar:
Unless I'm missing something, this PR only increased the passing tests; comparing https://travis-ci.org/JuliaParallel/MPI.jl/builds/235391591 (for another PR before this landed) and https://travis-ci.org/JuliaParallel/MPI.jl/builds/229235114 you can see that this package simply doesn't work on the Julia nightlies. Would perhaps be good to test explicitly against v0.5, v0.6, and allow failures on nightlies?

JaredCrean2:
The reason I reverted was this: https://travis-ci.org/JuliaParallel/MPI.jl/builds/236801435 .  It was unexpected that the v0.4 macOS tests failed.  I don't think your PR is the source of the problem and is probably a step towards solving it, but I want to find out more before merging.

I agree we should only be testing on older versions of Julia for the time being.

jgoldfar:
Ah, I see now. One build on OSX for v0.4 appears to have timed out because it had to build GCC from source, and the other because of an issue with the length of the directory Travis sets as a temporary directory. Interesting.

PR #182, opened 2017-05-23
Add Wtime and Wtick

JaredCrean2:
This PR adds the `MPI_Wtick` and `MPI_Wtick` functions.  One thing I did not add is the `MPI_WTIME_IS_GLOBAL` attribute, because 1) I don't know how to use pointers in Fortran, and 2) `MPI_Attr_get` was deprecated in favor for `MPI_Comm_Attr_get` in MPI 2.0, so we would have to add a system for dealing with that.


tbole:
Close? It's already in master.

JaredCrean2:
Right, good that these are available now.

PR #185, opened 2017-06-06
Fix `invoke` deprecation warning

Keno:


Keno:
Included as part of #187

PR #189, opened 2017-07-02
fix various depwarns on 0.7 (that break 0.6)

Sacha0:
This pull request fixes those depwarns under 0.7 that break 0.6. Best!

### DIFF
--- a/test/test_allgather.jl
+++ b/test/test_allgather.jl
@@ -10,7 +10,7 @@ end
 
 comm = MPI.COMM_WORLD
 
-for typ in (isdefined(:UnionAll) ? Base.uniontypes(MPI.MPIDatatype) : MPI.MPIDatatype.types)
+for typ in (@isdefined(UnionAll) ? Base.uniontypes(MPI.MPIDatatype) : MPI.MPIDatatype.types)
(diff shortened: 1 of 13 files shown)


Sacha0:
Updated CI here as well. Best!

vchuravy:
This looks like something that could be added to Compat.jl, so that it won't break v0.6?

Sacha0:
> This looks like something that could be added to Compat.jl, so that it won't break v0.6?

Good call. That addition should be straightforward. Best!

test/test_allgather.jl
```diff
@@ -10,7 +10,7 @@ end
 
 comm = MPI.COMM_WORLD
 
-for typ in (isdefined(:UnionAll) ? Base.uniontypes(MPI.MPIDatatype) : MPI.MPIDatatype.types)
```
Keno: Should just be `Base.uniontypes(MPI.MPIDatatype)`, since we've dropped pre 0.6 support.

andreasnoack:
I don't think this is relevant anymore

Sacha0:
Indeed! Thanks Andreas! :)

PR #195, opened 2017-08-31
Enable override of MPI_Comm_c2f detection

rcthomas:
To build using Cray compiler wrappers (`CC=cc` and `FC=ftn`) I find I have to override this check.  The problem is that the associated check doesn't work because it lacks the necessary `#include "mpi.h"` and furthermore, the Cray header defines this via macro:

    #define MPI_Comm_c2f(comm) (MPI_Fint)(comm)

Maybe there's a better check that could be used?  If so then ignore this PR but the right one has to be put in (not sure my CMake is good enough...)


rcthomas:
Actually I think this just doesn't fix the problem unfortunately; trying to use this function from an application results in an error because the function isn't defined in libjuliampi.

PR #198, opened 2017-11-17
Add wrappers for MPI_Info related routines, MPI_COMM_SPLIT and timers

tbole:
I'm planning to use MPIO for HDF5, which requires MPI related stuff, first communicators and second `MPI_Info`, for defining custom file system properties over in HDF5.jl. I have thus added some new wrappers:

`MPI_COMM_SPLIT` and `MPI_COMM_SPLIT_TYPE` for defining new communicators. Especially `MPI_COMM_SPLIT_TYPE` may be useful, since it eases creating node-level communicators. 
Second `MPI_Info` related functions (create,get,set,delete,finalize) have been added, since these are partially required by above routines. Two more convenience wrappers are for Wtime and Wtick.


tbole:
Those failures in `MPI_Info_get` are rather weird and are related to the MPI implementation. 
While OpenMPI and somewhat recent mpich versions work, the ancient 2013 mpich at Travis and MS MPI fail, due to a missing character and some MPI internal quirks between C/Fortran termination of strings.  Must admit that the passage in the MPI standard about null terminated strings is unclear at least. This error had been fixed in mpich [some years ago](https://trac.mpich.org/projects/mpich/ticket/2225) 3 years ago. I haven't tested Windows MPI though, but since MS MPI is pretty much following mpich and appveyor uses MS MPI v7 from 2015, it's probably the same bug. The most recent version is [8.1](https://msdn.microsoft.com/en-us/library/mt147726(v=vs.85).aspx) (June 2017). 

For my own projects I typically tend to dump backward compatibility for old fixed errors. Is there any preferred way of dealing with such situations for the Julia packages?

tbole:
Bump. Since there has been no feedback at all on this PR yet for quite a while, it would be great to know when / if this would land. There's another pull request of mine over at HDF5 implementing MPIO, which NEEDS  MPI_Info types to work. This PR is pretty much a road block for that feature.
Plus, the CI errors noted above seem to be fixed by #202 .

barche:
Maybe #202 can get merged first to get the CI running again, then this one and then I can rebase #201 onto this?

tbole:
Just rebased this to master and added some fixes for 0.7. Now all the 0.6 tests succeed and 0.7 at least starts up, but then segfaults in each test. Since this is unrelated to the changes here (maybe even a bug in 0.7?), I would propose to merge this branch so #201 can be rebased. 
After that it would be great to run femtocleaner to get rid of the ampersand/Ref deprecations.

PR #199, opened 2017-11-21
Add support for SubArrays

tbole:
Since using SubArrays has become quite comfortable due to the views macro and seems to be encouraged by many Julia devs, it would be great, if they would be directly supported by MPI (i.e. without creating a reference to the subarray first)

    @views begin
        sreq = MPI.Isend(arr[range1], dest, tag, comm)`
        rreq = MPI.Irecv!(arr[range2], src, tag, comm)`
    end

This would actually only require adding SubArray to MPIBuffertype.
  `MPIBuffertype{T} = Union{Ptr{T}, Array{T}, SubArray{T}, Ref{T}}`

However SubArrays can contain non-contiguous memory blocks, whereas MPI assumes contiguous memory. I wanted to have routines providing convenience interfaces for Array to also support SubArrays. The checks whether a SubArray is contiguous are placed only in these convenience functions, which I intended to be "safe".
The more stringent version would probably be to implement the checks in the routines using MPIBufferType directly. One could also think of avoiding the checks at all and leaving this up to the user.

### DIFF
--- a/src/mpi-base.jl
+++ b/src/mpi-base.jl
@@ -2,7 +2,7 @@ const MPIDatatype = Union{Char,
                             Int8, UInt8, Int16, UInt16, Int32, UInt32, Int64,
                             UInt64,
                             Float32, Float64, Complex64, Complex128}
-MPIBuffertype{T} = Union{Ptr{T}, Array{T}, Ref{T}}
+MPIBuffertype{T} = Union{Ptr{T}, Array{T}, SubArray{T}, Ref{T}}
 
 fieldoffsets(::Type{T}) where {T} = Int[fieldoffset(T, i) for i in 1:nfields(T)]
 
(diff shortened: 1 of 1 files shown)


tbole:
Bump... is there any interest in this feature?
It is rather helpful for array based codes and I've actually been using it for some time.

JaredCrean2:
Yes, I think this is a good feature.  My only question is whether we should do something more general like adding `AbstractArray` to `MPIBufferType` and adding the appropriate `iscontiguous` checks at the innermost function (just before the `ccall`).  Of course, that leads to the problem of using `MPI_Datatype` to support non-contiguous arrays...

Anyways, the Mac travis tests on 0.6 failed.  Could you try rebasing on master and see if that fixes it?  Once CI passes, I'll merge this PR if no one objects.

tbole:
I've actually already tried adding `AbstractArrays` to `MPIBufferType`, but noticed that `iscontiguous` is only implemented for `SubArrays`. We would thus have to explicitly check the for the type of the Array, so I found dispatching on `SubArray` to be more simple. We could also pack this into generated function, though I'm not too experienced with these.

After rebasing all 0.6 checks have succeeded.

JaredCrean2:
Ok, lets keep this in its current form the, I'll everyone until tomorrow evening to comment, and then merge if there are no objections.

PR #200, opened 2018-01-26
Patch to allow package to build & load on v0.6+

jgoldfar:


lcw:
Thanks, have you seen the work here https://github.com/JuliaParallel/MPI.jl/pull/210?  I am thinking about merging it.  Any comments you have on that patch set would be great.

jgoldfar:
I'll play around with the branch, but as it correctly builds and passes the tests one might expect it to, it definitely supersedes this one. Very impressive work! I'll be happy to contribute based on that.

PR #201, opened 2018-02-08
Add one-sided communcation calls

barche:
This adds some of the MPI window-based (RMA) communication primitives, as well as `MPI_COMM_SPAWN`.

PR #202, opened 2018-02-13
Upgrade linux mpich and open-mpi versions

barche:
As discussed in PR #201. I will also add the cman test fixes here if the problems are confirmed in the CI output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
   - MPI_IMPL=openmpi
 before_install:
   - sh ./conf/travis-install-mpi.sh $MPI_IMPL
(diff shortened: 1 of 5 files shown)


barche:
Looks like this passes Travis on 0.6, so it's ready as far as I'm concerned. I plan to fix 0.7 issues after my other changes are merged, so I can update everything in one go. And then maybe later also the windows CI, it just seems to time out on 0.6.

tbole:
This will also fix some CI errors in #198. I think having somewhat recent MPI versions for CI is a good idea, since those are the ones used on most clusters. In case of OpenMPI we should stick to either the current version or those [still in support](https://www.open-mpi.org/software/ompi/v3.0/).

PR #203, opened 2018-02-24
Add WindowIO and MPIWindowIOManager

barche:
The objective of this PR is to get some better cooperation between the Julia parallel processing system and MPI.

* `WindowIO` allows point-to-point communication between MPI processes using a Julia `IO`. Brief excerpt from the test:

```julia
using MPI

MPI.Init()
comm = MPI.COMM_WORLD
const rank = MPI.Comm_rank(comm)
const N = MPI.Comm_size(comm)

const winio = WindowIO(comm) # read from anyone
const writer = WindowWriter(winio, 0) # writes to rank 0
if rank != 0
    write(writer, rank)
    flush(writer) # Must flush to trigger communication
else
    result = read(winio, (N-1)*sizeof(Int)) # Blocks until all required data is read
end
```

* `MPIWindowIOManager` sets up a cluster manager to use the `WindowIO` buffer as communication layer. It adds two modes for `start_main_loop`:
  - `MPI_WINDOW_IO`: Normal cluster manager, with the workers only waiting on the process_events loop and all commands only executed on master
  - `MPI_WINDOW_NOWAIT`: The cluster manager is initialized, but workers don't process events and instead run in SPMD MPI fashion as usual. If a Julia parallel call is needed, it must be prefixed with the `@cluster` macro to activate the event loops on the workers. All code in the `@cluster` block then only runs on the master and can contain normal Julia parallel programming calls. After the block ends, SPMD mode is resumed.


lcw:
I don't know much about the cluster manager but WindowIO seems cool to me.

PR #205, opened 2018-03-08
Add omitted windows MPI constants

barche:
See discussion in PR #201


musm:
Thanks but it doesn't look like the type `CWin ` is defined anywhere

barche:
I added it, on my Windows machine it precompiles now, but there are still failing tests.

musm:
So where in the code is the dll for windows loaded? I noticed the the MS-MPI doesn't install the dll only lib files

barche:
You need to install the runtime, which is the .exe installer. The .msi installer only installs the SDK files for compiling MPI programs with Visual Studio.

I just committed an update, but there are still some errors that I don't know how to fix, I'm not sure getting a fully working Windows MPI.jl is an easy task, and I actually have 0 use for it myself.

barche:
Maybe this should be merged? Even if not all tests pass, at least it allows the package to precompile on Windows and the pre-onesided tests appear to work. Also, I have another PR I'd like to submit for adding `MPI_ACCUMULATE` and `MPI_GET_ACCUMULATE`.

musm:
sounds good

musm:
Bump

barche:
I have disabled the broken tests on Windows, can this be merged if it passes 0.6 CI?

barche:
Again a mystery why this happens, but disabling the spawn test on win32 helps. I also fixed a regression with macOS MPICH due to a newer CMake version.

PR #206, opened 2018-04-19
Add onesided accumulate functions

barche:
Adds two more one-sided comms functions, which I need for an experimental array library.


barche:
Any objections to merging this? It passes 0.6 CI, and I'm getting ready to announce https://github.com/barche/MPIArrays.jl (which depends on this PR)

PR #208, opened 2018-05-23
Optionally call MPI.Finalize at exit

barche:
This is useful for enabling cleanup using standard Julia finalizers, which otherwise run after `MPI.Finalize()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,12 @@ able to run the MPI job as expected, e.g., with
 
 `mpirun -np 3 julia 01-hello.jl`
 
+### Cleanup
+In Julia code building on this package, it may happen that you want to run MPI cleanup functions in a finalizer.
+This makes it impossible to manually call `MPI.Finalize()`, since the Julia finalizers may run after this call.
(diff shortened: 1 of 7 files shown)


barche:
Looks like this needs more work, judging by the spectacular CI fail :)

barche:
It passes the 0.6 CI now, any objections to merging this?

lcw:
No, seems useful to me.

tbole:
Looks like a good idea to me. But I'm wondering, why the finalizer needs to be a C function. 
Shouldn't this work in Julia just as well, or am I missing a point?

JaredCrean2:
@tbole The problem is ordering: julia `atexit` hooks, then finalizers, then the C `atexit` hooks.  If a julia `atexit` hook is used to finalize MPI, then no object can use MPI functions as part of its finalizer.

I don't think this repository has a dedicated maintainer at the moment, so I'll merge this tomorrow if there are no objections.

tbole:
Ah ok, it's about other functions usingMPI during finalization. Then using C makes sense.

PR #209, opened 2018-05-28
fix CComm depwarn

JaredCrean2:
Changes `bitstype` to `primitive type`.


JaredCrean2:
Oops, I wasn't on the latest master, where this issue is already fixed.

PR #210, opened 2018-06-01
V07

jkozdon:
Trying to set things up so that MPI.jl can work with v0.7.

Locally things are working file. And I believe things are still compatible with v0.6

I haven't fixed all the dep warnings, but those that popped when running (not the initial build)


jkozdon:
Note: I had to disable `test_cman_*` when rulling with Julia 0.7. I could not get these to run properly. There are also some depreciation warnings to `cman.jl` but I could not get these to be resolved and keep Julia v0.6 OK.

Since I don't use the Julia Distributed stuff I'm hoping that someone else can fix this.

There is a problem with the Windows build with Latest and v0.7

jkozdon:
I was able to fix the deprecation warnings on the tests starting (cman and windows still do not work).

barche:
Thanks for making the effort to do this!

I get the following error when doing `using MPI` on Julia 0.6:

```
INFO: Recompiling stale cache file /home/username/.julia/lib/v0.6/MPI.ji for module MPI.
ERROR: LoadError: LoadError: LoadError: UndefVarError: undef not defined
Stacktrace:
 [1] include_from_node1(::String) at ./loading.jl:576
 [2] include(::String) at ./sysimg.jl:14
 [3] include_from_node1(::String) at ./loading.jl:576
 [4] include(::String) at ./sysimg.jl:14
 [5] include_from_node1(::String) at ./loading.jl:576
 [6] include(::String) at ./sysimg.jl:14
 [7] anonymous at ./<missing>:2
while loading /home/username/.julia/v0.6/MPI/src/mpi-op.jl, in expression starting on line 21
while loading /home/username/.julia/v0.6/MPI/src/mpi-base.jl, in expression starting on line 704
while loading /home/username/.julia/v0.6/MPI/src/MPI.jl, in expression starting on line 20
ERROR: Failed to precompile MPI to /home/username/.julia/lib/v0.6/MPI.ji.
Stacktrace:
 [1] compilecache(::String) at ./loading.jl:710
 [2] _require(::Symbol) at ./loading.jl:463
 [3] require(::Symbol) at ./loading.jl:405
```

Also, when wrapping this up it's probably best to squash it into fewer commits if possible.

barche:
Never mind the comment about `undef`, it was because of an old Compat version.

tbole:
Great work! Hope to have this in master rather earlier than later!

jkozdon:
I think that I got most of the windows problems fixed. I can combine the commits and push out again if we're ready to merge.

Two known problems (neither of which affect me, so hard to keep pushing them):

- Using Julia Distributed stuff does not work (i.e., cman)
- Windows 32 with nightly and latest is crashing on the test_allreduce when using a non-MPI in reduction type.

Perhaps best to merge this into master and then setting up issue tickets?

lcw:
Thanks Jeremy!  Your plan of action sounds good to me.  Is it reasonable to squash some of the commits as @barche suggested?  Either way I would like to merge this work in the next couple of days unless someone objects.

KristofferC:
I should say that by default when a Manifest is checked in, the package manager will test the package at the configuration given by the Manifest. This may not be the same as the one that users install (since new, compatible releases of dependencies have been registered).

I have an open PR https://github.com/JuliaLang/Pkg.jl/pull/290 that will likely soon be finished that would also allow an option to `Pkg.test` that would ignore the Manifest. For now, perhaps it is best to wait a bit with checking in the manifest.

jkozdon:
I've removed the Manifest and squashed things down to a few logical commits.

PR #211, opened 2018-08-02
V07 further fixes

barche:
This fixes the clustermanager and some other v0.7 deprecations. I also upped the required Julia version to v0.7.0-beta, because not all fixes properly use compat. I propose that once this is release-ready, we move MPI.jl version 0.7, which will then no longer work on Julia 0.6. That would also allow cutting a lot of the existing compat stuff out, and 0.6 users could just use the previous release.


.travis.yml
```diff
@@ -5,8 +5,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
-  - 0.7
```
vchuravy: We should still have `0.7` here

appveyor.yml
```diff
@@ -1,11 +1,7 @@
 environment:
   matrix:
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7.0-alpha-win32.exe"
```
vchuravy: this should be `"https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7-latest-win32.exe"`

barche:
OK, I think it's done like this!

lcw:
Looks good to me.  Thanks!

PR #214, opened 2018-08-20
Support for precompiling modules based on MPI.jl

barche:
This turns the symbol names into references, and fills in the handles at `__init__` time. Without this, there were precompilation warnings when precompiling a module using MPI.jl on Julia 1.0.

### DIFF
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -23,20 +23,13 @@ include("cman.jl")
 const mpitype_dict = Dict{DataType, Cint}()
 const mpitype_dict_inverse = Dict{Cint, DataType}()
 
-function __init__()
-    @static if Compat.Sys.isunix()
-        # need to open libmpi with RTLD_GLOBAL flag for Linux, before
-        # any ccall cannot use RTLD_DEEPBIND; this leads to segfaults
-        # at least on Ubuntu 15.10
-        @eval const libmpi_handle =
-            Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
-
-        # look up all symbols ahead of time
-        for (jname, fname) in _mpi_functions
-            Core.eval(MPI, :(const $jname = Libdl.dlsym(libmpi_handle, $fname)))
-        end
+@static if Compat.Sys.isunix()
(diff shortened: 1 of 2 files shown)


lcw:
I am not sure if I understand what you mean about using ccall as it was intended.  We are using CMake to determine the Fortran name mangling and storing that in `_mpi_functions` so that we can call the right symbols.

barche:
I found the origin of this, which led to some discussion on the commit:
https://github.com/JuliaParallel/MPI.jl/commit/bb04bf843954cef63ba2775628117165fff1d255

Let's not merge this yet, as Windows still fails and going back to (symbol,library) tuples may be a better idea after all. I don't think this is still an issue, maybe we should ping one of the core devs on this?

vchuravy:
The tuple get's resolved at compile time (which it why it has to be constant values), also see src/ccall.cpp  in Julia.

```
using Libdl
# Static
f(x) = ccall(:sinf, Float32, (Float32, ), x)
f1(x) = ccall((:sinf, Base.libm_name), Float32, (Float32, ), x)

const ptr = Libdl.dlsym(Libdl.dlopen(Base.libm_name), :sinf)
g(x) = ccall(ptr, Float32, (Float32,), x)

# Dynamic
h(x) = ccall(Libdl.dlsym(Libdl.dlopen(Base.libm_name), :sinf), Float32, (Float32,), x)

# With extra pointer load
const rptr = Ref(Libdl.dlsym(Libdl.dlopen(Base.libm_name), :sinf))
s(x) = ccall(rptr[], Float32, (Float32,), x)
```

The performance of all of these is similar (except the fully dynamic one), but `s` does have an extra pointer load associated with it.

I don't know how bad the mangling is, but if it is predictable we could do it as part of a macro.

lcw:
As far as the name mangling goes, it really depends on which fortran compilers we want to support.  It sometimes adds an underscore, sometimes two; sometime the symbols are lowercase, sometimes uppercase.  It just depends.

In the long run moving from the Fortran ABI and writing a fixed ABI C wrapper library may be the way to go as discussed here <https://github.com/JuliaParallel/MPI.jl/issues/59>.  [MorphMPI](http://morphmpi.sourceforge.net/) is an example of such a library.

PR #221, opened 2018-11-02
Add Win_allocate_shared and Win_shared_query

JobJob:
Enables creating shared memory arrays using MPI. See tests for usage example.


JobJob:
Re Win32 test errors: it wasn't me... 
It says it was test_allreduce.
All other builds are passing.

JobJob:
Thanks folks!
> Out of interest, what is your use case for this?

I'm doing some distributed optimisation: on each node I have a master updating parameters in shared memory for workers, who then evaluate a fitness function using those parameters.

I'm hoping to release a package at some point with some tools to make this easier (pointers to related packages appreciated). While we're here: does anyone know if taking a view of an `Array{Float64,N}` is threadsafe? I can imagine it modifying some state to avoid the source array being garbage collected, and that that possibly could be an issue.

barche:
Well, you might want to look at my [MPIArrays.jl](https://github.com/barche/MPIArrays.jl) package, I gave a talk about it at JuliaCon,but I haven't gotten round yet to releasing it officially. Seems like using `Win_allocate_shared` would also be useful there, for intra-node communication.

JobJob:
> Well, you might want to look at my MPIArrays.jl package, I gave a talk about it at JuliaCon...

Looks cool. I'll have to watch the talk.

> Seems like using Win_allocate_shared would also be useful there, for intra-node communication.

Yeah might well be.

PR #223, opened 2018-11-26
Fix primitive type

JaredCrean2:
Previously, the code in `mpitype` would create a type with zero size for primitive types, because the number of fields is zero.  This caused problems for sending and receiving messages.  The new code attempts to find an existing type with the same number of bytes and use that `MPI_Datatype`, and throws an error otherwise


JaredCrean2:
While merging with `master`, I noticed the `type_create` function which appears to be an unused copy of `mpitype`.  I commented it out and, if the CI tests agree, I will remove it.

JaredCrean2:
The 32 bit AppVeyor failed, but the failure looks unrelated to any code I modified.  64 bit passed, and the merge commit Travis passed.  I think this is good to merge (and would like to merge it quickly, because it contains a bugfix).

barche:
Would it be possible to make the fix completely general by using `MPI_TYPE_CONTIGUOUS` to create a new type composed of 8-bit wide blocks, say something like 3 `MPI_INTEGER1` to represent a 24-bits primitive type? I never used custom MPI types, so I don't know if this will work as expected.

barche:
Just added a commit that tests the `MPI_TYPE_CONTIGUOUS`approach.

JaredCrean2:
I'm a bit concerned creating datatypes composed of single bytes might be slow (although I don't have evidence for this), so I modified the scheme to use `TYPE_CREATE_STRUCT` for primitive types, breaking the primitive type into integer sized pieces.  It turns out doing the factorization is easy because the integer sizes are all multiples of each other.

barche:
Looks fine to me, I have no idea if there is a performance difference either, but since it's just an implementation detail it can always be changed later, should the need arise.

JaredCrean2:
Cool, merging.

PR #224, opened 2018-12-30
Add an explanation about compilers' detection

ziotom78:
The rules followed by CMake to find the path to compilers are not immediate. This PR adds a few hints in the README.


lcw:
Thanks for the note.  I don't know why CMake is not using the PATH.

barche:
Thanks for the PR, and sorry for the late reaction. I think CMake uses the same set of prefixes as it uses to find libraries, so that might be why it ignores the path. Setting `CC` and the others is the recommended way to find the compiler:
https://cmake.org/cmake/help/latest/envvar/CC.html

For MPI, I usually set this to the MPI wrapper compilers.

Issue #242, opened 2019-04-24
Windows MPI 10.0 now available


hearnsj:
MS-MPI version 10.0 dated October 2018 is now available. Perhaps this link should be in the README
https://www.microsoft.com/en-us/download/details.aspx?id=57467

I can test this version on Windows 10

simonbyrne:
@hearnsj: thanks, I have added the changes to #241 and #243.

Do you know if it is necessary to add MPI to the PATH variable on Windows? I noticed it is in the Appveyor script, but not mentioned in the docs.

simonbyrne:
Ah, I think the PATH is required just to find `mpiexec`.

This issue can be closed (fixed in #241 and #243).

PR #243, opened 2019-04-24
Update CI scripts

simonbyrne:
Run on 1.0 and latest release, use standard Julia Appveyor script and latest Windows MPI (#242).


simonbyrne:
I'm guessing 32-bit windows is failing because the Windows MPI is 64-bit? Should we just disable the CI for that case?

lcw:
I'm okay with disabling the test.  Although, I don't use Windows MPI.

simonbyrne:
Okay, the 32-bit Windows failure has something to do with this line giving a `Char` when it should be an `Int32` (I think):
https://github.com/JuliaParallel/MPI.jl/blob/bc2e97949aae507414c96c4596961f0d05abc3e5/src/mpi-op.jl#L27

simonbyrne:
That is probably a separate issue. I'll just allow failures for now can figure it out later.

Happy for this to be merged when ready.

simonbyrne:
Alright, now there's a new issue in `test_reduce.jl`:
https://ci.appveyor.com/project/andreasnoack/mpi-jl/builds/24080532/job/qtoh8m4wk64syh3o#L99

lcw:
At this point I saw we merge if all CI passes except 32-bit Windows.

simonbyrne:
I've traced the 32-bit Windows issue to calling `MPI.Reduce!` with user-defined functions. Can't figure out the exact problem though.

Issue #244, opened 2019-04-24
One sided test fails on MacOS with OpenMPI


simonbyrne:
This is the cause of the current Travis failure.

```
├ mpiexec --oversubscribe -n 2 julia --project test/test_onesided.jl
[Indefatigable:78930] *** An error occurred in MPI_Get
[Indefatigable:78930] *** reported by process [3912892417,1]
[Indefatigable:78930] *** on win rdma window 3
[Indefatigable:78930] *** MPI_ERR_RMA_RANGE: invalid RMA address range
[Indefatigable:78930] *** MPI_ERRORS_ARE_FATAL (processes in this win will now abort,
[Indefatigable:78930] ***    and potentially your MPI job)
[Indefatigable.local:78928] PMIX ERROR: UNREACHABLE in file server/pmix_server.c at line 2079

signal (15): Terminated: 15
in expression starting at /Users/simon/.julia/dev/MPI/test/test_onesided.jl:19
__psynch_cvwait at /usr/lib/system/libsystem_kernel.dylib (unknown line)
unknown function (ip: 0xffffffffffffffff)
Allocations: 2342120 (Pool: 2341825; Big: 295); GC: 4
[Indefatigable.local:78928] 1 more process has sent help message help-mpi-errors.txt / mpi_errors_are_fatal
[Indefatigable.local:78928] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

This is the offending line:
https://github.com/JuliaParallel/MPI.jl/blob/bfc5bcea6932179f64d0aac2f9d0ef92f3d5a87a/test/test_onesided.jl#L19

simonbyrne:
actually seeing the same on a Linux machine:
```
├ mpiexec --version
mpiexec (OpenRTE) 3.1.2

Report bugs to http://www.open-mpi.org/community/help/
```

simonbyrne:
@barche is this related to your issue here: https://github.com/open-mpi/ompi/issues/6374?

simonbyrne:
indeed it was!

simonbyrne:
Okay, that is fixed on #243, but I'm seeing the following intermittently:
```
--------------------------------------------------------------------------
A system call failed during shared memory initialization that should
not have.  It is likely that your MPI job will now either abort or
experience performance degradation.
  Local host:  Traviss-Mac-80.local
[...]

PR #245, opened 2019-04-24
Work around a type issue on Windows 32 bit

lcw:
As @simonbyrne pointed out, it seems that
https://github.com/JuliaParallel/MPI.jl/blob/bc2e97949aae507414c96c4596961f0d05abc3e5/src/mpi-op.jl#L27
returns a `Char` on Windows 32 bit when it should (?) be an `Int32`.

To avoid this we make sure the inverse map doesn't contain the value `Char`.

### DIFF
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -61,7 +61,7 @@ function __init__()
 
     # Note: older versions of OpenMPI (e.g. the version on Travis) do not
     # define MPI_CHAR and MPI_*INT*_T for Fortran, so we don't use them (yet).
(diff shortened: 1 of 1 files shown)


simonbyrne:
Ah, that makes sense.

lcw:
Unfortunately it seems the build is still failing.

simonbyrne:
I think that's the same underlying problem causing #244, that is fixed on #243.

simonbyrne:
i'll just pull the #243 branch.

lcw:
sounds good

PR #248, opened 2019-04-25
cleanup .travis

vchuravy:
just a small refactor that makes the test a bit by a minute or so since we only run Coverage once.


simonbyrne:
I'm not sure this will work: the Travis docs say that there isn't any data persistence between jobs:
https://docs.travis-ci.com/user/build-stages/#data-persistence-between-stages-and-jobs
so we would lose the coverage info.

vchuravy:
ah merde.

simonbyrne:
It seems like a large chunk of the time is building MPI, so we could see if turning on caching helps:
https://docs.travis-ci.com/user/caching/

PR #253, opened 2019-05-02
remove unused part from CMake file

simonbyrne:
I still don't really understand how CMake works, but this doesn't appear to be used.


lcw:
Not sure why that is still around, thanks!

PR #255, opened 2019-05-03
Add reference counter for finalization

simonbyrne:
This is largely based on the approach in the LibGit2 standard library. This avoids the need to build a separate C library, and will work on Windows. Also it removes the need for `finalize_atexit` altogether (since it is be enabled by `Init()`).

TODO:
- [x] remove the C library. Since this is also used for "stepping though" to the actual MPI library, we will need to figure out a way to avoid that.
- [x] decide what to do with `MPI.Finalize()`: as it is, a user could call this before an object has been garbage collected, which would be invalid when it finalizers. This hasn't been an issue so far since currently it only affects `MPI.Info` objects, but if we are to add I/O functionality, then it could become a bigger  problem.
  - The simplest option I can think of is to rename the current `Finalize()` to `_Finalize()` to discourage its external use. We could then either deprecate `MPI.Finalize()` or simply have it call `refcount_dec()`.


vchuravy:
In the CUDA world we delay finalization until all refcounts are zero.

lcw:
This looks good to me!  I like the idea of `MPI.Finalize()` calling `refcount_dec()` so that simple examples mimic the C and Fortran counterparts.  The other way we could go is remove the explicit calling of `Init()` and `Finalize()` and have the library take care of both similar to [mpi4py](https://bitbucket.org/mpi4py/mpi4py/src/master/demo/helloworld.py).

simonbyrne:
> The other way we could go is remove the explicit calling of Init() and Finalize() and have the library take care of both similar to mpi4py.

I'm generally in favour of that: the only question is if there are packages which use MPI optionally, and so may `using MPI` but not want to initialise it?

JaredCrean2:
What happens in the case where some other library has initialized MPI?  By convention, the library that initializes MPI should also finalize it.  My initial reading is this implementation could finalize MPI even if we didn't initialize it.

JaredCrean2:
I will do a more thorough review tomorrow, but one concern I have whenever I see julia's `atexit` function is that it runs before finalizers (see this [comment](https://github.com/JuliaParallel/MPI.jl/pull/208#issuecomment-394507775)).  C's `atexit` function runs afterward, which was the motivation for the existing implementation.



> This avoids the need to build a separate C library,

We have to build a small library to get the values of MPI constants anyways, so I'm not sure this is a significant benefit.  The difference between having to build 0 and 1 C libraries is large, the difference between 1 and 2 is small.

simonbyrne:
> What happens in the case where some other library has initialized MPI? By convention, the library that initializes MPI should also finalize it. My initial reading is this implementation could finalize MPI even if we didn't initialize it.

Good point. 2b3903d should fix that issue.

> We have to build a small library to get the values of MPI constants anyways, so I'm not sure this is a significant benefit. The difference between having to build 0 and 1 C libraries is large, the difference between 1 and 2 is small.

True, that wasn't the motivation. The main benefit is that it works on Windows, and so can be enabled automatically, since calling `MPI.Finalize()` is currently unsafe.

JaredCrean2:
Just out of curiosity, how hard would it be to get the shared library to build on Windows (I don't develop on Windows, so I have no idea).  If we could get it to work, it would enable the removal of the `win_mpiconstants.jl` file and unify how we treat Linux/Mac and Windows (by generating the MPI constants the way we do on Linux/Mac).  It would also enable us to keep the current `atexit` approach, which is less invasive than reference counting.

If building the shared library on Windows is going to be particularly difficult, then I agree with the reference counting approach, and this implementation looks good to me.

Regarding automatically initializing MPI, I am opposed because eventually we may want to expose `MPI_Init_thread`, which requires an argument from the user, so it cannot be called automatically.

simonbyrne:
> Just out of curiosity, how hard would it be to get the shared library to build on Windows (I don't develop on Windows, so I have no idea).

It might be possible via BinaryBuilder: the main challenge would be figuring out how to link to Microsoft MPI.

lcw:
I agree with the points raised about why we might want leave the `Init()` call explicit.  As far as I am concerned I think this patch set is ready. @barche Have you had a chance to look at these changes?

JaredCrean2:
Why is BinaryBuilder related?  Isn't the main question whether CMake can build and link a shared library to the Microsoft MPI library (regardless of how Microsoft MPI came into existence)?

simonbyrne:
It's the easiest way to get compatible a binary built on Windows, which is notoriously tricky to do, and avoids the user needing to install a compiler.

JaredCrean2:
Oh, if we can't assume Windows has a working compiler then we can't treat it the same was as Linux at all.  I guess reference counting is the way to go.

PR #256, opened 2019-05-04
add type_create_subarray function

gasagna:
Hi, 

I have added wrappers for the MPI_Type_create_subarray function and included two new constants (MPI_ORDER_C/FORTRAN). I have defined the input arguments to be as close as possible to the openmpi documentation.

I am not sure how I would add tests here. Any suggestion is appreciated.


lcw:
Thanks for the contribution!  For an idea on how to test I would look at mpich's or openmpi's test suite to see how they test subarrays.  For example look [here][0].

[0]: https://github.com/pmodels/mpich/blob/6cb6a29a3d647b29c1a8061e108a44447c3331f8/test/mpi/datatype/subarray.c

gasagna:
Thanks for the link. They test subarray types in a simple way and that's what I did too here. In order to run the tests I have also ended up adding the ability to (I)send/(I)recv subarrays type.

As far as I can understand from the code, we were previously able to send contiguous subarrays only. Perhaps there is scope to generalise this into a single framework, or just provide enough docs to explain what is possible. 

I am going to write up some docs here and then this might be mergeable (?).

gasagna:
Looking at the travis build, it seems that testing has stalled on `test_cman_mpi.jl`. Any idea about what is going on here?

deps/build.jl
```diff
@@ -65,8 +65,13 @@ MPI_constants = [
     # proc
     "MPI_PROC_NULL",
 
+    # memory layout
+    "MPI_ORDER_C",
+    "MPI_ORDER_FORTRAN",
```
simonbyrne: these constants/functions need to be added to `src/win_mpiconstants.jl` as well. They're available from https://github.com/microsoft/Microsoft-MPI/blob/182a2815f6ce2dc4efc2f00b2bd8eb5e0f34ff7f/src/include/mpif.h

gasagna: I have added those in a previous commit.

gasagna:
The build is now failing for some reasons on  `test_shared_win.jl`. Any hints here?

gasagna:
There is a comment at the end of `test_shared_win.jl` saying 
```
run with `mpirun -np 3 julia --project test_shared_win.jl`
```
Locally, I have no issue with three processors but get this 
```
A system call failed during shared memory initialization that should
not have.  It is likely that your MPI job will now either abort or
experience performance degradation.

  Local host:  vega.local
  System call: unlink(2) /var/folders/dj/x44d4fyj56ndr9k5d2_8rg5w0000gn/T//ompi.vega.501/pid.41460/1/vader_segment.vega.11710001.0
  Error:       No such file or directory (errno 2)
```
with two processors (or four). Looking at `runtests.jl`, the number of processors we run test on anything between two and four processors.

barche:
It seems to be `test_sendrecv.jl` that is failing. I see the system call error locally too, I think it's a bug in openmpi related to oversubscribe.

gasagna:
Is this the first time this bug comes up (i.e. is this PR the problem)?

gasagna:
Bumping this, sorry. Does anyone have a suggestion on how to best proceed here? It seems this is not failing because of this PR, but for some other unrelated reasons.

barche:
The failed system call is definitely not related to this PR. But the errors on Travis seem to be different, I'm not sure what's causing those.

gasagna:
I have pushed one more commit with no major changes to bump the travis tests. It looks like the travis failures are random: some tests now fail on mac os, julia nightly, openmpi, whereas previously they were failing for julia 1 and 1.1 (still mac os and openmpi).

Not sure what's going on here.

simonbyrne:
I've been seeing this intermittently as well.

It could be that one of the tests isn't getting finalized properly?

lcw:
Simon Byrne <notifications@github.com> writes:

> I've been seeing this intermittently as well.
>
> It could be that one of the tests isn't getting finalized properly?

From what I can see the error usually occurs when calling
`test/test_shared_win.jl`, is that correct?

If so, I would suggest accepting this pull request with an issue opened
up to track errors in the tests.  If there is no objection I will do
this later this week.

PR #258, opened 2019-05-08
revamp Info object

simonbyrne:
It is generally not recommended to mutate String objects. Instead the correct way is to allocate a buffer, and construct the String from that.

Also I believe that gfortran Logical values are integers, not Bools (https://stackoverflow.com/a/26022787/392585)


simonbyrne:
I realised there was a lot to be left desired with regards to the Info object handling, so I changed it to match the AbstractDict interface

simonbyrne:
Okay, after more thought, I realised it should be an `AbstractDict{Symbol, String}` so that we can splat keyword arguments (since that roughly matches how it is used by MPI).

PR #259, opened 2019-05-08
JULIA_MPI_Fortran_INCLUDE_PATH doesn't need -I

milthorpe:
JULIA_MPI_Fortran_INCLUDE_PATH is a path to the include directory for mpif90 (not a flag passed to the compiler), so it doesn't need the "-I" at the start


lcw:
Thanks!

PR #266, opened 2019-06-05
support CUDA-aware MPI

### LINKED ISSUE #219: Support for CUDA aware MPI
Just capturing a discussion with @lcw I had a while ago.

A more and more common usecase is to combine CUDA and MPI, and some MPI implementations allow for RDMA to CUDA pointers.

The pattern one usually sees is
```
comm_buffer = ...
# recv data
copyto!(A, comm_buffer)
 ```

Where `A` is a `CuArray` and `comm_buffer` could either be a host array allocated in pinned memory or with CUDA-aware MPI could be a `CuArray` itself.

1. Add support for pinned memory e.g. `cudaMallocHost` to speed up the pattern.
2. Detect CUDA-aware MPI configuration and allow the usage of CUDA pointers directly.

This feature doesn't really fit MPI.jl or Cu*.jl, so we might want to create a glue repository that takes care of this.

- https://devblogs.nvidia.com/benchmarking-cuda-aware-mpi/
- https://github.com/parallel-forall/code-samples/tree/master/posts/cuda-aware-mpi-example/src

### PR DESCRIPTION
**Pull request by vchuravy**
fixes #219, still need to figure out how to test for CUDA-aware MPI and how to test this code.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,6 +9,7 @@ Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
+Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
 
 [compat]
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -1,6 +1,7 @@
 module MPI
 
 using Libdl, Serialization
+using Requires
 
 @static if Sys.iswindows()
     const depfile = "win_mpiconstants.jl"
@@ -76,6 +77,8 @@ function __init__()
 
         recordDataType(T, mpiT)
     end
+
+    @require CuArrays="3a865a2d-5b23-5a0f-bc46-62713ec82fae" include("mpi-cuda.jl")
 end
 
 end
--- a/src/mpi-cuda.jl
+++ b/src/mpi-cuda.jl
@@ -1,0 +1,20 @@
+import .CuArrays: CuArray
+import .CuArrays.CUDAdrv: CuPtr
+
+function Isend(buf::CuArray{T}, dest::Integer, tag::Integer, comm::MPI.Comm) where T
+    GC.@preserve buf begin
+        _buf = Base.cconvert(CuPtr{T}, buf)
+        cuptr = Base.unsafe_convert(CuPtr{T}, _buf)
+        ptr = reinterpret(Ptr{T}, cuptr)
+        MPI.Isend(ptr, length(buf), dest, tag, comm)
+    end
+end
+
+function Irecv!(buf::CuArray{T}, src::Integer, tag::Integer, comm::MPI.Comm) where T
+    GC.@preserve buf begin
+        _buf = Base.cconvert(CuPtr{T}, buf)
+        cuptr = Base.unsafe_convert(CuPtr{T}, _buf)
+        ptr = reinterpret(Ptr{T}, cuptr)
+        MPI.Irecv!(ptr, length(buf), src, tag, comm)
+    end
+end

### CODE REVIEW

**Comment by simonbyrne** [comment]
@vchuravy Any chance you can hook this up to the JuliaGPU GitLab CI?

**Comment by simonbyrne** [comment]
> still need to figure out how to test for CUDA-aware MPI and how to test this code.

It doesn't seem like there is a standard way. OpenMPI provides [`MPIX_Query_cuda_support()`](https://www.open-mpi.org/faq/?category=runcuda#mpi-cuda-aware-support). Others apparently require environment variables to be set (https://devblogs.nvidia.com/benchmarking-cuda-aware-mpi/).

**Comment by leios** [comment]
I don't know if this helps the PR, but we have tested this with the following script on OLCF's Ascent machine:

```
using CuArrays, CUDAnative, CUDAdrv, MPI

function GPU_transfer_test(comm)

    MPI.Barrier(comm)
    rank = MPI.Comm_rank(comm)
    size = MPI.Comm_size(comm)

    a = zeros(3)

    d_a = CuArray(zeros(3))

    if (rank == 0)
        a = [1.0, 2, 3]
        d_a = CuArray(a)
        a_check = Array(d_a)

        println(a_check == a, '\t', rank)

        sreq = MPI.Isend(d_a,1,1,comm)
        sreq = MPI.Isend(a,1,0,comm)
    end


    if (rank == 1)
        a_check = [1.0, 2.0, 3.0]

        rreq_gpu = MPI.Irecv!(d_a, 0, 1, comm)
        rreq_cpu = MPI.Irecv!(a, 0, 0, comm)
        MPI.Waitall!([rreq_cpu])
        MPI.Waitall!([rreq_gpu])

        println(a_check == Array(d_a), '\t', rank)
        println(a_check == a, '\t', rank)

    end


end

function main()
    MPI.Init()
    comm = MPI.COMM_WORLD

    GPU_transfer_test(comm)

    MPI.Finalize()
end

main()
```

It works as expected. Good job @vchuravy!

**Comment by simonbyrne** [comment]
It would be great if we could get this to work without adding new definitions for every function.

This would probably require getting rid of the `MPIBuffertype` union (since we can't extend it). Instead, we could define an `MPIPtr` type and require any array types to define appropriate conversion methods, i.e.

Something like (based on #271):
```julia
function Isend(buf, count::Integer, datatype::Union{Datatype, MPI_Datatype},
               dest::Integer, tag::Integer, comm::Comm) where T
    req = Request()
    # int MPI_Isend(const void* buf, int count, MPI_Datatype datatype, int dest,
    #               int tag, MPI_Comm comm, MPI_Request *request)
    @mpichk ccall((:MPI_Isend, libmpi), Cint,
          (MPIPtr, Cint, MPI_Datatype, Cint, Cint, MPI_Comm, Ptr{MPI_Request}),
                  buf, count, datatype, dest, tag, comm, req)
    req.buffer = buf
    return req
end

# behind a Requires block
Base.cconvert(::Type{MPIPtr}, buf::CuArray{T}) where{T} = buf
function Base.unsafe_convert(::Type{MPIPtr}, buf::CuArray{T}) where {T}
    _buf = Base.cconvert(CuPtr{T}, buf)
    cuptr = Base.unsafe_convert(CuPtr{T}, _buf)
    reinterpret(Ptr{T}, cuptr)
end
```

vchuravy:
That sounds like a good way to go about it.

kose-y:
To my knowledge, a `CuArray` should be synchronized before `Isend` or any other data-sending operations. OpenMPI does not synchronize CUDA in its implementation.

simonbyrne:
From what I understand, CuArrays handles synchronization itself, but @vchuravy should confirm.

kose-y:
I have a little bit outdated code (it worked 10 months ago when this package used Fortran API, and it looks like casting a CuPtr{T} to Ptr{T} is no longer working):
 https://github.com/kose-y/MPICuArrays.jl.

simonbyrne:
Thanks, that is useful.

I think you need to `reinterpret` instead of `convert`.

vchuravy:
> to my knowledge, a CuArray should be synchronized before Isend or any other data-sending operations. OpenMPI does not synchronize CUDA in its implementation.

Dp you have a reference for that? A device synchronization would be rather expensive.

kose-y:
[This paper](https://ieeexplore.ieee.org/abstract/document/6650987) shows some examples of CUDA-side synchronization before MPI data transfer.

Last summer, I lost a huge amount of time to this issue alone. I wrote a distributed matrix data structure on PyTorch with its MPI backend to support distributed matrix multiplication and blockwise operations (similar to [MPIArrays.jl](https://github.com/barche/MPIArrays.jl)) and wrote several statistical applications based on it. However, some applications were mysteriously giving incorrect results. The size of the matrix was about 10000 x 10000, distributed over 2, 4, or 8 GPUs. I tried inserting the barrier function, which merely called MPI's barrier function, but they still gave incorrect results. I inserted `print` functions everywhere to figure out what was going wrong, but then they mysteriously gave the correct result. Then I learned about the environment variable `CUDA_LAUNCH_BLOCKING`, and set it to get the correct result. It was only then when it was clear to me that not synchronizing was a problem. 

You might consider not putting synchronization in MPI.jl package and instructing users to synchronize the `CuArray` before sending the data. However, after what I experienced, I have a strong opinion on including such synchronization inside an MPI package that supports CUDA. If someone is trying to use multiple GPUs, the computation itself would be so heavy that the cost from synchronization would be relatively small.

vchuravy:
Thanks for sharing your experience. This is corroborate by p48-49 of
https://www.google.com/url?sa=t&source=web&rct=j&url=http://on-demand.gputechconf.com/gtc/2017/presentation/s7128-davide-rossetti-how-to-enable.pdf&ved=2ahUKEwiz7p6loqnjAhVQSN8KHZ1FCBAQFjAFegQIBxAI&usg=AOvVaw0gW0yxLrUW8_9Wz84Bs-QR

kose-y:
@vchuravy Thanks for the reference!

eschnett:
@kose-y The `MPI.jl` package is supposed to be a thin wrapper around an MPI library. While it adds serialization for Julia objects, the intent of this library is to be transparent and to let the user control what is happening. It might be that an array is already synchronized, so that any extra synchronization would be a wast of time.

I also disagree that MPI CUDA programs would usually be compute bound. For example, the application I have in mind requires a lot of memory, and is thus bound by the available memory on the GPUs. Any additional synchronisation would just add to the overhead.

kose-y:
OK. I understand. I will remove synchronization in PR #286, and add explicit synchronization in the examples there.

eschnett:
@kose-y I might have been too negative in my comment above. There is nothing wrong with a high-level interface that is easy to use and that catches common user errors. The point is that there also needs to be a low-level interface.

You could introduce a high-level function `IrecvSync` that synchronizes after sending. This could then be extended to non-CUDA cases as well, e.g. if the array is in shared memory (memory fence) or resides on the file system (flushing).

In these cases, is `sync` still the right term to use? Maybe using the multi-threading "acquire" and "release" terms would be better. Or maybe (since this is an MPI library) we should use MPI's terminology (MPI has "windows" for shared memory), i.e. "put" and "get".

vchuravy:
Yes I think we should clearly document it and see if we can wrap/detect the MPI extension

kose-y:
@eschnett Thank you for the encouragement. I agree that it would be better to consider possible non-CUDA extension.

vchuravy:
superseded by #302

PR #273, opened 2019-06-17
Cart coords fix

samo-lin:
Fix the function name to comply with the Julia convention of the usage of '!'.


simonbyrne:
Can you add a deprecation to deprecated.jl?

src/deprecated.jl
```diff
@@ -9,7 +9,7 @@ const FINALIZE_ATEXIT = Ref(true)
 @deprecate Info_get(info::Info,key::AbstractString) info[Symbol(key)] false
 @deprecate Info_delete(info::Info,key::AbstractString) delete!(info, Symbol(key)) false
 @deprecate Info_free(info::Info) free(info) false
-
+@deprecate Cart_coords!(comm::Comm, maxdims::Integer) Cart_coords(comm, maxdims) false
```
samo-lin: @simonbyrne : is this how it should be (coords allocated in the function and then returned)?

simonbyrne: Yes, that looks good, thanks.

simonbyrne:
Thanks!

PR #280, opened 2019-06-27
Add Universe_size

beltedmetric:
As part of development of MPIExecutor.jl, a package inspired by mpi4py that I'm working on, I need to be able to get the universe size to provide sane defaults.

### DIFF
--- a/src/comm.jl
+++ b/src/comm.jl
@@ -101,5 +101,8 @@ function Universe_size(comm::Comm)
     # int MPI_Comm_get_attr(MPI_Comm comm, int comm_keyval, void *attribute_val, int *flag)
     @mpichk ccall((:MPI_Comm_get_attr, libmpi), Cint,
         (MPI_Comm, Cint, Ptr{Cvoid}, Ptr{Cint}), comm, MPI_UNIVERSE_SIZE, result, flag)
+    if flag[] == 0
+        error("Attribute MPI_UNIVERSE_SIZE is not attached")
+    end
     unsafe_load(reinterpret(Ptr{Cint}, result[]))
 end
--- a/src/comm.jl
+++ b/src/comm.jl
@@ -100,6 +100,6 @@ function Universe_size(comm::Comm)
     result = Ref(Ptr{Cvoid}(C_NULL))
     # int MPI_Comm_get_attr(MPI_Comm comm, int comm_keyval, void *attribute_val, int *flag)
     @mpichk ccall((:MPI_Comm_get_attr, libmpi), Cint,
-        (MPI_Comm, Cint, Ptr{Cvoid}, Ptr{Cint}), comm, MPI_UNIVERSE_SIZE, result, rank)
+        (MPI_Comm, Cint, Ptr{Cvoid}, Ptr{Cint}), comm, MPI_UNIVERSE_SIZE, result, flag)
     unsafe_load(reinterpret(Ptr{Cint}, result[]))
 end
--- a/src/comm.jl
+++ b/src/comm.jl
@@ -95,3 +95,10 @@ function Intercomm_merge(intercomm::Comm, flag::Bool)
     return newcomm
 end
 
+function Universe_size(comm::Comm)
+    rank = Ref{Cint}()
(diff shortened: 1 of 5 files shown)


src/comm.jl
```diff
@@ -95,3 +95,10 @@ function Intercomm_merge(intercomm::Comm, flag::Bool)
     return newcomm
 end
 
+function Universe_size(comm::Comm)
+    rank = Ref{Cint}()
+    result = Ref(Ptr{Cvoid}(C_NULL))
+    @mpichk ccall((:MPI_Comm_get_attr, libmpi), Cint,
```
simonbyrne: I've started pasting the C signature as comments above the ccalls: it's been useful to ensure consistency.
```suggestion
    # int MPI_Comm_get_attr(MPI_Comm comm, int comm_keyval, void *attribute_val, int *flag)
    @mpichk ccall((:MPI_Comm_get_attr, libmpi), Cint,
```

src/comm.jl
```diff
@@ -95,3 +95,10 @@ function Intercomm_merge(intercomm::Comm, flag::Bool)
     return newcomm
 end
 
+function Universe_size(comm::Comm)
+    rank = Ref{Cint}()
```
simonbyrne: it doesn't store the MPI rank, so we should call it something else.
```suggestion
    flag = Ref{Cint}()
```

src/comm.jl
```diff
@@ -95,3 +95,10 @@ function Intercomm_merge(intercomm::Comm, flag::Bool)
...
+    result = Ref(Ptr{Cvoid}(C_NULL))
+    @mpichk ccall((:MPI_Comm_get_attr, libmpi), Cint,
+        (MPI_Comm, Cint, Ptr{Cvoid}, Ptr{Cint}), comm, MPI_UNIVERSE_SIZE, result, rank)
```
simonbyrne: (to match change above)
```suggestion
        (MPI_Comm, Cint, Ptr{Cvoid}, Ptr{Cint}), comm, MPI_UNIVERSE_SIZE, result, flag)
```

src/comm.jl
```diff
@@ -95,3 +95,10 @@ function Intercomm_merge(intercomm::Comm, flag::Bool)
     return newcomm
 end
 
+function Universe_size(comm::Comm)
```
simonbyrne: Since this doesn't correspond exactly to an MPI function, maybe make it lowercase?

```suggestion
function universe_size(comm::Comm)
```

src/comm.jl
```diff
@@ -95,3 +95,10 @@ function Intercomm_merge(intercomm::Comm, flag::Bool)
...
+    rank = Ref{Cint}()
+    result = Ref(Ptr{Cvoid}(C_NULL))
+    @mpichk ccall((:MPI_Comm_get_attr, libmpi), Cint,
+        (MPI_Comm, Cint, Ptr{Cvoid}, Ptr{Cint}), comm, MPI_UNIVERSE_SIZE, result, rank)
+    unsafe_load(reinterpret(Ptr{Cint}, result[]))
```
simonbyrne:
```suggestion
    if flag[] == 0
        error("Attribute MPI_UNIVERSE_SIZE is not attached")
    end
    unsafe_load(reinterpret(Ptr{Cint}, result[]))
```

simonbyrne:
Looks like MPICH doesn't define it unless specified as part of mpiexec or as an environment variable.

beltedmetric:
What do you propose to proceed? Can we throw an exception if running with
MPICH for example?

simonbyrne:
Supposedly MPICH should allow setting the universe size by the MPIEXEC_UNIVERSE_SIZE environment variable, but I couldn't get it to work.

Instead I changed it to return `nothing` if the attribute is not defined. Also since it is only attached tgo `MPI.COMM_WORLD`, I just made it a zero-argument function.

simonbyrne:
@beltedmetric Does this work for you?

beltedmetric:
Yes, this should be fine. Thanks.

PR #271, opened 2019-06-16
Move to C API

simonbyrne:
Fixes #59.

Works around the issues encountered last time (#169) by generating a small C
program which in turn generates the necessary constants, types and
offsets. To avoid the problem of link-time definitions, it uses the
Fortran constants for all handles, which are converted to the C variants
at `MPI_Init`.

This approach has several benefits:
 - No need for CMake (fixes #22) or a Fortran compiler
 - No implicit use of gfortran ABI for strings
 - Simpler to extend, as the new functions only need to be added in one place.

I've also taken the liberty splitting the source into several files
based on functionality, which should make it easier to find things.

PR #282, opened 2019-06-30
Store full path of libmpi and mpiexec in deps.jl

simonbyrne:
This makes it slightly easier to use on clusters since it avoids the need to "module load mpiXXX" each time it is used, and makes it more robust if the wrong MPI library is loaded.

Also improves the error message if no MPI library is found.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -38,14 +38,17 @@ end
 const libmpi = find_library(Sys.iswindows() ? "msmpi.dll" : "libmpi",
                             MPI_LIBRARY_PATH !== nothing ? [MPI_LIBRARY_PATH] : [])
 
-libptr = dlopen_e(libmpi)
-if libmpi == "" || libptr == C_NULL
-    error("No MPI library found")
+if libmpi == ""
+    error("No MPI library found.\nEnsure an MPI implementation is loaded, or set the `JULIA_MPI_PATH` variable.")
 end
-@info "Using MPI library $libmpi"
 
-libpath = dlpath(libptr)
-libsize = filesize(libpath)
+# expand paths
+libmpi = dlpath(libmpi)
+libsize = filesize(libmpi)
+
+mpiexec = Sys.which(mpiexec)
+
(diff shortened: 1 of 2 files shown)


vchuravy:
Thanks! I was wondering if we can use `pkg-config` to replace the auto-configuration through CMake.

simonbyrne:
> Thanks! I was wondering if we can use `pkg-config` to replace the auto-configuration through CMake.

Would that solve the issues you mentioned in #271?

vchuravy:
Yeah probably, once Summit is back up I will put up a PR:

```
vchuravy@cyclops:~$ pkg-config --libs-only-L mpi
-L/usr/lib/x86_64-linux-gnu/openmpi/lib
```

vchuravy:
Maybe this is enough:
```
vchuravy@cyclops:~$ pkg-config --variable=libdir mpi
/usr/lib/x86_64-linux-gnu/openmpi/lib
```
And add that as a optional path to `find_library`

simonbyrne:
Is that able to find paths that we wouldn't get via `find_library`/`dlpath`?

vchuravy:
Yes. `find_library` only finds things in `DL_LIBRARY_PATH` whereas `pkg-config` has it's own mechanism under which packages are being registered and made available. (through pc files).

~edit:~
Huh, according to the mapage of `dlopen` - e.g. what Julia uses internally we should be  looking into `LD_LIBRARY_PATH`... 

>        o   If,  at  the time that the program was started, the environment variable LD_LIBRARY_PATH was defined to contain a colon-separated list of directories, then these are searched.  (As a security measure, this variable is ignored for set-user-ID and set-group-ID programs.)

So I guess I need to do some investigating why that didn't work.

simonbyrne:
`DL_LIBRARY_PATH`/`LD_LIBRARY_PATH`: that's not confusing at all...

vchuravy:
Yeah one of them is the system the other one is Julia...

PR #284, opened 2019-07-10
use cfunction closures for user-defined operators

simonbyrne:
See https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/index.html#Closure-cfunctions-1

I've also tweaked it so that the type is now an argument, so that we should be able to use operators which are defined for builtins (say `+`) with user-defined types. 

@lcw I was thinking of adding a test-case for `DoubleFloat`s: do you have a simple one?


lcw:
I don't have a particular test for `DoubleFloat`s at hand.  I think it would be good to add.

simonbyrne:
Okay, I added a simple summation test.

PR #286, opened 2019-07-10
CUDA-aware MPI

kose-y:
Updated code in https://github.com/kose-y/MPICuArrays.jl to work with the current master. 

- extends #266. Synchronization before sending any data in CuArray level. 
- includes #285 
- added examples in examples/cuda, which are tweaks of existing examples for CuArrays. Each rank selects a device in a round-robin fashion.


examples/cuda/02-broadcast.jl
```diff
@@ -0,0 +1,17 @@
+import MPI
+MPI.Init()
+comm = MPI.COMM_WORLD
+using CuArrays
+using CuArrays.CUDAdrv
+using CUDAnative
+CUDAnative.device!(MPI.Comm_rank(comm) % length(devices()))
```
vchuravy: I believe the code below is a slightly better way to ensure the correct allocation of device id to rank on a machine.

```julia
lcomm = MPI.Comm_split_type(comm, MPI.MPI_COMM_TYPE_SHARED,  MPI.Comm_rank(comm))
CUDAnative.device!(MPI.Comm_rank(lrank))
```

simonbyrne:
I still think switching to something like https://github.com/JuliaParallel/MPI.jl/pull/266#issuecomment-505170266 would solve the boilerplate problem.

examples/cuda/02-broadcast.jl
hiddenidol: Thanks! I had no experience with Comm_split_type before.

kose-y: Thanks! I had no experience with GPUs over multiple machines.

kose-y:
@simonbyrne I agree, but that would be a major design change for the entire package. I was trying to add something that works based on the code base I had with this PR. I will try to switch to that design, and it will take some time for me.

kose-y:
By the way, what do you think about https://github.com/JuliaParallel/MPI.jl/blob/615ca462a8905f6b17b5c7ae17edcdf2c4e86b7d/src/pointtopoint.jl#L204 

and 

```Julia
function Isend(buf::CuArray{T}, dest::Integer, tag::Integer, 
	                comm::MPI.Comm) where T
```
?
Should we merge them?

kose-y:
@simonbyrne 

- In `datatypes.jl`: defined `MPIPtr`, `cconvert` and `unsafe_convert`
- In `pointtopoint.jl`: generalized `Isend`, `Irecv!`
- In `collective.jl`: generalized `Reduce!`, and `Bcast!`
- In `cuda.jl`: defined `MPIPtr(x::CuArray{T})`, `Base.cconvert(::Type{MPIPtr}, buf::CuArray{T}) where T`, and `Base.unsafe_convert(::Type{MPIPtr}, buf::DeviceBuffer)`. All implementations of MPI functions are removed.

src/pointtopoint.jl
```diff
@@ -188,9 +188,9 @@ of communicator `comm` using with the message tag `tag`
 
 Returns the commication `Request` for the nonblocking send.
 """
-function Isend(buf::MPIBuffertype{T}, count::Integer,
-               dest::Integer, tag::Integer, comm::Comm) where T
-    Isend(buf, count, mpitype(T), dest, tag, comm)
+function Isend(buf, count::Integer,
+               dest::Integer, tag::Integer, comm::Comm)
+    Isend(buf, count, mpitype(Base.eltype(buf)), dest, tag, comm)
```
kose-y: Note on test failure on Julia nightly: On commits https://github.com/JuliaParallel/MPI.jl/pull/286/commits/d452f85019746ac49e69a7b9e2e15972327e8b8e and https://github.com/JuliaParallel/MPI.jl/pull/286/commits/c0eb592a2b3f1dc28426b8ddb8d579c0a38c92d7, the test `test/test_cman_mpi.jl` failed on Julia nightly build only. The reason seemed to be this line, so I changed lines 205 and 219  to avoid this line altogether on `cman.jl`. What should be done with this line?

src/cuda.jl
```diff
@@ -1,28 +1,19 @@
 import .CuArrays: CuArray
 import .CuArrays.CUDAdrv: CuPtr, synchronize
+import .CuArrays.CUDAdrv.Mem: DeviceBuffer
 
-function get_ptr(buf::CuArray{T}) where T
-    _buf = Base.cconvert(CuPtr{T}, buf)
-    cuptr = Base.unsafe_convert(CuPtr{T}, _buf)
-    reinterpret(Ptr{T}, cuptr)
-end
+MPIPtr(x::CuArray{T}) where T = reinterpret(MPIPtr, Base.unsafe_convert(x))
```
vchuravy: This is slightly dangerous since it loses a GC reference.

src/datatypes.jl
```diff
@@ -11,6 +11,26 @@ MPIBuffertype{T} = Union{Ptr{T}, Array{T}, SubArray{T}, Ref{T}}
...
+if sizeof(Ptr{Cvoid}) == 8
+    primitive type MPIPtr 64 end
+else
+    primitive type MPIPtr 32 end
+end
+
+MPIPtr(x::Cint) where T = reinterpret(MPIPtr, x)
+
+
+MPIPtr(x::MPIBuffertype{T}) where {T} = reinterpret(MPIPtr, Base.unsafe_convert(Ptr{T}, x))
```
vchuravy: Same here

src/cuda.jl
kose-y: Removed those constructors. They were not really necessary.

kose-y:
Updated other MPI APIs on `pointtopoint.jl`, `collective.jl`, and `window.jl`.

src/datatypes.jl
```diff
@@ -24,7 +24,11 @@ end
 MPIPtr(x::Cint) where T = reinterpret(MPIPtr, x)
 
 
-MPIPtr(x::MPIBuffertype{T}) where {T} = reinterpret(MPIPtr, Base.unsafe_convert(Ptr{T}, x))
+function MPIPtr(x::MPIBuffertype{T}) where {T}
+    GC.@preserve x begin
```
vchuravy: the GC.preserve won't help. Why are these methods needed?

kose-y: Sorry for the confusing commit message. Those methods were removed in the next commit (https://github.com/JuliaParallel/MPI.jl/pull/286/commits/0349b34303ffe78bf0beff626c65a7f3bba11a05), not https://github.com/JuliaParallel/MPI.jl/pull/286/commits/0ec5fb0c7bd125d4925654049ea8c7631171f7ac .  I figured that GC won't help so just removed it, but forgot to stage it for https://github.com/JuliaParallel/MPI.jl/pull/286/commits/0ec5fb0c7bd125d4925654049ea8c7631171f7ac.

vchuravy:
I am in progress of adding a GPU CI to here #299 once we have that I think this is ready to get merged.
Sadly the CI machine only has one GPU.

src/cuda.jl
```diff
@@ -0,0 +1,144 @@
...
+function Base.cconvert(::Type{MPIPtr}, buf::CuArray{T}) where T
+    Base.cconvert(CuPtr{T}, buf) # returns DeviceBuffer
+end
+
+function Base.unsafe_convert(::Type{MPIPtr}, buf::DeviceBuffer)
+    reinterpret(MPIPtr, buf.ptr)
+end
+
+
+#=
```
vchuravy: these commented methods should be removed

simonbyrne:
Included in #302.

simonbyrne:
Thanks @kose-y for your persistence here, I'm very excited for this.

PR #295, opened 2019-07-28
Modify Test interface

simonbyrne:
This changes the return values of `Test!`/`Testany!`/`Testall!` to match those of the corresponding `Wait` functions, except that they return `nothing` if the requests have not completed. Also, add some more docs.

Note this is a breaking change. I couldn't see a clear way to deprecate this.

### DIFF
--- a/src/pointtopoint.jl
+++ b/src/pointtopoint.jl
@@ -36,8 +36,18 @@ if !@isdefined(Status)
             offset += 4
         end
 
-        @eval struct Status
-            $(struct_fields...)
+        @eval begin
+            """
+    Status
+
+Represents the status of an MPI operation. This has 3 documented fields
+  - `source`
+  - `tag`
+  - `error`
+"""
+            struct Status
+                $(struct_fields...)
+            end
         end
     end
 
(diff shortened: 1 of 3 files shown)


src/pointtopoint.jl
```diff
@@ -532,9 +573,17 @@ function Waitsome!(reqs::Vector{Request})
...
+    resize!(stats, outcount)
     (indices, stats)
 end
 
+"""
+    (indices, statuses) = Testsome!(reqs::Vector{Request})
+
+Check if the requests in `reqs` have completed. Returns the indices of the completed
+requests in `reqs`, and their corresponding [`Status`](@ref)'s. If none of the requests
+have completed, then both arrays will be empty.
```
simonbyrne: Note that `Testsome!` is the exception to this rule.

simonbyrne:
The only way I can see to deprecate it is to switch to lowercase function names (which might not be a bad idea for matching Julia style).

lcw:
I am okay with switching to lowercase function names.

If we go this way, I think it would be useful to know if
`MPI.serialize(obj)` is being called. Or specifically disabling it via
`MPI.allowserialize(false)`, similar to `CuArrays.allowscalar(false)`.

simonbyrne:
Actually there are two other things we may want to consider before making more changes:
- A way to avoid allocating for `Status` objects (corresponding to `MPI_STATUS_IGNORE`/`MPI_STATUSES_IGNORE`)
- Whether to drop the completed requests from the arrays?

simonbyrne:
This was resolved in #479

Issue #289, opened 2019-07-12
Tests failing on nightly.


simonbyrne:

simonbyrne:
I have no idea how to figure this out. Given the ancillary nature, should we move the cluster managers to a separate package?

barche:
Moving seems like a good idea to me, that might also provide an opportunity to integrate PR #203 into the new package.

simonbyrne:
I'm guessing the actual issue is due to the new threading runtime, which is now in nightly (since it seems to fail at `yield`. Unfortunately I don't really understand the code well enough to figure out what is going on.

PR #296, opened 2019-07-28
Deprecate cluster manager

simonbyrne:
See discussion in #289.

I'm in the process of setting up a new package, so we should probably wait until that is completed.


simonbyrne:
Okay, I have changed it to a deprecation, and created the new package here:
https://github.com/JuliaParallel/MPIClusterManagers.jl

PR #298, opened 2019-08-03
disable UCX memory hooks

vchuravy:
Work around https://github.com/openucx/ucx/issues/4001

@simonbyrne can you check that this fully fixes your issue?


simonbyrne:
Yes, this did fix it, thanks.

PR #303, opened 2019-08-20
use ccache on gitlab

vchuravy:


vchuravy:
Failure is: https://github.com/JuliaLang/julia/pull/32948

PR #305, opened 2019-08-29
Improve build process

simonbyrne:
- search for libmpich if libmpi isn't found
- run gen_consts under mpi launcher

This should fix some of the issues when running with Cray MPI.

cc @markwdalton


deps/build.jl
```diff
@@ -74,18 +79,15 @@ open("deps.jl","w") do f
...
         println(f, :(include("consts_msmpi.jl")))
-    end
-else
-    include("gen_consts.jl")
+    else
+        include("gen_consts.jl")
 
-    run(`$mpicc gen_consts.c -o gen_consts $CFLAGS`)
+        run(`$mpicc gen_consts.c -o gen_consts $CFLAGS`)
+        run(`$mpiexec -n 1 ./gen_consts`)
```
vchuravy: Really that's needed? :/

PR #307, opened 2019-09-12
Add and use Mprobe, Mrecv, Improbe, Imrecv

bn222:
If probe is used to find the size of a message followed by a recv, MPI does not
guarantee that the two calls refer to the same message. On larger systems, with
many ranks communicating, different message sizes, using probe+recv results in
race conditions. Therefore, MPI introduced the Mprobe (returning a handle) which
can be passed to Mrecv to receive the same message.


simonbyrne:
Superseded by #699

PR #319, opened 2019-11-19
Change Sendrecv to Sendrecv!

simonbyrne:
Since it is a mutating operation.

cc: @gasagna


vchuravy:
When did we add it? Probably needs a deprecation

simonbyrne:
Yesterday, and there haven't been any releases with it yet

PR #321, opened 2019-11-20
Add reference counting for Request objects

simonbyrne:
Also Cancel! will no longer remove reference to buffer, as communication may still occur until the request is deallocated.


simonbyrne:
I've also updated docstrings and slightly changed the behaviour of Testany to match MPI. As a result, I've bumped the version and removed deprecations from last version.

PR #323, opened 2019-12-12
Change "Fortran Function" => "C Function" in docs

kose-y:
There are several mentions of "Fortran Function"s in the docs, which are now changed to C functions


simonbyrne:
Thanks. I’m in the process of moving that page into new sections which should make it easier to understand.

Issue #308, opened 2019-09-12
Is that possible to send/receive a number instead of array?


zhaotianjing:
Hi, I want to send a number between two rank via point-to-point communication.
But I found the send/receive function only allows the array to be sent.

Assume we have 2 ranks. The following code failed:
```
using MPI
MPI.Init()

comm = MPI.COMM_WORLD
rank = MPI.Comm_rank(comm)

if rank==0
    data=7.0
    MPI.Send(data,1,99,comm) 
elseif rank==1
    data=[]
    MPI.Recv!(data,0,99,comm)
    println(data)
end


MPI.Finalize()
```
Could you please tell me how to send a number?

Thanks,
Tianjing

simonbyrne:
To follow up on this, you can also now do:
```julia
if rank == 0
    data = 7.0
    MPI.Send(data,1, 99,comm)
elseif rank == 1
    data = MPI.Recv(Float64,0,99, comm)
    println(data)
end
```
which should be more efficient as it won't incur serialization overhead.

PR #325, opened 2019-12-15
Document `MPI.send` and `MPI.recv`

ali-ramadhan:
Found out about them from issue #308 as I was hoping to send and receive some named tuples so they were useful. But I noticed that they weren't showing up in the documentation so this PR adds a couple of docstrings and adds these functions to `pointtopoint.md`.

I think it might also be useful to spend a paragraph explaining what can be sent and received using which functions. From my understanding:

MPI functions expect arrays of MPI data types (simple C data types only). And if you want to send anything else you must either define your own MPI data type (not sure how to do this) or serialize the object you want to send (which `MPI.send` and `MPI.recv` do for you).

### DIFF
--- a/docs/src/pointtopoint.md
+++ b/docs/src/pointtopoint.md
@@ -19,7 +19,10 @@ MPI.Get_count
 
 ```@docs
 MPI.Send
+MPI.send
 MPI.Recv!
+MPI.Recv
+MPI.recv
 MPI.Sendrecv!
 ```
 
@@ -27,6 +30,7 @@ MPI.Sendrecv!
 
 ```@docs
 MPI.Isend
+MPI.isend
 MPI.Irecv!
 MPI.Test!
 MPI.Testall!
@@ -45,5 +49,3 @@ MPI.Iprobe
 MPI.Probe
 MPI.Cancel!
 ```
-
-
--- a/src/pointtopoint.jl
+++ b/src/pointtopoint.jl
@@ -172,7 +172,7 @@ Get_count(stat::Status, ::Type{T}) where {T} = Get_count(stat, mpitype(T))
          dest::Integer, tag::Integer, comm::Comm) where T
 
 Perform a blocking send of `count` elements of type `datatype` from `buf` to MPI
-rank `dest` of communicator `comm` using the message tag `tag`
+rank `dest` of communicator `comm` using the message tag `tag`.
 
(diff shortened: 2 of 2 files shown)


ali-ramadhan:
As a side note, to avoid missing docstrings MPI.jl could use [`@autodocs`](https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@autodocs-block-1) to splice in all available docstrings.

simonbyrne:
Thanks @ali-ramadhan!

For any bitstype, you probably do want to use MPI datatypes instead of serialization: if you know the types, the `Send` and `Recv` should now just work. Do you have an example?

ali-ramadhan:
Turns out serialization is expensive so I ended up getting rid of the send/recv calls and just inferring information locally.

Will look into MPI data types in the future!

I was originally trying to `MPI.Gather` a bunch of named tuples when I noticed that I have to use send/recv as `MPI.Gather` doesn't allow you to pass a datatype (and doesn't serialize stuff for you).

simonbyrne:
As in you have a `NamedTuple` on each rank, and want a vector of `NamedTuple`s on the root?

simonbyrne:
e.g. the following works:
```julia
using Test, Pkg
using MPI

MPI.Init()

comm = MPI.COMM_WORLD
root = 0
rank = MPI.Comm_rank(comm)
sz   = MPI.Comm_size(comm)

x = (a=Float64(rank), b=Int(rank))

C = MPI.Gather(x, root, comm)

if rank == root
    @show C
end
```

ali-ramadhan:
Thanks for the example! Indeed that does work.

Hmmm, I had some nothings in my named tuple and ended up with a cryptic error when I added a `nothing` to the tuple: `Can't convert 0-size type to MPI`

I suppose this makes sense in hindsight as MPI/C don't know what a `nothing` is. I guess a `nothing` has size 0.

```julia
using MPI

MPI.Init()

comm = MPI.COMM_WORLD
root = 0
rank = MPI.Comm_rank(comm)
sz   = MPI.Comm_size(comm)

x = (a=Float64(rank), b=Int(rank), c=nothing)

C = MPI.Gather(x, root, comm)

if rank == root
    @show C
end
```

simonbyrne:
We should be able to fix that: can you open an issue?

PR #331, opened 2020-01-07
Improve README

simonbyrne:
Provide some links and guidance.

Any other thoughts on what would be good here?

[skip ci]


lcw:
Your changes look good to me.

It might be useful to add a quick start link pointing to <https://juliaparallel.github.io/MPI.jl/latest/usage/#Basic-example-1>?

simonbyrne:
Good call, thanks.

PR #339, opened 2020-01-19
use MPICH_jll binaries

simonbyrne:
This is exploratory to see if we can use the BinaryProvider binaries.


simonbyrne:
All builds seem to be failing on the I/O tests:
https://travis-ci.org/JuliaParallel/MPI.jl/jobs/639035821#L820

cc: @ViralBShah

ViralBShah:
Do they work locally?

simonbyrne:
No, I see the same error

ViralBShah:
The test does `tempname()` but that only generates a filename. Opening it should be an error. Perhaps a bug in the test?

simonbyrne:
The first step in the test creates and writes to it, though it also fails if the path exists.

I think it is some sort of error in the ROMIO sub library

ViralBShah:
I guess the next step is to see if a regular C test works. Something like:

```
#include "mpi.h"
#include <stdio.h>

int main( int argc, char *argv[] )
{
    MPI_Fint handleA, handleB;
    int rc;
    int errs = 0;
    int rank;
    MPI_File cFile;

    MPI_Init( &argc, &argv );
    MPI_Comm_rank( MPI_COMM_WORLD, &rank );
    rc = MPI_File_open( MPI_COMM_WORLD, "temp", MPI_MODE_RDWR | MPI_MODE_DELETE_ON_CLOSE | MPI_MODE_CREATE, MPI_INFO_NULL, &cFile );
    if (rc) {
        printf( "Unable to open file \"temp\"\n" );fflush(stdout);
    }
    else {
        MPI_File_close( &cFile );
    }
    MPI_Finalize();
    return 0;
}
```

simonbyrne:
I think the issue is this line:
```
[22:53:11] checking configured file systems... 
```
https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=1006&view=logs&j=8489172e-6a78-5283-55c2-fc688c24d963&t=6834a5e3-c1b7-5310-45b4-a7f4f98fab9f&l=13785

When I run it locally, I get
```
checking configured file systems... testfs ufs nfs
```

simonbyrne:
it appears to be working! thanks @ViralBShah and @giordano.

simonbyrne:
Replaced by #367

PR #340, opened 2020-01-19
Define MPI.Cart_rank and add topology docs

jipolanco:


simonbyrne:
Thanks for this, it is good to have it wrapped, and to clean up the interfaces as bit. 

Can we also get rid of the `maxdims` argument in some of these calls by calling `MPI_Cartdim_get`?

jipolanco:
I agree that it would be nice to provide an interface without the `maxdims` argument. I just modified `Cart_get` and `Cart_coords` as you suggest, marking the old signatures (with `maxdims`) as deprecated.

src/topology.jl
```diff
@@ -52,34 +95,71 @@ function Cart_get(comm::Comm, maxdims::Integer)
...
+function Cart_coords(comm::Comm)
+    maxdims = Cartdim_get(comm)
     ccoords = Vector{Cint}(undef, maxdims)
     rank    = Comm_rank(comm)
-    Cart_coords!(comm, rank, maxdims, ccoords)
+    Cart_coords!(comm, rank, ccoords)
     Int.(ccoords)
 end
 
+@deprecate(Cart_coords!(comm, rank, maxdims::Integer, coords),
```
simonbyrne: Can you move the deprecated calls to `deprecated.jl`?

src/topology.jl
```diff
@@ -104,3 +192,8 @@ function Cart_sub(comm::Comm, remain_dims)
     end
     comm_sub
 end
+
+function Cart_sub(comm::Comm, remain_dims)
```
simonbyrne: This is a bit of an inconsistent interface (and won't work on 32-bit machines where `Int == Cint`). One option would be to make it a keyword argument, i.e. `function Cart_sub(comm::Comm; keepdims=keepdims)` 

(it probably shouldn't be just `dims`, since this has basically the opposite behaviour in e.g. `mapslices` or `reduce`).

simonbyrne: Oh, i misunderstood what his was, nevermind.

src/topology.jl
```diff
@@ -33,14 +49,41 @@ function Cart_create(comm_old::Comm, dims::AbstractArray{T,N}, periods::Array{T,
...
+    rank = Ref{Cint}()
+    # int MPI_Cart_rank(MPI_Comm comm, const int coords[], int *rank)
+    @mpichk ccall((:MPI_Cart_rank, libmpi), Cint,
+                  (MPI_Comm, Ptr{Cint}, Ptr{Cint}),
+                  comm, coords, rank)
+    Int(rank[])
+end
+
+function Cart_rank(comm::Comm, coords::AbstractArray{T}) where {T <: Integer}
+    ccoords = Cint.(coords[:])
```
simonbyrne: Why not just?
```suggestion
    ccoords = Cint.(coords)
```

src/topology.jl
```diff
@@ -91,9 +171,17 @@ function Cart_shift(comm::Comm, direction::Integer, disp::Integer)
     Int(rank_source[]), Int(rank_dest[])
 end
 
-function Cart_sub(comm::Comm, remain_dims)
+"""
+    comm_sub = Cart_sub(comm::Comm, remain_dims)
+
+Create lower-dimensional Cartesian communicator from existent Cartesian
+topology.
```
simonbyrne: Maybe explain what `remain_dims` is (i.e. a boolean vector)

src/topology.jl
jipolanco: Ohh right, I missed that somehow.

src/topology.jl
jipolanco: I agree, I just saw the same pattern used in other parts of the code (in `topology.jl` anyway), so I copied it. I'll change that one and a few other occurrences.

simonbyrne: It might be good to actually define it as
```
ccoords = [Cint(c) for c in coords]
```
and relax the type signature so that we can feed in tuples.

jipolanco: Right. I guess we could do the same with `Dims_create!` and `Cart_create`, no? In their current version they don't accept tuples:
```julia
Dims_create!(nnodes::Integer, dims::AbstractArray{T,N}) where {T<:Integer, N}
Cart_create(comm_old::Comm, dims::AbstractArray{T,N}, periods::Array{T,N}, reorder) where T <: Integer where N
```

src/topology.jl
jipolanco: Done in https://github.com/JuliaParallel/MPI.jl/pull/340/commits/8e8e0ed08f721b6294bdd88ee6ccea07dbeac1ab

src/topology.jl
jipolanco: Done in https://github.com/JuliaParallel/MPI.jl/pull/340/commits/c0a54f988cfc2213d876a2c2345e0e6eab464d9a

src/topology.jl
jipolanco: Oops, sorry, I accidentally started a couple of reviews by pasting the commit url...

src/topology.jl
jipolanco: By the way, is there any reason to accept `AbstractArray`s of arbitrary dimension in these functions? For me, it would only make sense for the `dims` argument to be an `AbstractVector` or a tuple...

I say this, because the current version of `Dims_create!` has the line
```julia
cdims = Cint.(dims[:])
```
Replacing this by `cdims = [Cint(d) for d in dims]` would make things fail if `dims` were a matrix. And that is actually the case in `test/test_cart_create.jl`, where `dims` is initialised as `[0 0 0]` (a 1×3 matrix).

simonbyrne: > Replacing this by cdims = [Cint(d) for d in dims] would make things fail if dims were a matrix.

That should still work (you would get a matrix back, but that's fine to pass to a `ccall`).

simonbyrne:
Thanks!

PR #342, opened 2020-01-25
rename MPI.Types module to MPI.Type

simonbyrne:
Makes the name consistent with MPI (see #338).


simonbyrne:
Will close for now. Can reopen if people feel differently

PR #343, opened 2020-01-25
mention UCX multithreading issue

simonbyrne:
document issue and fix to #337.


simonbyrne:
I decided to just set the variable in init, see #363.

PR #345, opened 2020-01-30
add JULIA_MPIEXEC_ARGS env variable

### LINKED ISSUE #344: Allow arguments to JULIA_MPIEXEC
Building MPI with cray-mpich (v7.7.10) works perfectly well, except for the following little issue.

Due to the configuration of our scheduler (SLURM), I need to manually change `deps/build.log` and add  `-C gpu` or `-C mc` to line 89 (note that I set `JULIA_MPIEXEC=srun`). I.e. I need to replace
```
run(`$mpiexec -n 1 ./gen_consts`)
```
with e.g. 
```
run(`$mpiexec -C mc -n 1 ./gen_consts`)
```
Thus, it would be good to allow arguments to `$mpiexec` for this run command. Even if this fix is very simple, it is important for us as we use automatic build scripts and the requirement of changes in the source code is an obstacle.

Thanks!

Here is the error message for completeness:
```
  Building MPI → `~/.julia/1.3.1/daint-mc/packages/MPI/hcbnk/deps/build.log`
┌ Error: Error building `MPI`: 
│ srun: error: You have to specify, at least, what sort of node you need: -C gpu for GPU enabled nodes, or -C mc for multicore nodes.
│ srun: error: Other features are possible, but 'gpu' and 'mc' are exclusive.
│ srun: error: Unable to allocate resources: Requested node configuration is not available
│ [ Info: Using MPI library /opt/cray/pe/mpt/7.7.10/gni/mpich-gnu/8.2/lib/libmpich.so
│ ERROR: LoadError: failed process: Process(`srun -n 1 ./gen_consts`, ProcessExited(1)) [1]
```

simonbyrne:
Fixes #344

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -24,6 +24,8 @@ const mpiexec_path = get(ENV, "JULIA_MPIEXEC") do
     end
 end
 
+const mpiexec_args = split(get(ENV, "JULIA_MPIEXEC_ARGS", ""))
+
 const libmpi = get(ENV, "JULIA_MPI_LIBRARY") do
     libmpi = find_library(["libmpi", "msmpi", "libmpich"],
                           MPI_LIBRARY_PATH !== nothing ? [MPI_LIBRARY_PATH] : [])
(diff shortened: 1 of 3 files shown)


omlins:
thanks @simonbyrne !

PR #353, opened 2020-02-26
Skip unsupported tests on PPC/ARM

vchuravy:


simonbyrne:
For future reference, the reason they're unsupported is that custom reduction operators currently use closure cfunctions: https://github.com/JuliaLang/julia/pull/34326

PR #355, opened 2020-02-27
expose interface to check CUDA support

simonbyrne:
cc: @kpamnany

### DIFF
--- a/src/environment.jl
+++ b/src/environment.jl
@@ -172,8 +172,8 @@ or `false`.
 function has_cuda()
     flag = get(ENV, "JULIA_MPI_HAS_CUDA", nothing)
     if flag === nothing
-        # Only OpenMPI provides a function to check CUDA support
-        @static if dlsym_e(dlopen(libmpi), :MPIX_Query_cuda_support) != C_NULL
+        # Only Open MPI provides a function to check CUDA support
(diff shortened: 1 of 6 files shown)


kpamnany:
LGTM.

Also, I manually tested this on Central with OpenMPI, with and without CUDA support, and with MPICH, without CUDA support. Works fine.

src/environment.jl
```diff
@@ -158,3 +158,28 @@ function Wtime()
...
+Check if the MPI implementation is known to have CUDA support. Currently only Open MPI
+provides a mechanism to check, so it will return `false` with other implementations
+(unless overriden).
+
+This can be overriden by setting the `JULIA_MPI_HAS_CUDA` environment variable to `true`
+or `false`.
+"""
+function has_cuda()
+    flag = get(ENV, "JULIA_MPI_HAS_CUDA", nothing)
+    if flag === nothing
```
vchuravy: Could we use this to actually provide a function `MPI_vendor()`?

@leios wanted something like that.

simonbyrne: My plan is to do this as part of adding BinaryProvider support, but I don’t have the bandwidth to do this now

kpamnany: #354 does that but doesn't help with CUDA awareness.

simonbyrne: I've switched this to use #354.

PR #361, opened 2020-03-16
Use consts for known ABIs

simonbyrne:
Adds a mechanism to detect the ABI via MPI_LIBRARY_VERSION_STRING, and use that to detect the appropriate consts file, falling back on the C generator only when that fails.


simonbyrne:
@vchuravy or @leios: can you try this out on Spectrum MPI? I don't know if the constants are the same as OpenMPI or not.

simonbyrne:
Note that this disables the automatic `--oversubscribe` for OpenMPI tests: you can however set the environment variable `JULIA_MPIEXEC_ARGS="--oversubscribe"`

simonbyrne:
@leios What does
```
MPI.MPI_LIBRARY
MPI.MPI_LIBRARY_VERSION
MPI.MPI_LIBRARY_ABI
```
give?

leios:
```
julia> MPI.MPI_LIBRARY_ABI
OpenMPIABI::MPIABI = 2

julia> MPI.MPI_LIBRARY_VERSION
v"10.3.0"

julia> MPI.MPI_LIBRARY
IBMSpectrumMPI::MPIImpl = 5
```

PR #386, opened 2020-05-09
Add shell wrapper around `mpiexec`

### LINKED ISSUE #380: external mpiexec interface
Was discussed in https://github.com/JuliaParallel/MPI.jl/pull/367#issuecomment-624094910:
> Rather than a symlink to mpiexec/mpirun you probably want a wrapper script (Julia or bash?), as it'll need to know also where to find libmpi. I may find some time this coming weekend to cook something for this.

giordano:
Fix #380.


simonbyrne:
@lcw and @PhilipVinc you both expressed interest in this: any thoughts/comments?

contrib/mpiexecjl
```diff
@@ -0,0 +1,60 @@
...
+const script = """
+    using MPI
+    ENV["JULIA_PROJECT"] = dirname(Base.active_project())
+    mpiexec() do exe
+        if exe isa Cmd
+            cmd = Cmd([MPI.mpiexec_path, ARGS...])
+        else # if exe isa String
+            cmd = Cmd([exe, ARGS...])
+        end
+        run(cmd)
```
simonbyrne, line 52: Any reason why not just do
```suggestion
        run(`$exe $ARGS`)
```
?

giordano: Not at computer now, but if I remember correctly this doesn't work when `exe` is a `Cmd`, which is the case when `JULIA_MPIEXEC_BINARY=system`

simonbyrne: It does for me:
```
julia> exe = `mpiexec -foo`; ARGS = ["aa", "bb"]
2-element Array{String,1}:
 "aa"
 "bb"

julia> `$exe $ARGS`
`mpiexec -foo aa bb`
```

giordano: Well, at this point I don't know what I had tried :confused:  I simplified the function as you suggested

simonbyrne:
Can we add a test for this? Maybe stick install it in a temporary directory and check it runs?

lcw:
> Can we add a test for this? Maybe stick install it in a temporary directory and check it runs?

I think we should add a test for it.

Should the script go in a `bin/` directory instead of `contrib/` as I see it being critical to using the JLL versions of MPI?  Also, the docs (e.g., <https://juliaparallel.github.io/MPI.jl/latest/examples/01-hello/>) could be updated to use this instead since users will not have direct access to `mpiexec` for the JLL'd MPIs.

contrib/mpiexecjl
```diff
@@ -0,0 +1,55 @@
...
+                val = split(f, '=')[2]
+            end
+
+            # Drop this value from our ARGS
+            filter!(x -> x != f, ARGS)
+            return (true, String(val))
+        end
+    end
+    return (false, val)
+end
```
simonbyrne, line 38:
```suggestion
function extract_flag(ARGS, flag, val = "")
    flag_args = String[]
    other_args = String[]
    for f in ARGS
        if startswith(f, flag)
            push!(flag_args, flag)
        else
            push!(other_args, flag)
        end
    end
    return (flag_args, other_args)
end
```

contrib/mpiexecjl
```diff
@@ -0,0 +1,55 @@
...
+
+_ARGS = deepcopy(ARGS)
+project, project_dir = extract_flag!(_ARGS, "--project")
```
simonbyrne, line 41:
```suggestion
project_args, _ARGS = extract_flag(_ARGS, "--project")
```

contrib/mpiexecjl
```diff
@@ -0,0 +1,55 @@
...
+    mpiexec() do exe
+        run(`$exe $ARGS`)
+    end
+    """
+
+if project
+    run(`$(joinpath(Sys.BINDIR, Base.julia_exename())) --color=yes --startup-file=no -q --compile=min -O0 --project=$(project_dir) -e $(script) -- $(_ARGS)`)
+else
+    run(`$(joinpath(Sys.BINDIR, Base.julia_exename())) --color=yes --startup-file=no -q --compile=min -O0 -e $(script) -- $(_ARGS)`)
+end
```
simonbyrne, line 55:
```suggestion
run(`$(joinpath(Sys.BINDIR, Base.julia_exename())) --color=yes --startup-file=no -q --compile=min -O0 $(project_args) -e $(script) -- $(_ARGS)`)
```

contrib/mpiexecjl
giordano: Thanks!

giordano:
I've moved the script to `bin/`, used it in the examples in the documentation and added a test.

I don't have a strong opinion about exporting `install_mpiexecjl`, I didn't want to pollute the scope where `MPI` is used, but it's unlikely to cause conflicts.

simonbyrne:
Very nice, thanks!

bin/mpiexecjl
```diff
@@ -0,0 +1,47 @@
...
+function extract_flag(ARGS, flag, val = "")
+    flag_args = String[]
+    other_args = String[]
+    for f in ARGS
+        if startswith(flag, f)
+            push!(flag_args, f)
+        else
+            push!(other_args, f)
+        end
+    end
```
charleskawczynski, line 33:
```suggestion
    flag_args = filter(f->startswith(flag, f), ARGS)
    other_args = filter(f->!startswith(flag, f), ARGS)
```
Is it better to use `filter` here?

simonbyrne: I guess it requires two passes over the ARGS, but that's probably not a big deal. Would be nice if there was an easy way to do this.

giordano:
I was wondering if it makes sense to have a shell script instead of a Julia one, since this is very simple.  We use a Julia process to run a Julia process to run `mpiexec` to run the jobs: having a shell script at the top would reduce the number of Julia processes involved.

bin/mpiexecjl
charleskawczynski: True

giordano:
This should be a POSIX-compliant script effectively equivalent to the Julia one currently in the pull request:
```sh
#!/bin/sh

PROJECT_ARGS=()
OTHER_ARGS=()
for arg in ${@}; do
    case "${arg}" in
        --project*)
            PROJECT_ARGS+=("${arg}")
            ;;
        *)
            OTHER_ARGS+=("${arg}")
            ;;
    esac
done

SCRIPT='
using MPI
ENV["JULIA_PROJECT"] = dirname(Base.active_project())
mpiexec(exe -> run(`$exe $ARGS`))'

julia "${PROJECT_ARGS[@]}" --color=yes --startup-file=no -q --compile=min -O0 -e "${SCRIPT}" -- "${OTHER_ARGS[@]}"
```

simonbyrne:
I don't think you can use arrays in POSIX shells (at least https://www.shellcheck.net/ doesn't think so).

I think this should work (it only lets you use one `--project` option, but I think that's fine)
```sh
#!/bin/sh

for arg; do
    shift    
    case "$arg" in
        --project | --project=*)
            PROJECT_ARG="$arg"
            ;;
        *)
            set -- "$@" "$arg"
            ;;
    esac
done

SCRIPT='
using MPI
ENV["JULIA_PROJECT"] = dirname(Base.active_project())
mpiexec(exe -> run(`$exe $ARGS`))'

case "$PROJECT_ARG" in
    --project*)        
        julia "$PROJECT_ARG" --color=yes --startup-file=no -q --compile=min -O0 -e "$SCRIPT" -- "$@"
        ;;
    *)
        julia --color=yes --startup-file=no -q --compile=min -O0 -e "$SCRIPT" -- "$@"
        ;;
esac
```
```

**Comment by giordano** [comment]
> I don't think you can use arrays in POSIX shells (at least shellcheck.net doesn't think so).

Right :persevere:   That's why I said "should" :sweat_smile:   Thanks for spellcheck.net, I didn't know that!

I now pushed the version you suggested, after also testing it locally with [`dash`](https://wiki.archlinux.org/index.php/Dash), which is the simplest POSIX-compliant shell I know.

**Comment by simonbyrne** [comment]
Could we add help text somehow, without stopping users from accessing the help of the underlying `mpiexec`?

**Comment by simonbyrne** [comment]
I guess calling it with no arguments would otherwise be an error?

File: bin/mpiexecjl [hunk 6]
```diff
@@ -0,0 +1,46 @@
...
+# Command line utility to call the `mpiexec` binary used by the `MPI.jl` version
+# in the given Julia project.  It has the same syntax as the `mpiexec` binary
+# that would be called, with the additional `--project=...` flag to select a
+# different Julia project.
+#
+# Examples of usage (the MPI flags available depend on the MPI implementation
+# called):
+#
+#   $ mpiexecjl --version
+#   $ mpiexecjl -n 40 mpi-script.jl
```
simonbyrne, line 17:
```suggestion
#   $ mpiexecjl -n 40 julia mpi-script.jl
```
to make clear that `mpiexecjl` doesn't start julia itself.

giordano:
> Could we add help text somehow, without stopping users from accessing the help of the underlying mpiexec?

I've also been wondering how to provide some usage information, in addition to the help of the underlying `mpiexec`.  So now I'm printing the help message when `-h` or `--help` are passed as options.  When no non-project arguments are passed, an error message is printed, in addition to the usage text.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,18 @@
...
+    tags: '*'
+  pull_request:
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: sudo apt install shellcheck
```
giordano, line 16: Maybe need to ignore the "error" with https://github.com/koalaman/shellcheck/wiki/Ignore?

simonbyrne: I added a shellcheck disable to the script.

bin/mpiexecjl
```diff
@@ -0,0 +1,67 @@
...
+mpiexec(exe -> run(`$exe $ARGS`))'
+
+case "${PROJECT_ARG}" in
+    --project*)
+        julia "${PROJECT_ARG}" --color=yes --startup-file=no -q --compile=min -O0 -e "${SCRIPT}" -- "${@}"
+        ;;
+    *)
+        julia --color=yes --startup-file=no -q --compile=min -O0 -e "${SCRIPT}" -- "${@}"
+        ;;
+esac
```
simonbyrne, line 67: I realized this can be simplified to
```suggestion
if [ -n "${PROJECT_ARG}" ]; then
    julia "${PROJECT_ARG}" --color=yes --startup-file=no -q --compile=min -O0 -e "${SCRIPT}" -- "${@}"
else
    julia --color=yes --startup-file=no -q --compile=min -O0 -e "${SCRIPT}" -- "${@}"
fi
```

bin/mpiexecjl
```diff
@@ -0,0 +1,67 @@
...
+if [ -z "${*}" ]; then
+    echo "ERROR: no arguments specified." 1>&2
+    echo
+    usage
+    exit 1
+fi
+
+# shellcheck disable=SC2016
+SCRIPT='
+using MPI
```
simonbyrne, line 56: Do we want to throw an informative error message if this fails? Currently it shows
```
ERROR: ArgumentError: Package MPI not found in current path:
- Run `import Pkg; Pkg.add("MPI")` to install the MPI package.

Stacktrace:
 [1] require(::Module, ::Symbol) at ./loading.jl:887
```

giordano: That's a good idea, but I don't know what to print to be more informative.

simonbyrne:
Thanks!

PR #391, opened 2020-05-22
Simplifiy tests infrastructure

giordano:
* remove the `runtests` function
* remove the `nfail` variable cruft
* generate a testset for each included file
* do a dummy `@test true` for each test in the loop: if the process errors out
the testset would error out as well

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -14,36 +14,26 @@ end
 
 args = Base.shell_split(get(ENV, "JULIA_MPIEXEC_TEST_ARGS", ""))
 
-function runtests()
-    nprocs = clamp(Sys.CPU_THREADS, 2, 4)
-    testdir = dirname(@__FILE__)
-    istest(f) = endswith(f, ".jl") && startswith(f, "test_")
-    testfiles = sort(filter(istest, readdir(testdir)))
+nprocs = clamp(Sys.CPU_THREADS, 2, 4)
+testdir = @__DIR__
+istest(f) = endswith(f, ".jl") && startswith(f, "test_")
+testfiles = sort(filter(istest, readdir(testdir)))
 
-    nfail = 0
-    printstyled("Running MPI.jl tests\n"; color=:white)
-
-    for f in testfiles
-        mpiexec() do cmd
-            cmd = `$cmd $args`
-            if f == "test_spawn.jl"
-                run(`$cmd -n 1 $(Base.julia_cmd()) $(joinpath(testdir, f))`)
-            elseif f == "test_threads.jl"
-                withenv("JULIA_NUM_THREAD" => "4") do
-                    run(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`)
-                end
-            elseif f == "test_error.jl"
-                r = run(ignorestatus(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`))
-                @test !success(r)
-            else
+@testset "$f" for f in testfiles
+    mpiexec() do cmd
+        cmd = `$cmd $args`
+        if f == "test_spawn.jl"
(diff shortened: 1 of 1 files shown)


test/runtests.jl
```diff
@@ -14,33 +14,24 @@ end
...
+            withenv("JULIA_NUM_THREAD" => "4") do
+                run(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`; wait = true)
             end
+        else
+            run(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`; wait = true)
         end
-        Base.with_output_color(:green,stdout) do io
-            println(io,"\tSUCCESS: $f")
-        end
+        wait(p)
```
simonbyrne, line 34: isn't this redundant with `wait=true`?

test/runtests.jl
```diff
@@ -14,33 +14,24 @@ end
...
-            cmd = `$cmd $args`
-            if f == "test_spawn.jl"
-                run(`$cmd -n 1 $(Base.julia_cmd()) $(joinpath(testdir, f))`)
-            elseif f == "test_threads.jl"
-                withenv("JULIA_NUM_THREAD" => "4") do
-                    run(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`)
-                end
-            else
-                run(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`)
+@testset "$f" for f in testfiles
```
simonbyrne, line 22: is the `@testset` worth it? each one only has one test.

test/runtests.jl
```diff
@@ -14,33 +14,24 @@ end
...
+                run(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`; wait = true)
             end
+        else
+            run(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`; wait = true)
         end
-        Base.with_output_color(:green,stdout) do io
-            println(io,"\tSUCCESS: $f")
-        end
+        wait(p)
+        @test success(p)
```
giordano, line 35: The process `p` is run asynchronously, which has the side effect of making harder to see why it's failing.  As an (probably better) alternative, the process can be run with `wait = true` and here we can simply have
```suggestion
        @test true
```
If the process fails, `run` will throw an error and the tests will stop anyway

test/runtests.jl
giordano: In the REPL ``run(`...`; wait=true)`` returns immediately, but in a script it does seem to wait without this.  But see my comment below about not running the processes asynchronously

test/runtests.jl
giordano: It's just a way to get the names of the tests printed in a consistent fashion using the existing testing framework

test/runtests.jl
simonbyrne: > In the REPL `` run(`...`; wait=true) `` returns immediately

I can't replicate this:
```
julia> @time run(`sleep 3`; wait=true)
  3.003904 seconds (50 allocations: 1.906 KiB)
Process(`sleep 3`, ProcessExited(0))
```

test/runtests.jl
simonbyrne: fair enough.

simonbyrne:
I'm not sure I understand the `wait=true` problem.

This now has conflicts after #390.

test/runtests.jl
giordano: I looked at my REPL history: I was testing
```julia
run(`sleep 10`; wait = false)
```
:expressionless:

giordano:
Rebased on `master`.  I've just left the processes synchronous.

simonbyrne:
Thanks!

Issue #394, opened 2020-06-04
2 minor typos in `collective.jl`


amartinhuertas:
Dear `MPI.jl` developers,

I am not wrong, I guess that in the following two lines of code:

https://github.com/JuliaParallel/MPI.jl/blob/master/src/collective.jl#L853
https://github.com/JuliaParallel/MPI.jl/blob/master/src/collective.jl#L856

`sendbuf` should actually be `buf`. 

I am indeed obtaining `UndefVarError: sendbuf not defined` when calling these functions.

I can do a PR and fix it if u find it useful.

@amartinhuertas

simonbyrne:
Yes, you are correct. A PR would be most welcome!

amartinhuertas:
> Yes, you are correct. A PR would be most welcome!

I already triggered the PR! More details at https://github.com/JuliaParallel/MPI.jl/pull/396

simonbyrne:
Fixed by #396.

PR #396, opened 2020-06-05
Several fixes to `Exscan!`

amartinhuertas:
* Fixed several issues related to the Exscan! API and implementation (one ambiguity, double call to `Op(opfunc, eltype(buf))`, together with the typos described in
  https://github.com/JuliaParallel/MPI.jl/issues/394

* Added two tests to test_exscan.jl in order to stress two untested
  overloaded variants of Exscan.


amartinhuertas:
You are welcome!

PR #405, opened 2020-07-07
allow declaration of custom operations at precompilation

### LINKED ISSUE #404: Avoid closures for custom operators
To support custom operators for PowerPC and ARM, we will need to avoid using closures as these are not supported (https://github.com/JuliaLang/julia/pull/34326 and https://github.com/JuliaLang/julia/issues/32154). There might be performance advantages on other platforms as well.

We currently skip the relevant tests on these platforms (#353)

Some options: 
- convert `Op(f, T=Any)` to a generated function
- write a macro that allows us to "declare" custom operator/type combinations, i.e.
   ```julia
   @Op f T
   ```
   would define a non-closure `cfunction` for `Op(::typeof(f),::Type{T})`

simonbyrne:
Fixes #404. 

needs tests.

@vchuravy would this work on PPC? We should set up a CI build now that Travis supports PPC.


src/operators.jl
```diff
@@ -74,16 +83,56 @@ function Op(f, T=Any; iscommutative=false)
...
+macro declareOp(f, T, iscommutative=false)
+    opwrap = gensym(:opwrap) # we need to manually gensym for use with `@cfunction` macro
+    quote
+        if !Base.issingletontype(typeof($(esc(f))))
+            error("@declareOp macro can only be used with instances of singleton types")
+        end        
+        const op = Op(OP_NULL.val, nothing)
+        const $(esc(opwrap)) = OpWrapper{typeof($(esc(f))),$(esc(T))}($(esc(f)))
+        function initop()
+            fptr = @cfunction($opwrap, Cvoid, (Ptr{Cvoid}, Ptr{Cvoid}, Ptr{Cint}, Ptr{MPI_Datatype}))
```
vchuravy, line 125: Yes that should work with PPC. It is the dynamic version of `@cfunction` that requires an LLVM feature that doesn't work on PPC.

simonbyrne:
Is there any way we could do this without a macro, e.g. via a generated function? I don't see how because of the `@cfunction`, but it would be nice.

sloede:
Would this fix #404 also for the M1? If yes, is there anything I can do to help getting it merged?

mikesha2:
Also interested in this, since a few packages I use depend on MPI.jl

PR #406, opened 2020-07-07
use Spack to install system MPI

simonbyrne:


giordano:
Can this be closed or we still want an MPI built with Spack?

PR #408, opened 2020-07-17
try github actions

simonbyrne:


.github/workflows/tests.yml
```diff
@@ -18,6 +18,7 @@ jobs:
 
     env:
       MPICH_INTERFACE_HOSTNAME: localhost
+      JULIA_MPIEXEC_TEST_ARGS: "-hosts localhost"
```
simonbyrne, line 21: @jakebolewski this seems to make it work

jakebolewski: ok good to know, we'll have to have to add something similar to clima

jakebolewski: probably needed both to avoid the fallback behavior :-/

jakebolewski:
@simonbyrne I used `-host` (part of the standard) and not `-hosts`, not sure if that matters

simonbyrne:
ah, good to know. This wasn't the cause, but I updated it.

PR #421, opened 2020-09-03
Add missing get/set MPI.File info functions

jakebolewski:


src/io.jl
```diff
@@ -71,6 +71,36 @@ function close(file::FileHandle)
...
+"""
+    MPI.File.get_info(file::FileHandle)
+
+Returns the hints for a file that are actually being used by MPI.
+
+# External links
+$(_doc_external("MPI_File_get_info"))
+"""
+function get_info(file::FileHandle)
+    file_info = Info(init=true)
```
simonbyrne, line 84: Should the `Info` object be initialized? The spec suggests that the returner creates a new object:
https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node315.htm#Node315

simonbyrne:
Also looks like you pushed this to master already? I'll revert that for now.

src/io.jl
jakebolewski: Good catch.  It wasn't clear if the caller or callee allocated the object from the mpich api docs.

jakebolewski:
Whoops, that was a mistake.  It seems like openmpi doesn't fill in the tested reserved kwargs by default as well.

src/io.jl
```diff
@@ -71,6 +71,36 @@ function close(file::FileHandle)
...
+
+Returns the hints for a file that are actually being used by MPI.
+
+# External links
+$(_doc_external("MPI_File_get_info"))
+"""
+function get_info(file::FileHandle)
+    file_info = Info(init=false)
+    @mpichk ccall((:MPI_File_get_info, libmpi), Cint,
+        (MPI_File, Ptr{MPI_Info}), file, file_info)
```
simonbyrne, line 86: will also need to attach finalizer and increment reference counter.

jakebolewski: ok done

test/test_io.jl
```diff
@@ -45,3 +45,39 @@ data = zeros(Int64, 1)
...
+    fh_info = MPI.File.get_info(fh)
+    @test parse(Int, fh_info[:cb_buffer_size]) == 33554432
+
+elseif MPI.MPI_LIBRARY == MPI.OpenMPI
+
+    # OpenMPI does not set any MPI-IO hints by default
+    @test length(keys(fh_info)) == 0
+
+    # Test that default info hints on mpi-io implementation are present
+    @test parse(Int, fh_info[:coll_write_bufsize]) > 0
```
simonbyrne, line 77: This is causing the error:
https://travis-ci.com/github/JuliaParallel/MPI.jl/jobs/381341367#L6871

jakebolewski: I think i need to play around with OpenMPI's MPI IO implementation some more to understand what the defaults are across different versions.

PR #429, opened 2020-11-03
Allow to pass --compiled-modules flag in mpiexecjl

mfherbst:
I found that sometimes it is quite useful to disable precompilation when using mpiexecjl. My use case is when using mpiexecjl in Travis tests to test an MPI.jl-based code in parallel. This adds a flag to the script to do that which is passed on to julia and not mpiexec.

More generally this of course raises the question what to do with the other flags (e.g. sysimage, threads, procs, ...) could also be useful to be passed onto julia and not mpiexec.

### DIFF
--- a/bin/mpiexecjl
+++ b/bin/mpiexecjl
@@ -8,7 +8,7 @@
 # Command line utility to call the `mpiexec` binary used by the `MPI.jl` version
 # in the given Julia project.  It has the same syntax as the `mpiexec` binary
 # that would be called, with the additional `--project=...` flag to select a
-# different Julia project.
+# different Julia project and `--compiled-modules=...` to disable module precompilation.
 #
 # Examples of usage (the MPI flags available depend on the MPI implementation
 # called):
@@ -20,7 +20,7 @@
 ### Code:
 
 usage () {
-    echo "Usage: ${0} [--project=...] MPIEXEC_ARGUMENTS..."
+    echo "Usage: ${0} [--project=...] [--compiled-modules=...] MPIEXEC_ARGUMENTS..."
     echo "Call the mpiexec binary in the Julia environment specified by the --project option."
     echo "If no project is specified, the MPI associated with the global Julia environment will be used."
     echo "All other arguments are forwarded to mpiexec."
@@ -31,6 +31,9 @@ for arg; do
     case "${arg}" in
         --project | --project=*)
             PROJECT_ARG="${arg}"
+            ;;
+        --compiled-modules=*)
+            COMPILED_ARG="${arg}"
             ;;
         -h | --help)
             usage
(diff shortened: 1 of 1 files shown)


simonbyrne:
Note that this flag only affects the julia session that runs `mpiexec`, not the one which is actually doing the computational work. Is that what you intended?

mfherbst:
Ah yes of course you're right, I completely missed that, sorry. In that case that PR is obsolete from my point of view. Sorry about that ... I've been trying to get MPI to work in our CI in the past hour and I must have been blind not to see that. Thanks for your remark.

Feel free to close this if you want.

simonbyrne:
Can you link to where you're having problem?

mfherbst:
Sure if you want to take a look. So the current build is https://travis-ci.com/github/JuliaMolSim/DFTK.jl/jobs/427158671 and the current commit is https://github.com/JuliaMolSim/DFTK.jl/tree/b2d28f642d9edbb9c800bca70ed8726da405b00c.

The issue is basically that we have quite a number of optional dependencies and dependencies just for testing and if I run Pkg.test() typically not yet all packages are installed and precompiled. Therefore I run into the usual issues with MPI and package installation / precompilation. I now work around it by doing first a short "warmup" test in the MPI-specific travis job and afterwards do the "real" MPI test. Let's see if that does it.

If you have a better idea, I'm happy to hear it.

simonbyrne:
The issue appears to be in the ErrorfreeArithmetic package:
https://travis-ci.com/github/JuliaMolSim/DFTK.jl/jobs/427158671#L2127-L2129

It shows up in both the MPI and non-MPI tests.

simonbyrne:
> The issue is basically that we have quite a number of optional dependencies and dependencies just for testing and if I run Pkg.test() typically not yet all packages are installed and precompiled.

You could adding something like
```julia
using Pkg
pkg"precompile"
```
to the top of your `runtests.jl`.

mfherbst:
Thanks for taking a look. In fact the error I get now is completely different from what I had before (which was directly related to precompilation and package installation and occurred much earlier), so the trick I introduced in the recent commit (to run a stripped-down test suite) did help.

> It shows up in both the MPI and non-MPI tests.

Really? Maybe I missed it in the logs, but I could not find that problem in the non-MPI tests in fact.

> You could adding something like

Sure. I tried that before having the stripped-down test suite, but that did not cut it, because that will trigger precompilation in parallel on all processes and it the issues during the installation still happened. I also tried with an MPI barrier and only precompiling on master without luck.

In fact I don't think the current error is still related to precompilation. For the moment I think I will just disable the relevant test to try and get a subset to pass. For us MPI is anyway experimental at the moment and not yet supported by all features of our code.

mfherbst:
Just to let you know: Disabling the test "resolved" the issue and our tests pass, which is good enough for me at the moment. Thanks for your time @simonbyrne!

simonbyrne:
Okay, i'll close this for now.

Issue #430, opened 2020-11-04
Yell if run under a different MPI


antoine-levitt:
I was initially confused by the fact that I ran my julia/MPI script like `mpiexec julia test.jl` and everything worked fine... except that it used only one process. Is it possible to raise an error/warning if the executing MPI doesn't match MPI.jl's?

simonbyrne:
Not that I know of, unfortunately.

simonbyrne:
We might be able to come up with some ad-hoc rules, e.g. if the environment variables `MPI_LOCALNRANKS` (set by the MPICH launcher) or `OMPI_COMM_WORLD_SIZE` (set by the Open MPI launcher) is different than `MPI.Comm_rank(MPI.COMM_WORLD)`, then throw an error?

antoine-levitt:
That'd be cool : it doesn't have to be 100% foolproof, just a friendly hint that you're doing something wrong. I can make a PR with that. Is that still a valid usage (and therefore just a warning) or an error?

simonbyrne:
That would be great thanks. We do something similar when people appear to be using the default MPI binary on a cluster:
https://github.com/JuliaParallel/MPI.jl/blob/20424263680e0c1339ea1f72814ad6cf25cef7b3/deps/build.jl#L158-L161
You can copy that.

giordano:
The [`mpiexecjl`](https://juliaparallel.github.io/MPI.jl/v0.16/configuration/#Julia-wrapper-for-mpiexec) wrapper should also help with this.

PR #431, opened 2020-11-05
Warn if wrong MPI

antoine-levitt:
Fix https://github.com/JuliaParallel/MPI.jl/issues/430. If somebody has a non-OpenMPI/MPICH implementation lying around and can post `mpiexec julia -e 'display ENV'` we can possibly support more.

### DIFF
--- a/src/environment.jl
+++ b/src/environment.jl
@@ -100,10 +100,10 @@ function Init()
     REFCOUNT[] = 1
     atexit(refcount_dec)
 
-    _warn_if_wrong_mpi()
     for f in mpi_init_hooks
         f()
     end
+    _warn_if_wrong_mpi()
 end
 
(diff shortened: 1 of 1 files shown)


src/environment.jl
```diff
@@ -85,6 +100,7 @@ function Init()
     REFCOUNT[] = 1
     atexit(refcount_dec)
 
+    _warn_if_wrong_mpi()
```
simonbyrne, line 103: Needs to be moved to after the `mpi_init_hooks`, as that is when `MPI.COMM_WORLD` is set for Open MPI.

antoine-levitt: Good catch, thanks!

simonbyrne:
Thanks!

PR #441, opened 2020-11-13
Allow libmpi.so to be in lib64

calebwin:


calebwin:
I have `libmpi.so` stored in an `openmpi/lib64`. Is there really no way for MPI.jl to pick up the system-provided MPI without this PR? Would this PR be sufficient for it to pick up `openmpi/lib64/libmpi`?

simonbyrne:
Thanks!

calebwin:
Welp, I just realized I _can_ actually specify the full path to libmpi, eliminating my original need for this PR:
```
julia-1.5.3/bin/julia --project -e 'ENV["JULIA_MPIEXEC"]="srun"; ENV["JULIA_MPI_LIBRARY"]="/opt/amazon/openmpi/lib64/libmpi"; using Pkg; Pkg.build("MPI"; verbose=true)'
```

simonbyrne:
That works too, but this change makes sense in any case.

francispoulin:
I think I might be having a similar problem.  Instead of setting the `ENV`'s, as in the example above, I am using `export` to set the different variables.  This hs worked for me in the past so I don't think this is a problem.

Below you can see what `JULIA_MPI_LIBRARY` is defined as and below that the file actually exists.

However, when I try building MPI I get a message saying that `libmpi` is not found.  Is there a subtlty here that I am missing?

```
[fpoulin@cedar1 scratch]$ echo $JULIA_MPI_LIBRARY
/cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/CUDA/intel2020/cuda11.0/openmpi/4.0.3/lib64/libmpi.so
[fpoulin@cedar1 scratch]$ ls /cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/CUDA/intel2020/cuda11.0/openmpi/4.0.3/lib64/libmpi.so
/cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/CUDA/intel2020/cuda11.0/openmpi/4.0.3/lib64/libmpi.so
```

simonbyrne:
what is the full output of `Pkg.build(verbose=true)`?

francispoulin:
sorry for not including the output before.  Here you go


```
(@v1.6) pkg> build MPI
    Building MPI → `~/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/714909d6b8dd4287bdf1a21ae1ffe3da1a873ace/build.log`
ERROR: Error building `MPI`: 
[ Info: using system MPI
ERROR: LoadError: libmpi could not be found
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] top-level scope
   @ ~/.julia/packages/MPI/E3Wer/deps/build.jl:76
 [3] include(fname::String)
   @ Base.MainInclude ./client.jl:444
 [4] top-level scope
   @ none:5
in expression starting at /home/fpoulin/.julia/packages/MPI/E3Wer/deps/build.jl:64
```

simonbyrne:
Can you look through the logic in `deps/build.jl` and see what isn't working?

francispoulin:
I will most certainly try

francispoulin:
The problem happens [here](https://github.com/JuliaParallel/MPI.jl/blob/a2920684c613317cff86f6a06ac1ffaf85ba68b0/deps/build.jl#L74) .

At this stage we are using `find_library` to find `libmpi.so`.  The first argument is
``` 
library = "/cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/CUDA/intel2020/cuda11.0/openmpi/4.0.3/lib64/libmpi.so"
```

It's looking in the following paths
 ```
"/cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/CUDA/intel2020/cuda11.0/openmpi/4.0.3/lib"

"/cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/CUDA/intel2020/cuda11.0/openmpi/4.0.3/lib64"
```

I checked and both paths have a file called `libmpi.so` however when I execute that file llne by line, at this  point I get nothing, which is what causes the error
```
julia> find_library(library, [joinpath(path, "lib"), joinpath(path, "lib64")])
""
```

When I try `find_library` on each separately, each also returns "".  

I have never used `find_library` before but it seems to not be finding the libraries even though there are two to choose from.

Any idea why it might be failing?

**Comment by francispoulin** [comment]
I looked around and see that if we fail to find the library it's not because the file isn't necessarily there.  It's possible that it's one of the files needed by the library that be found.  

The following [site ](https://discourse.julialang.org/t/libdl-dlopen-or-libdl-find-library-does-not-find-a-shared-library/55161/5) suggesting using `strace`, but I don't know how to do that.

It also suggested using `dsopen`, and I tried that below and see that it failed to open, but I checked the path and it is correct.  However now it says that `libevent-2.1.so.6` cannot be found.  

I agree that it's not in either of these directories.  Do I need to find this library and add it to my path as well?

```
julia> Libdl.dlopen("/cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/CUDA/intel2020/cuda11.0/openmpi/4.0.3/lib/libmpi.so")
ERROR: could not load library "/cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/CUDA/intel2020/cuda11.0/openmpi/4.0.3/lib/libmpi.so"
libevent-2.1.so.6: cannot open shared object file: No such file or directory
Stacktrace:
 [1] dlopen(s::String, flags::UInt32; throw_error::Bool)
   @ Base.Libc.Libdl ./libdl.jl:114
 [2] dlopen (repeats 2 times)
   @ ./libdl.jl:114 [inlined]
 [3] top-level scope
   @ REPL[64]:1
```

**Comment by simonbyrne** [comment]
> However now it says that `libevent-2.1.so.6` cannot be found.

That seems to be the problem. From what it looks like, you're on a HPC cluster: typically they have a module system or something similar that will set the necessary environment variables for the various pieces of software that are available on the cluster.

**Comment by francispoulin** [comment]
Thank you.  That is a very good observation.

I am on an HPC cluster and I have found a module called `libevent/2.1.11` and have loaded it.  Unfortunately, when I try to build it still says that it can't find libmpi.so.  

Furthermore, when I try 
```
Libdl.dlopen("/cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/CUDA/intel2020/cuda11.0/openmpi/4.0.3/lib64/libmpi.so")
```
to see if it can open the mpi library it still complains about libevent-2.1.so.6. 

Do I need to find out where the library is located and then add that into my path somewhere?

**Comment by simonbyrne** [comment]
Do they have a module for the MPI library?

**Comment by simonbyrne** [comment]
I also seem to recall that the intel tools require you to run some script before using them. You might need to ask your system administrators.

**Comment by francispoulin** [comment]
> Do they have a module for the MPI library?

Yes and I have added it.  You can see below what I have currently loaded

```
Currently Loaded Modules:
  1) CCconfig             4) imkl/2020.1.217  (math)   7) cudacore/.11.0.2 (H,t)  10) ucx/1.8.0             13) libevent/2.1.11
  2) gentoo/2020    (S)   5) intel/2020.1.217 (t)      8) cuda/11.0        (t)    11) libfabric/1.10.1
  3) gcccore/.9.3.0 (H)   6) StdEnv/2020      (S)      9) gdrcopy/2.1             12) openmpi/4.0.3    (m)
```

**Comment by francispoulin** [comment]
> I also seem to recall that the intel tools require you to run some script before using them. You might need to ask your system administrators.

Ah, I haven't heard about that but I will ask them on Monday.  Thanks a lot @simonbyrne for the help and I will let you know what I find out.

**Comment by simonbyrne** [comment]
you have `openmpi/4.0.3` loaded: you shouldn't need to specify the library path?

**Comment by francispoulin** [comment]
> you have `openmpi/4.0.3` loaded: you shouldn't need to specify the library path?

Maybe?  Does that mean that I don't need to specify path and library?

**Comment by simonbyrne** [comment]
The module system should set the appropriate environment variables so that you don't need to specify the path

**Comment by francispoulin** [comment]
Good point.  I see that `$PATH` seems to point to the right library.  But when I tried the build script this didn't appear in the path, and why `find_library` failed, I think.  

Should I try and pass this as my path? If yes is there a special syntax for `$PATH` because of the $?

**Comment by francispoulin** [comment]
Sorry to bother you again @simonbyrne but should the build scrpt be aware of my `$PATH`?  At the moment it is not.

**Comment by francispoulin** [comment]
Also, when I try to build MPI today I see there is a probem with `DocStringExtensions`.  Is this a known problem?

```
(Benchmarks) pkg> build MPI
    Building MPI → `~/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/6e8c30afdcbb6167cf5d470b6333f4db01cc366f/build.log`
┌ Warning: Could not use exact versions of packages in manifest, re-resolving
└ @ Pkg.Operations /tmp/ebuser/avx2/Julia/1.6.0/GCCcore-9.3.0/julia-1.6.0/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:1524
ERROR: Unsatisfiable requirements detected for package DocStringExtensions [ffbed154]:
 DocStringExtensions [ffbed154] log:
 ├─DocStringExtensions [ffbed154] has no known versions!
 └─restricted to versions 0.8 by an explicit requirement — no versions left
```

**Comment by simonbyrne** [comment]
I don't think it uses `PATH`, but it does use `LD_LIBRARY_PATH`: usually the module system will add the necessary library paths to that. What does `printenv LD_LIBRARY_PATH` give you?

**Comment by francispoulin** [comment]
When I try this in linux I get the following:

```
$ printenv LD_LIBRARY_PATH
/cvmfs/soft.computecanada.ca/easybuild/software/2020/Core/cudacore/11.0.2/lib64:
```

What's strange is that today, when I try copying and pasting the first part of the build script it actually seems to find the library and so `libmpi` is non-empty.  So I guess this is working now, for reasons I don't understand.

Now the problem seems to be with `DocStringExtensions`, as I mentioned above.  Do you know where this might come from?

simonbyrne:
> Now the problem seems to be with `DocStringExtensions`, as I mentioned above. Do you know where this might come from?

No idea. Try `Pkg.update()`?

francispoulin:
Will do.  

One more question. If I wanted to specify `$Folder` in the `ENV` environment, do you happen to know how to do that?  I tried `ENV["$Folder"]` and that didn't seem to work.

simonbyrne:
You don't need the `$`: that just signifies interpolation (in both the shell and julia), i.e. `echo $PATH` is interpolating `PATH` into a string, and then printing it.

To set environment variables you can either do them at the shell (`export Folder=...` in bash) or via `ENV["Folder"]` in Julia.

francispoulin:
Great to know and thank you again @simonbyrne !

simonbyrne:
No problem. In future, it would be better to ask on discourse for these types of questions (where more people will typically look at it).

francispoulin:
I will try and do that, and sorry that I didn't do it this time around.

francispoulin:
Sadly, what I found was working is no longer working.  I've gone to Discourse, in hopes that would be more convenient.

PR #449, opened 2020-12-14
Add new example to docs

rafabench:
Hi,
I've built an example where you want to apply a function to an array of data with job scheduling. This example is inspired on https://www.hpc.ntnu.no/ntnu-hpc-group/vilje/user-guide/software/mpi-and-mpi-io-training-tutorial/basic-mpi/job-queue

### DIFF
--- a/docs/examples/05-job_schedule.jl
+++ b/docs/examples/05-job_schedule.jl
@@ -1,7 +1,7 @@
 # This example demonstrates a job scheduling through adding the
-# number 100 to every component of the vector data. The master
-# assigns one element to each slave to compute the operation.
-# When the worker is finished, the master sends another element
+# number 100 to every component of the vector data. The root
+# assigns one element to each worker to compute the operation.
+# When the worker is finished, the root sends another element
 # until each element is added 100
 # Inspired on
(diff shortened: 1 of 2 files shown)


docs/examples/05-job_schedule.jl
```diff
@@ -0,0 +1,133 @@
+# This example demonstrates a job scheduling through adding the
+# number 100 to every component of the vector data. The master
+# assigns one element to each slave to compute the operation.
```
simonbyrne, line 3:
```suggestion
# assigns one element to each worker to compute the operation.
```
and maybe change `master` to `root` as well (since that is MPI parlance)

docs/examples/05-job_schedule.jl
```diff
@@ -0,0 +1,133 @@
...
+
+using MPI
+
+function job_queue(data,f)
+    MPI.Init()
+
+    comm = MPI.COMM_WORLD
+    rank = MPI.Comm_rank(comm)
+    world_size = MPI.Comm_size(comm)
+    workers = world_size - 1
```
simonbyrne, line 17:
```suggestion
    nworkers = world_size - 1
```
since it is the number of workers, not the set of workers.

simonbyrne:
Thanks!

rafabench:
You're welcome!

PR #450, opened 2020-12-31
Make MPI.jl compatible with profilers that use LD_PRELOAD

lyon-fnal:
See #444. Use of `ccall( (func, lib), ...)` is not compatible with applications that use `LD_PRELOAD` to inject their own code for profiling and tracing. [Darshan](https://www.mcs.anl.gov/research/projects/darshan/) is an example of an application that does this for MPI and HDF5 I/O profiling. 

This PR has changes to explicitly load the `libmpi` shared object with `Libdl.dlopen` in `MPI.__init__`. Most `ccall` statements are changed to not specify the MPI library. There are some`ccall` statements in `implementations.jl` that I left alone because they get called before `MPI.__init__` and all they do is determine MPI version and implementation details. 

I tested on NERSC Cori Haswell nodes. MPI tests pass except for `test_spawn.jl`, but this fails with the released MPI.jl too. I was able to get Darshan to work with this PR. 

Happy New Year too!

PR #451, opened 2021-01-04
use global symbols on Unix to support LD_PRELOAD interception

simonbyrne:
This is an alternative to #450 that only uses global symbols on unix systems.

@lyon-fnal would you be able to try this out?

### DIFF
--- a/src/implementations.jl
+++ b/src/implementations.jl
@@ -1,5 +1,6 @@
 const use_stdcall = startswith(basename(libmpi), "msmpi")
 
+# This is required in addition to __init__() so that we can call library at precompilation time.
 if Sys.isunix()
     Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
 end
--- a/src/implementations.jl
+++ b/src/implementations.jl
@@ -1,5 +1,9 @@
 const use_stdcall = startswith(basename(libmpi), "msmpi")
 
+if Sys.isunix()
+    Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
+end
+
 macro mpicall(expr)
     @assert expr isa Expr && expr.head == :call && expr.args[1] == :ccall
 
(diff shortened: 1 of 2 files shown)


lyon-fnal:
Thanks - I'll try this hopefully later today.

src/implementations.jl
```diff
@@ -1,5 +1,9 @@
 const use_stdcall = startswith(basename(libmpi), "msmpi")
 
+if Sys.isunix()
+    Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
```
lyon-fnal, line 4: Just wondering - In #450 I put the `dlopen` call in the `MPI.__init__` function to ensure it only gets called once. Is this a worry? Perhaps not?

vchuravy: Yeah this change should not be necessary since:

https://github.com/JuliaParallel/MPI.jl/blob/1ebbbd9cf987ecd6061b4cb19cde426d14733e6f/src/MPI.jl#L64

lyon-fnal: So you are loading the library twice ... here and in `MPI.__init__`. Actually, I don't think that hurts anything.

simonbyrne: It is here because this file is also included from `deps/build.jl`, and we evaluate `Get_library_version()` at precompile time (before `__init__` is called).

musm:
Does this make https://github.com/JuliaIO/HDF5.jl/pull/791 unecessary ? 

Also note Mose's reply in https://github.com/JuliaIO/HDF5.jl/pull/791#issuecomment-753362970

simonbyrne:
No, but you could use the same trick (macro wrapper around `ccall`)

lyon-fnal:
Sorry - couldn't get back to this until now. I'm trying it!

lyon-fnal:
It works fine on my simple test!

simonbyrne:
Thanks. @vchuravy any additional thoughts before we merge?

lyon-fnal:
I'm trying a more complex test and with the package compiler.

lyon-fnal:
My complex test worked. I didn't get much info out of Darshan, because I'm not doing collective MPI-IO. But that's not your problem :-) . If you are waiting on me to merge, I think  you can go ahead. Thanks!

simonbyrne:
Thanks @lyon-fnal!

PR #452, opened 2021-02-24
RFC: Define wait(req) to use threadcall

simonbyrne:
Prompted by [this discussion](https://discourse.julialang.org/t/how-to-combine-mpi-non-blocking-isend-irecv-and-julia-tasks/52524/5?u=simonbyrne) I tried to see if I could use [`@threadcall`](https://docs.julialang.org/en/v1/base/multi-threading/#Base.@threadcall) to wait on nonblocking MPI operations inside tasks.

And it seems to work! I've defined this to be a method of `Base.wait`, in that it acts mostly the same way (i.e. will yield until it is complete). Obviously one problem is that it is limited by the size of the libuv thread pool, but I don't think that would be a big issue.

If people are in favor, I can define analogous `waitany`/`waitall`/`waitsome` operations.

cc: @giordano @stevengj @vchuravy @fverdugo

### DIFF
--- a/src/pointtopoint.jl
+++ b/src/pointtopoint.jl
@@ -411,6 +411,19 @@ function Wait!(req::Request)
     stat_ref[]
 end
 
+function Base.wait(req::Request)
+    stat_ref = Ref{Status}(MPI.STATUS_EMPTY)
+    alreadynull = isnull(req)
+    errcode = @threadcall((:MPI_Wait, libmpi), Cint,
+                          (Ptr{MPI_Request}, Ptr{Status}),
+                          req, stat_ref)
+    errcode == MPI_SUCCESS || throw(MPIError(errcode))
+    if !alreadynull
+        req.buffer = nothing
+    end
+    stat_ref[]
+end
+
 """
     (flag, status) = Test!(req::Request)
 
--- a/test/test_waitthread.jl
+++ b/test/test_waitthread.jl
@@ -1,0 +1,28 @@
+using MPI, Test
+
+MPI.Init()
+
+rank = MPI.Comm_rank(MPI.COMM_WORLD)
+
+
+X = zeros(10)
+if rank == 1
+    MPI.Recv!(X,0,2,MPI.COMM_WORLD)
+    MPI.Send(rand(10),0,0,MPI.COMM_WORLD)
+elseif rank == 0
+    X = zeros(10)
+    req = MPI.Irecv!(X,1,0,MPI.COMM_WORLD)
+    task = @async begin
+        wait(req)
+        # success!
+    end
+    yield()
+    
+    # trigger rank 1
+    MPI.Send(zeros(10),1,2,MPI.COMM_WORLD)
+    wait(task)
+    
+    flag,_ = MPI.Test!(req)
+    @test flag
+end
+


stevengj:
Couldn’t waitall be defined as a wait method that takes an array of requests, maybe with a flag argument for waitall vs waitany?

src/pointtopoint.jl
```diff
@@ -411,6 +411,19 @@ function Wait!(req::Request)
     stat_ref[]
 end
 
+function Base.wait(req::Request)
+    stat_ref = Ref{Status}(MPI.STATUS_EMPTY)
+    alreadynull = isnull(req)
+    errcode = @threadcall((:MPI_Wait, libmpi), Cint,
+                          (Ptr{MPI_Request}, Ptr{Status}),
+                          req, stat_ref)
```
simonbyrne, line 419: One issue is that on 32-bit windows we need to use the `stdcall` convention: this is usually handled by the `@mpicall`/`@mpichk` macros, but `@threadcall` doesn't seem to support nonstandard calling conventions.

On the other hand, it passed tests so 🤷 ?

simonbyrne: (alternatively, we could disable this on 32-bit Windows, like we do the custom operator stuff)

vchuravy:
It would be good to have this be configurable, I have to think a bit more
about this, but MPI implementation are free to not be threadsafe (or use
libraries under the hood that are not threadsafe, I suspect that half the
UCX builds are built without support for threads...)

fverdugo:
Nice to see this!

Is this approach composable? In some situations, we need to wait for a precondition (Request 1) in order to start a non-bloking communication that will generate Request 2. It would be nice to be able to compose these two requests in a asynchronous Task (or whatever object) that waits for Request 1, starts the non-blocking communication and waits for Request 2.

simonbyrne:
> It would be good to have this be configurable, I have to think a bit more about this, but MPI implementation are free to not be threadsafe (or use libraries under the hood that are not threadsafe, I suspect that half the UCX builds are built without support for threads...)

Yes, the user would be required to initialize using `MPI.Init_thread(MPI.THREAD_SERIALIZED)` or `MPI.Init_thread(MPI.THREAD_MULTIPLE)`.


> Is this approach composable? In some situations, we need to wait for a precondition (Request 1) in order to start a non-bloking communication that will generate Request 2. It would be nice to be able to compose these two requests in a asynchronous Task (or whatever object) that waits for Request 1, starts the non-blocking communication and waits for Request 2.

Yes, you can do something like:
```
request1 = MPI.Irecv!(...)

commtask = @async begin
    wait(request1)
    request2 = MPI.Send(...)
    wait(request2)
end
# do other work
wait(commtask)
```

Alternatively, it might be better to just use Julia threading for this:
```
request1 = MPI.Irecv!(...)

commtask = Threads.@spawn begin
    MPI.Wait!(request1)
    request2 = MPI.Send(...)
    MPI.Wait!(request2)
end
# do other work
wait(commtask)
```
The main difference is that the first is blocking a libuv thread, the second is blocking a Julia thread. I'm not sure what other implications of this are.

simonbyrne:
After further thought, I don't think this is a good idea. 

I think it would be better to define `wait` as a simple busy loop with `yield`, e.g.
```julia
function Base.wait(req::MPI.Request)
   while !MPI.Test(req)
       yield()
   end
end
```

This should work, since

> If an MPI_TEST that completes a receive is repeatedly called with the same arguments,
and a matching send has been started, then the call will eventually return flag = true, unless
the send is satisfied by another receive. If an MPI_TEST that completes a send is repeatedly
called with the same arguments, and a matching receive has been started, then the call will
eventually return flag = true, unless the receive is satisfied by another send

fverdugo:
@simonbyrne thanks for the new idea.

In my code I am finally doing something like 
```julia
t = @async begin
   while !MPI.Test(req)
       yield()
   end
end
```

and then one can consume task `t` as usual in Julia. Perhaps we don't even need to define the new method for wait. Just add some comment on the docstrings of Test and TestAll regarding on how to use them in combination with `@async`.

simonbyrne:
> Just add some comment on the docstrings of Test and TestAll regarding on how to use them in combination with `@async`.

That's a good idea, would you mind opening a draft PR?

fverdugo:
> That's a good idea, would you mind opening a draft PR?

Yes! I'll do that.

PR #456, opened 2021-03-16
Avoid InitError on Julia 1.3 with system MPI libraries

### LINKED ISSUE #455: InitError on Julia 1.3 with system MPI
For some reason that I don't really understand, MPI.jl 0.17 throws an `InitError` when loaded using Julia 1.3 on Unix (or at least Linux) systems, when using the system MPI libraries.

See [here](https://github.com/jipolanco/PencilArrays.jl/pull/20/checks?check_run_id=2114908429) for an example, that I have also reproduced locally on Fedora 33. I also tested on Julia 1.5, 1.6 and nightly, where the issue does not appear.

The issue seems to be related to #451. The problem is that `ccall` does not seem to find `MPI_Get_library_version`. I managed to workaround the issue by adding an exception for this specific function in the `@mpicall` macro.

My workaround can be seen [here](https://github.com/JuliaParallel/MPI.jl/compare/master...jipolanco:jip/fix-julia1.3), and I can open a PR if this is a proper solution.

Here is a minimum example:

```julia
➜  MPI git:(master) ~/opt/julia-1.3/bin/julia --project -e 'using MPI'
ERROR: InitError: ccall: could not find function MPI_Get_library_version
Stacktrace:
 [1] Get_library_version() at /home/jipolanco/.julia/dev/MPI/src/implementations.jl:36
 [2] __init__deps() at /home/jipolanco/.julia/dev/MPI/deps/deps.jl:11
 [3] __init__() at /home/jipolanco/.julia/dev/MPI/src/MPI.jl:67
 [4] _include_from_serialized(::String, ::Array{Any,1}) at ./loading.jl:692
 [5] _require_from_serialized(::String) at ./loading.jl:743
 [6] _require(::Base.PkgId) at ./loading.jl:1034
 [7] require(::Base.PkgId) at ./loading.jl:922
 [8] require(::Module, ::Symbol) at ./loading.jl:917
during initialization of module MPI
```

jipolanco:
This PR works around #455 by adding an exception when calling `MPI_Get_library_version` via `ccall` on Unix-like systems, when using Julia versions older than 1.5.

Closes #455.

### DIFF
--- a/src/implementations.jl
+++ b/src/implementations.jl
@@ -11,7 +11,7 @@ macro mpicall(expr)
     # On unix systems we call the global symbols to allow for LD_PRELOAD interception
     # It can be emulated in Windows (via Libdl.dllist), but this is not fast.
     if Sys.isunix() && expr.args[2].head == :tuple &&
-            expr.args[2].args[1] != :(:MPI_Get_library_version)
+            (VERSION ≥ v"1.5-" || expr.args[2].args[1] ≠ :(:MPI_Get_library_version))
         expr.args[2] = expr.args[2].args[1]
     end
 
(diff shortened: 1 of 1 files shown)


simonbyrne:
Thanks!

PR #457, opened 2021-03-18
add basic error handler functionality

### LINKED ISSUE #453: Support for error handlers
It looks like MPI as early as 3.0 has error handling and it would be really awesome to have this supported in MPI.jl.

simonbyrne:
Fixes #453.


simonbyrne:
Note that mpi4py sets `MPI_ERRORS_RETURN` on the default communicators:
https://mpi4py.readthedocs.io/en/stable/overview.html#error-handling

simonbyrne:
@kpamnany @vchuravy any thoughts on this?

vchuravy:
So this will no longer kill the process on error? That seems sensible, altough you could still end up in a deadlock.

simonbyrne:
> So this will no longer kill the process on error? That seems sensible, altough you could still end up in a deadlock.

Yes, it will throw an `MPIError` instead.

PR #460, opened 2021-04-04
Add finalizer to `create_struct`

### LINKED ISSUE #459: Sending and receiving custom types causes memory leaks
We ran into a problem where our code was allocating more and more memory the longer the computation ran. It turns out the culprit was sending and receiving arrays with custom `eltypes`.

From my limited understanding, it appears that a new instance of `MPI.Datatype` is created on each `Send`/`Recv!`/`Get_count` and the instances are kept around forever.

The problem can be fixed by manually creating an `MPI.Datatype` instance and constructing `MPI.Buffer`s with that instance before sending/receiving. This does not happen with MPI.jl `≤v0.11.0` (i.e. before `Buffer`s were introduced).

If this is not a bug, perhaps it should be documented somewhere. I'd be happy to help with making a fix or updating the docs.

Here's an MWE:

```julia
# script.jl
using MPI

struct MyType
    i::Int
end

MPI.Initialized() || MPI.Init()

function mwe(n_steps)
    comm = MPI.COMM_WORLD
    nprocs = MPI.Comm_size(comm)
    rank = MPI.Comm_rank(comm)
    recbuffer = [MyType(0)]
    result = 0
    for _ in 1:n_steps
        send_target = mod(rank + 1, nprocs)
        MPI.Send([MyType(1)], send_target, 0, comm)

        rec_target = mod(rank - 1, nprocs)
        status = MPI.Probe(rec_target, 0, comm)
        count = MPI.Get_count(status, MyType)
        MPI.Recv!(recbuffer, rec_target, 0, comm)
        result += recbuffer[1].i
    end
    printstyled("Rank $rank collected $result\n", color=:green)
end

mwe(10_000) # Set this to a high value (e.g. 10_000_000) to run out of memory
```

mtsch:
Adding a finalizer to the `Datatype` in `create_struct` fixes #459 
Please let me know if this is the correct fix.

### DIFF
--- a/src/datatypes.jl
+++ b/src/datatypes.jl
@@ -213,6 +213,7 @@ function create_struct(blocklengths, displacements, types)
                       (Cint, Ptr{Cint}, Ptr{MPI_Aint}, Ptr{MPI_Datatype}, Ptr{MPI_Datatype}),
                       N, blocklengths, displacements, mpi_types, newtype)
     end
+    finalizer(free, newtype)
     return newtype
 end
 


lcw:
Thanks!  This looks like the correct fix to me what do you think @simonbyrne?

simonbyrne:
Thanks, not sure how that one got missed!

joachimbrand:
Thanks a lot for the fix!

Issue #462, opened 2021-04-08
`MPI.Datatype()` creates new instance on every `MPI.Put()`


joachimbrand:
This is a follow up on #459 . After fixing the issue with a memory leakage by installing a finalizer to `create_struct` #460 , we are still having issues with our code exiting by strange errors during `MPI.Put()`, but only after a decent amount of communication, See [joachimbrand/Rimu.jl#47](https://github.com/joachimbrand/Rimu.jl/issues/47). 

We noticed that `MPI.Datatype()` when called repeatedly on the same custom type returns a different value every time (kudos to @mtsch ). Even though the garbage collector can now clean up these custom structs, it still appears wasteful to not reuse the struct in subsequent calls. I suspect that this is the issue causing  [joachimbrand/Rimu.jl#47](https://github.com/joachimbrand/Rimu.jl/issues/47), as the mysterious errors seem to disappear after defining a new method for `MPI.Datatype()` that just returns the value generate by the initial call.

simonbyrne:
Unfortunately, I don't think eval can really be used like this (though I could be wrong).

simonbyrne:
Ah, we really should modify the interface for `Put` to use the new `Buffer` interface we use for point-to-point and collective operations. This will let you pass `Datatype`s manually, or create a `Buffer` and reuse it.

In the long run, it would probably make sense to cache the `Datatype`s somehow, either using a dictionary lookup, declaring datatypes via a macro, or using generated functions.

joachimbrand:
Thanks for the feedback!

Yes, my first thought was to pass a `Datatype` or `Buffer`, but looking at the code of `Put` I saw that that is not easily possible.

joachimbrand:
@simonbyrne Thanks for updating the interface!

I'll try the new interface passing `Buffer`s in a bit (and will report back). Before modifying our code I wanted to run the existing code with the master branch of MPI.jl.

Just running our code without modification (develop branch) [joachimbrand/Rimu.jl#develop](https://github.com/joachimbrand/Rimu.jl/tree/develop) using our standard test script for MPI produces an error that appears to come from the deprecated (old) interface:
```
it092078:Rimu brand$ mpirun -np 2 julia test/script_mpi_minimum.jl 
hello hello!
hello hello!
┌ Info: after barrier 1
│   mpi_rank() = 0
│   mpi_size() = 2
└   Threads.nthreads() = 1
┌ Info: after barrier 1
│   mpi_rank() = 1
[...]

PR #463, opened 2021-04-08
Define new method for `Datatype` to fix issue #462

joachimbrand:
I'm not sure whether this is the correct fix for the issue, but it seems to help with errors we are seeing when using one-sided communication as in [joachimbrand/Rimu.jl#47](https://github.com/joachimbrand/Rimu.jl/issues/47). Any comments are welcome.

### DIFF
--- a/src/datatypes.jl
+++ b/src/datatypes.jl
@@ -308,6 +308,7 @@ function Datatype(::Type{T}; commit=true) where T
     if commit
         commit!(dt)
     end
+    @eval Datatype(::Type{$T}) = $dt
     return dt
 end
 
--- a/test/test_datatype.jl
+++ b/test/test_datatype.jl
@@ -48,6 +48,11 @@ end
     end
 end
 
+@testset "Unique return type" begin
+    dt = MPI.Datatype(Boundary)
+    @test dt ===  MPI.Datatype(Boundary)
+end
+
 struct Boundary2
     a::UInt32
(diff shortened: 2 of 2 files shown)


joachimbrand:
May fix #462

src/datatypes.jl
```diff
@@ -308,6 +308,7 @@ function Datatype(::Type{T}; commit=true) where T
     if commit
         commit!(dt)
     end
+    @eval Datatype(::Type{$T}) = $dt
```
vchuravy, line 311: While technically legal it is not a great idea, you can run into `worldage` issues with this.

if you need caching, I would go through a global `IDDict{Type, Dataype}`.

simonbyrne: Maybe we should just make `Datatype(T)` a generated function?

joachimbrand: Thanks for the comments! I'm still learning about world age, but now I see how this can be a problem. So one would have to instruct users to once call `MPI.Datatype(T)` at top level, in order to be able to take advantage of the new method later. That would not be elegant.

It seems to me that performance should be much better than with dictionary lookup, though, which presumably would have to be done at runtime on every `Put`.

Making `Datatype()` a generated function sounds like a good solution, as I understand that generated functions can cache the generated (and compiled) code.

simonbyrne:
I think this should now be solved by #471.

PR #464, opened 2021-04-08
use Buffers with one-sided ops

simonbyrne:


src/onesided.jl
```diff
@@ -143,42 +143,38 @@ function Win_unlock(rank::Integer, win::Win)
...
     # int MPI_Put(const void *origin_addr, int origin_count,
     #             MPI_Datatype origin_datatype, int target_rank,
     #             MPI_Aint target_disp, int target_count,
     #             MPI_Datatype target_datatype, MPI_Win win)
-    T = eltype(origin_buffer)
     @mpichk ccall((:MPI_Put, libmpi), Cint,
                   (MPIPtr, Cint, MPI_Datatype, Cint, Cptrdiff_t, Cint, MPI_Datatype, MPI_Win),
-                  origin_buffer, count, Datatype(T), target_rank, Cptrdiff_t(target_disp), count, Datatype(T), win)
+                  origin_buf.data, origin_buf.count, origin_buf.datatype,
+                  target_rank, Cptrdiff_t(target_disp), origin_buf.count, origin_buf.datatype, win)
```
joachimbrand, line 171: I notice that `count::Integer` is passed as an argument but not used in the new method definition. It should probably be removed.

simonbyrne: Thanks!

simonbyrne:
@joachimbrand @mtsch: The one-sided ops haven't been widely used, so I'd appreciate feedback on these, and am open to other changes if you have any suggestions.

We probably should rename some of them at some point (e.g. `MPI.Win_sync` could just be `MPI.sync`, add `!` suffixes where appropriate).

Issue #469, opened 2021-04-30
An example on how to `Scatterv!` and `Gatherv!` a two-dimensional array?


francispoulin:
I have managed to use `Bcast!` successfully when I define it to be a `Vector` or a `Matrix`.  At the moment I'm stuck because I want to split (for example) a `5x3` matrix and I want to split it between two cores: the first should be `3x3` and the other `2x3`.  However, I need to define this in one object.

I figured out how to do this using `Vector{Matrix{Float64}}(undef, comm_size)` but then when I try and use `MPI.Bcast!` it complains that this is not `isbitstype`, which I'm sure is true.  I've only been able to successfully broadcast `Vector` or `Array`, but neither of these can describe the split that I want.

Can someone help me figure out what buffers `Bcast!` accepts and how I might be able to produce this object?

The error I get is copied below.

```
ERROR: LoadError: ArgumentError: Type must be isbitstype
Stacktrace:
 [1] MPI.Datatype(::Type{Matrix{Float64}}; commit::Bool)
   @ MPI.Types ~/.julia/packages/MPI/26D0s/src/datatypes.jl:267
 [2] MPI.Datatype(::Type{Matrix{Float64}})
   @ MPI.Types ~/.julia/packages/MPI/26D0s/src/datatypes.jl:266
 [3] MPI.Buffer(arr::Vector{Matrix{Float64}})
   @ MPI ~/.julia/packages/MPI/26D0s/src/buffers.jl:138
 [4] Bcast!(data::Vector{Matrix{Float64}}, root::Int64, comm::MPI.Comm)
   @ MPI ~/.julia/packages/MPI/26D0s/src/collective.jl:39
 [5] top-level scope
```

francispoulin:
Yes, I first want to `Scatterv!` and then `Gatherv!`.  As a check I was trying to do `Bcast!` as a simple test and that failed.  I figured if that failed then others would fail.  

The example that I'm trying to translate is in python and [here.](https://stackoverflow.com/questions/36025188/along-what-axis-does-mpi4py-scatterv-function-split-a-numpy-array/36082684#36082684).  

My sample code that does produce the correct results but needs some work is [here](https://github.com/francispoulin/Linear-Stability-Calculators/blob/main/QuasiHydrostataic/Julia/test_2D_array.jl).  In this example I need to assume that the number of rows is a multile of the comm size, but I want to generalize that.  

I was also thinking that others might benefit from this and perhaps this could become an example, somewhere.  

[...]

PR #470, opened 2021-05-02
Add Scatterv example based on #469

simonbyrne:


francispoulin (approved):
I think this is a very nice example and will be very helpful for new users.

PR #471, opened 2021-05-10
Improve Datatype handling

simonbyrne:
- Uses generated functions to avoid creating new instances `Datatype`s on every communication operation (#462). These are lazily initialized to allow for use in precompiled modules.
- Attach the corresponding Julia to an MPI Datatype as an attribute
- Nicer printing of Datatypes
- Allow specifying bits type as return type of MPI.Scatter
- Minor optimisation when constructing buffers of a single NTuple


src/datatypes.jl
```diff
@@ -53,11 +136,47 @@ for (mpiname, T) in [
...
+        datatype = $newtype
+        # lazily initialize so that it can be safely precompiled
+        if datatype.val === MPI.MPI_Datatype_default
+            Types.create!(datatype, T)
+            Types.commit!(datatype)
+            set_attr!(datatype, JULIA_TYPE_PTR_ATTR[], pointer_from_objref(T))
+        end
+        return datatype
+    end
+end
```
simonbyrne, line 158: @vchuravy Any thoughts on the pros/cons of using a generated function here?

vchuravy: hm... this __should__ be okay.

simonbyrne: YOLO

PR #472, opened 2021-06-03
Run CI on aarch64 with Drone

giordano:
With CPUs like A64FX, `aarch64` is an interesting platform for HPC.  I think
it'd be good to tests `MPI.jl` works here.  I just ran the tests on Isambard 2
and they are successful.

However, I have a small application which I tried to run on a MacBook M1
(`aarch64-apple-darwin`) and it crashes at
https://github.com/JuliaLang/julia/blob/58ffe7e3ed3a93a9d816097548e785284f57fbd4/src/codegen.cpp#L5531-L5536
in a call to `MPI.Reduce!`.  Tests are successful also here, so perhaps they
don't capture this error (or the error lies somewhere else), but I'll need to
produce a minimal working example to get a useful test out of it.

If you want to go on with this one, someone will need to enable this repository
at https://cloud.drone.io/.

---

*Edit*: regarding the error I got, I'm using a custom reduction, Valentin pointed out this is known to not work at the moment on non-Intel platforms and tests about them are skipped.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -1,0 +1,30 @@
+local Pipeline(os, arch, version, alpine=false) = {
+    kind: "pipeline",
+    name: os+" - "+arch+" - Julia "+version+(if alpine then " (Alpine)" else ""),
+    platform: {
+        os: os,
+        arch: arch
+    },
+    steps: [
+        {
+            name: "Run tests",
+            image: "julia:"+version+(if alpine then "-alpine" else ""),
+            commands: [
+                "julia --project=. --check-bounds=yes --color=yes -e 'using InteractiveUtils; versioninfo(verbose=true)'",
+                "julia --project=. --check-bounds=yes --color=yes -e 'using Pkg; Pkg.build()'",
+                "julia --project=. --check-bounds=yes --color=yes -e 'using Pkg; Pkg.test(coverage=true)'"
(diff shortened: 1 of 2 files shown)


simonbyrne:
What was the code doing? It sounds like you were using a custom reduction operator: these aren't supported on ARM or PPC (we currently disable tests on those platforms, see #353).

#404 had some ideas to workaround this.

giordano:
Yes, as I wrote in the "Edit", I was using a custom reduction

giordano:
Julia v1.6 on aarch64 is failing with
```
Test Failed at /drone/src/test/test_io_shared.jl:22
  Expression: MPI.File.get_position_shared(fh) == 0
   Evaluated: 9 == 0
ERROR: LoadError: There was an error during testing
in expression starting at /drone/src/test/test_io_shared.jl:22
```
Interestingly enough, Julia v1.3 is successful :confused:

simonbyrne:
Yeah, that happens intermittently. It's unclear what the consistency semantics are for shared file pointers.

simonbyrne:
Thanks!

giordano:
BTW, I forgot to mention that tests are successful on `aarch64-apple-darwin` with `ENV["JULIA_MPI_BINARY"]="OpenMPI_jll"`

PR #473, opened 2021-06-05
Throw an error when using custom reductions on non-Intel platforms

giordano:
The error thrown by Julia when getting to closures unsupported on non-Intel
platforms is like this:

```
ERROR: LoadError: cfunction: closures are not supported on this platform
in expression starting at /path/to/file.jl:line
```

and nothing else, without a full stacktrace, but `/path/to/file.jl:line` is only
the top-level frame of the stack.  It can take a while to understand where the
error is actually coming from especially if several packages are involved,
provided that you're willing to debug it, otherwise users are out of luck.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -69,12 +69,14 @@ end
 
 
 function Op(f, T=Any; iscommutative=false)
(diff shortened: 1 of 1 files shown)


simonbyrne:
Thanks!

PR #482, opened 2021-07-12
use unsafe_convert instead of cconvert to ensure handles aren't GCed during ccall

simonbyrne:


simonbyrne:
Thanks to @jkozdon for pointing this out.

PR #485, opened 2021-07-29
add nice init function

### LINKED ISSUE #484: More user-friendly MPI.Init
While C/C++/Fortran applications using MPI would call `MPI_Init` at the beginning of the program and that's it, in Julia it's more common that you may want to put `MPI.Init()` inside a function that you may call multiple times, especially when playing with it in the REPL.

To avoid calling `MPI_Init` multiple times, which isn't allowed, it's common to use a pattern like the following:
```julia
if !MPI.Initialized()
    MPI.Init()
end
```
([example 1](https://github.com/trixi-framework/Trixi.jl/blob/784650dbc903037c557202275cc885cf26dc78d2/src/auxiliary/mpi.jl#L19-L26), [example 2](https://github.com/JuliaMolSim/DFTK.jl/blob/d26aaea37a0b5f6dd9683ddcbaa4287183e7ebaf/src/common/mpi.jl#L4-L15), [example 3](https://github.com/eschnett/ADIOS2.jl/blob/1a3d52216fea306b63928217da584b45b768ebba/test/runtests.jl#L9-L13), [example 4](https://github.com/kibaekkim/DualDecomposition.jl/blob/b780adb9de630ba3769564f4c79403f5326f2e39/src/parallel.jl#L6-L8), [etc...](https://github.com/search?l=Julia&p=2&q=MPI.Initialized&type=Code)).

It'd be nice to have `MPI.Init` to automatically check whether MPI is already initialised and call `MPI_Init` only if it isn't the case.  However, @sloede pointed out on Slack that

> it would be necessary to support the different multi-threading settings and make sure that you either abort or warn if the requested threading level is not supported - or if an earlier call to MPI_Init did set a lower level than what you need.

Could do the same also for `MPI.Finalize`

simonbyrne:
Fixes #484.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MPI"
 uuid = "da04e1cc-30fd-572f-bb4f-1f8673147195"
 authors = []
-version = "0.18.2"
+version = "0.19.0"
 
 [deps]
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
--- a/docs/src/environment.md
+++ b/docs/src/environment.md
@@ -18,7 +18,6 @@ MPI.ThreadLevel
 ```@docs
 MPI.Abort
 MPI.Init
-MPI.Init_thread
 MPI.Query_thread
 MPI.Is_thread_main
 MPI.Initialized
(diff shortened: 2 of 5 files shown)


giordano (approved):
Looks great!  Would it be too magical/expensive to check `MPI.Finalized` in `MPI.Init` and `MPI.Inilized` in `MPI.Finalized`?

simonbyrne:
Good point. Note that we don't need to check `MPI.Initialized` in `Finalize`, since Finalized implies Initialized

giordano:
```julia
julia> using MPI

julia> MPI.Finalize()
Attempting to use an MPI routine before initializing MPICH
%
```

simonbyrne:
Ah I see. Any MPI operation will give an error like that, so I don't think we need to make `Finalize` special here.

PR #492, opened 2021-08-12
Waitall!: Work around const-prop failure in setproperty!

vchuravy:
@aviatesk any ideas why `setproperty!` failed to const-prop?

On 1.7 I am seeing:

```
julia> using Cthulhu

julia> using MPI

julia> descend(MPI.Waitall!, Tuple{Vector{MPI.Request}})

462 └───        invoke Base.setproperty!(%84::MPI.Request, :buffer::Symbol, MPI.nothing::Nothing)::Any

   %95  = invoke setproperty!(::MPI.Request,::Symbol,::Nothing)::Any

```

### DIFF
--- a/src/pointtopoint.jl
+++ b/src/pointtopoint.jl
@@ -459,7 +459,7 @@ function Waitall!(reqs::Vector{Request})
         req = reqs[i]
         if !isnull(req)
             req.val = reqvals[i]
-            req.buffer = nothing
+            Base.setfield!(req, :buffer, nothing)
         end
     end
     return stats
--- a/src/pointtopoint.jl
+++ b/src/pointtopoint.jl
@@ -406,7 +406,7 @@ function Wait!(req::Request)
                   (Ptr{MPI_Request}, Ptr{Status}),
                   req, stat_ref)
     if !alreadynull
-        req.buffer = nothing
+        Base.setfield!(req, :buffer, nothing)
     end
     stat_ref[]
 end


aviatesk:
This is because the return type is already accurate enough and so there is no need for constant-prop' in terms of type-inference: especially, it hits this line with `is_improvable(Const(nothing)) => false`:
https://github.com/JuliaLang/julia/blob/f19b9a20dde2688c642b7dd709b5d9446e4df2f6/base/compiler/abstractinterpretation.jl#L605

I wonder we actually don't need constant-prop' here, since inlinear will inline this `setproperty!` to `setfield!` anyway ? But I agree that Cthulhu output is misleading for this case since it can be interpreted as "there is a constant-prop' failure".

vchuravy:
The issue is that this is actually poisoning codegen. So the produced LLVM IR is suboptimal.

aviatesk:
Yes, it looks like this actually prevents inlining:
```
julia> code_typed(MPI.Waitall!, (Vector{MPI.Request},); optimize=true) |> 
           first
CodeInfo(
...
28 ─ %83 = Base.arrayref(true, %11, %74)::Int32
│          Base.setfield!(%76, :val, %83)::Int32
└───       invoke Base.setproperty!(%76::MPI.Request, :buffer::Symbol, MPI.nothing::Nothing)::Any
...
) => Vector{MPI.Status}
```

This seems better to be fixed, I will look into the solutions.

For now, manual inlining as you did looks reasonable.

aviatesk:
It turns out the inlining failure comes from the fact that `Base.convert` inside `setproperty!` costs too much w/o constant prop' when there is `Any`-typed field:
```julia
setproperty!(x, f::Symbol, v) in Base at Base.jl:43
│ ─ %-1  = invoke setproperty!(::MPI.Request,::Symbol,::Nothing)::…
43 1 ─    0 %1 = Base.fieldtype(MPI.Request, _3)::Union{Type{Any}, Type{Int32}}
   │      1 %2 = (isa)(%1, Type{Any})::Bool                                 │
   └──    0      goto #3 if not %2                                          │
   2 ─    0      goto #4                                                    │
   3 ─ 1000 %5 = Base.convert(%1, _4)::Core.Const(nothing)                  │
   └──    0      goto #4                                                    │
   4 ┄    0 %7 = φ (#2 => nothing, #3 => %5)::Core.Const(nothing)           │
   │      3 %8 = Base.setfield!(_2, _3, %7)::Core.Const(nothing)            │
   └──    0      return %8
```

simonbyrne:
Should this be closed then?

aviatesk:
This inlining failure is fixed on the very latest master. I think it's still reasonable to have this change if you work on v1.7 or lower.

vchuravy:
I only needed this to get Enzyme to differentiate through some MPI.jl code. I don't think it matters for other use-cases.

PR #496, opened 2021-08-23
update one sided API

simonbyrne:
- Make use of exclamation consistent
- Allocations return Arrays
- add docs


simonbyrne:
@joachimbrand @barche: any thoughts?

joachimbrand:
This looks good to me. I like the consistent use of the "!" and improved documentation.

I've tested with our code and it seems to run mostly fine (using a small subset of one-sided functionality, though: `Win_create`, `Win_fence`, `Win_lock`, and `Put!`).
In one case (out of 5, not reproducible) I got an unexplained segfault on termination. I have no idea why that would be and whether it is related to the code changes, but I thought I should mention it:
```
it092078:test brand$ mpirun -n 2 julia --depwarn=yes mpi_runtests.jl 
Saving report to `result.arrow`.
[        1000 | shift:   -15.7531 | norm:  9975.5972
[        2000 | shift:   -11.4098 | norm: 10017.0083
[        3000 | shift:    -8.7169 | norm: 10003.2468
[        4000 | shift:     -5.305 | norm:  9977.4107
[        5000 | shift:      -4.12 | norm:  9975.2381
[        6000 | shift:    -7.1634 | norm:   9999.186
[        7000 | shift:    -7.1397 | norm:  10008.305
[        8000 | shift:    -8.4171 | norm: 10016.1549
[        9000 | shift:    -5.6871 | norm:  9978.8602
[       10000 | shift:    -6.2201 | norm:  9997.9236
Finalizing.
Finished!
Test Summary: | Pass  Total
MPI tests     |  457    457

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 33399 RUNNING AT it092078
=   EXIT CODE: 11
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
YOUR APPLICATION TERMINATED WITH THE EXIT STRING: Segmentation fault: 11 (signal 11)
This typically refers to a problem with your application.
Please see the FAQ page for debugging suggestions
```
This was on macOS 11.5.2 using mpich 3.3.1. The script I ran is here: [mpi_runtests.jl](https://github.com/joachimbrand/Rimu.jl/blob/feature/mpi-0.19plus-compat/test/mpi_runtests.jl). It probably doesn't mean anything, though.

joachimbrand:
Thanks, @simonbyrne ! This might explain it. We could, of course, avoid allocating a new array every time (would make sense anyway) and re-use the send (`Put!`) buffer. I've changed our code accordingly and this seems to work fine and runs without errors (however, the error I saw was not reproducible). Our one-sided implementation is not really optimised and has not seen as much love as the point-to-point option, which is currently used as the default.

In any case, for MPI.jl it would of course be much better if you could avoid the gc cleaning up put buffers before they are read out 😄

vchuravy:
Yeah, having the object in Win like in Requests makes sense.

joachimbrand:
It looks like this API change has not yet made it into a released version despite being mentioned in the change log for v0.19.1. Would it be possible to make a new release that includes this change?

simonbyrne:
The automatic change log generator got confused. Since it's a breaking change, it will have to wait for the next release. We should probably tag one soon.

joachimbrand:
Ah, it was deliberate. No worries, then. Thanks!

PR #497, opened 2021-08-24
Broadcasting without a bang

carstenbauer:
Introduces `Bcast` to complement `Bcast!`. Similar to `Send(obj, ...)` the new variant is useful to directly broadcast objects, scalars in particular, to all ranks without having to introduce buffers or the like.

The usecase that made me create this PR is as simple as `Bcast(1.23, ...)` or `Bcast(p, ...)` where `p` is an instantiation of a struct, say:
```julia
struct Params
    a::Float64
    b::Float64
    n::Int64
end
```
Before the PR, one had to wrap these objects manually into `Ref`s to then use `Bcast!(Ref(obj),...)`. This automates this process (and special cases arrays, where we'll simply make a copy, and `Ref` input objects).

### DIFF
--- a/src/collective.jl
+++ b/src/collective.jl
@@ -65,11 +65,7 @@ end
 
 Broadcast the `obj` from `root` to all processes in `comm`. Returns the object.
 """
-function Bcast(obj::T, root::Integer, comm::Comm) where T
-    buf = Ref{T}(obj)
-    Bcast!(buf, root, comm)
-    return buf[]
-end
+Bcast(obj, root::Integer, comm::Comm) = Bcast!(Ref(obj), root, comm)[]
 Bcast(arr::AbstractArray, root::Integer, comm::Comm) = Bcast!(copy(arr), root, comm)
 Bcast(r::Ref, root::Integer, comm::Comm) = Ref(Bcast(r[], root, comm))
 
--- a/src/collective.jl
+++ b/src/collective.jl
@@ -61,13 +61,17 @@ function Bcast!(data, root::Integer, comm::Comm)
 end
 
 """
-    Bcast(obj::T, root::Integer, comm::Comm) where T
+    Bcast(obj, root::Integer, comm::Comm)
 
 Broadcast the `obj` from `root` to all processes in `comm`. Returns the object.
+Currently `obj` must be `isbits`, i.e. `isbitstype(typeof(obj)) == true`.
 """
-Bcast(obj, root::Integer, comm::Comm) = Bcast!(Ref(obj), root, comm)[]
-Bcast(arr::AbstractArray, root::Integer, comm::Comm) = Bcast!(copy(arr), root, comm)
-Bcast(r::Ref, root::Integer, comm::Comm) = Ref(Bcast(r[], root, comm))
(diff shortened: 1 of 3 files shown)


carstenbauer:
Would have wanted this again today (for broadcasting scalars). What do you think Simon, should we merge this?

src/collective.jl
```diff
@@ -39,6 +39,19 @@ function Bcast!(data, root::Integer, comm::Comm)
...
+    Bcast(obj::T, root::Integer, comm::Comm) where T
+
+Broadcast the `obj` from `root` to all processes in `comm`. Returns the object.
+"""
+function Bcast(obj::T, root::Integer, comm::Comm) where T
+    buf = Ref{T}(obj)
+    Bcast!(buf, root, comm)
+    return buf[]
+end
+Bcast(arr::AbstractArray, root::Integer, comm::Comm) = Bcast!(copy(arr), root, comm)
```
vchuravy, line 52: I am skeptical of the benefit of these copying definitions. In general this is a bit awkward since it requires the user to allocate a consistent `arr` across all ranks and then they don't get a benefit out of having done that allocation.

`bcast` below you can at least use as:

```
obj = Comm_rank(comm) == root ? bcast(obj, root, comm) : bcast(nothing, root, comm)
```

simonbyrne: I think this is my main reluctance as well: `Bcast!(copy(arr), ...)` doesn't seem that much less useful.

src/collective.jl
```diff
@@ -39,6 +39,19 @@ function Bcast!(data, root::Integer, comm::Comm)
...
+"""
+    Bcast(obj::T, root::Integer, comm::Comm) where T
+
+Broadcast the `obj` from `root` to all processes in `comm`. Returns the object.
+"""
+function Bcast(obj::T, root::Integer, comm::Comm) where T
+    buf = Ref{T}(obj)
+    Bcast!(buf, root, comm)
+    return buf[]
+end
```
simonbyrne, line 51: This ends up being a one-liner
```suggestion
Bcast(obj, root::Integer, comm::Comm) = Bcast!(Ref(obj), root, comm)[]
```
I'm still not sure it is worth it?

src/collective.jl
carstenbauer: Agreed, let's drop this. My main use case is easy broadcasting of scalars.

carstenbauer:
Alright, I've dropped the copying definitions and restricted the introduced `Bcast` to `isbitstype`s where we can just wrap the object into a `Ref`. This still removes the ugly manual wrapping/unwrapping ala
```
result = Bcast!(Ref(1.23), ...)
result[] # actually retrieve the data
```
in favor of the nicer
```
result = Bcast(1.23, ...)
```
for (isbits) structs and scalars such as simple numbers (my original motivation). However, I must admit that I'm not sure it's worth introducing a separate function for this. What do you think?

simonbyrne:
Okay, I'm happy to merge it. Needs to be added to the docs though.

carstenbauer:
> Needs to be added to the docs though.

Done.

PR #512, opened 2021-10-11
Added interface to MPI Ibarrier collective

mikegros:
Added ability to call MPI Ibarrier. Useful for allowing ranks to continue listening for messages or otherwise processing while waiting at a barrier. Tested successfully with openmpi 2.1.2.


mikegros:
I believe the test I've added should work. I tested it with openmpi 2.1.2 and it worked. Let me know if there's anything more needed.

mikegros:
Thanks for catching that file extension typo. Embarrassing. Cheers and have a great day.

PR #520, opened 2021-11-15
Various fixes to documentation

giordano:
In the documentation there are currently plenty of missing docstrings and broken references in docstrings, see for example: https://juliaparallel.github.io/MPI.jl/dev/pointtopoint/

I tried to fix all issues and add to documentation all functions with docstrings.  Please double check I didn't mess up anything, and also that I didn't add docstrings you didn't want to show in the docs.  Now building the documentation locally gives me only a couple of innocuous warnings about `DocTestSetup` being already set.

_Edit_: see the clean log when building the documentation: https://github.com/JuliaParallel/MPI.jl/runs/4217747616?check_suite_focus=true#step:4:1

### DIFF
--- a/docs/src/advanced.md
+++ b/docs/src/advanced.md
@@ -10,6 +10,7 @@ MPI.free
 
 ```@docs
 MPI.Datatype
+MPI.to_type
 MPI.Types.extent
 MPI.Types.create_contiguous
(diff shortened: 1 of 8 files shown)


simonbyrne:
Thank you!

PR #281, opened 2019-06-27
Add group functions

vchuravy:
x-ref: #277

PR #521, opened 2021-11-24
Add group functions (v2.0)

carlocastoldi:
Hey!

During my internship I needed multiple communicators so that I could write to file with MPI I/O whenever all the workers in the same groups (one worker could be in multiple groups) finished their part, without waiting for workers outside of their groups.
In order to do so I adapted #281 pull request to my use-case and rebased it to the current master. I additionally modified `deps/consts_mpich.jl` using the constant values found in [`mpi.h.in`](https://github.com/pmodels/mpich/blob/c6d34d2ceed003af344e7cb59da10e968e9d8d8e/src/include/mpi.h.in#L91) and tested it with HPE-MPI, the MPI implementation my workplace uses.

You can test it with the following script that tests my use-case:
given  an input array it splits it into 2 slices, one per worker.
* worker 1: `[11, 11, 11, 11, 11, 11, 11, 11, 11, 11, 11, 12, 12, 12, 12, 12, 12, 12]`
* worker 2: `[12, 12, 12, 12, 12, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13]`
We now want to group the workers based on the data they work upon:
* `11` → {worker 1}
* `12` → {worker 1, worker 2}
* `13` → {worker 2}
```julia
# run on TWO nodes
# using Distributed, SlurmClusterManager
# addprocs(SlurmManager())

@everywhere using MPI
@everywhere function working_range(total_pts::Int, rank_id)
    pts_per_worker = total_pts / nworkers() |> ceil |> Int64
    first_i =     pts_per_worker * (rank_id-1) + 1
    last_i  = min(pts_per_worker *  rank_id, total_pts)
    return first_i:last_i
end

@everywhere function find_id2rank_mask(arr)
    ranges = [working_range(length(arr), rank) for rank in 1:nworkers()]
    ids = unique(arr)
    id2rank = Dict(id => falses(nworkers()) for id in ids)
    for rank in 1:length(ranges), i in ranges[rank]
        id2rank[arr[i]][rank] = true
    end
    return id2rank
end

@everywhere function ranks_mask2comm(base_comm::MPI.Comm, id::Int, ranks_mask::BitVector)
    base_grp = MPI.Comm_group(base_comm)
    ranks = Cint.((0:MPI.Comm_size(base_comm)-1)[ranks_mask])
    id_group = MPI.Group_incl(base_grp, ranks)
    comm = MPI.Comm_create_group(base_comm, id_group, id)
    return comm
end

@everywhere function find_id2comm(arr::Vector)
    base_comm = MPI.COMM_WORLD
    rank_id = MPI.Comm_rank(base_comm)+1
    id2ranks_mask = find_id2rank_mask(arr)
    id2comm = Dict(id => ranks_mask2comm(base_comm, id, ranks_mask) for (id, ranks_mask) in id2ranks_mask if ranks_mask[rank_id])
    return id2comm
end

@everywhere arr = vcat([fill(i+10,i+10) for i in 1:nprocs()]...)
println(arr)
@sync @distributed for i in 1:nworkers()
    if !MPI.Initialized()
        MPI.Init()
    end
    id2comm = find_id2comm(arr)
    for (id,comm) in id2comm
        println("RANK $(MPI.Comm_rank(MPI.COMM_WORLD)) CHUNK-$id --- MPI.Comm_size: $(MPI.Comm_size(comm)), MPI.Comm_rank: $(MPI.Comm_rank(comm))")
    end
end
```
OUTPUT:
```
[11, 11, 11, 11, 11, 11, 11, 11, 11, 11, 11, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13]

RANK 0 CHUNK-11 --- MPI.Comm_size: 1, MPI.Comm_rank: 0
RANK 1 CHUNK-13 --- MPI.Comm_size: 1, MPI.Comm_rank: 0
RANK 0 CHUNK-12 --- MPI.Comm_size: 2, MPI.Comm_rank: 0
RANK 1 CHUNK-12 --- MPI.Comm_size: 2, MPI.Comm_rank: 1
Task (done) @0x00007fffb4ef30a0
```
fixes: #277

P.S. just one question: i feel like functions such as `Comm_create(comm::Comm, group::Group)`, `Comm_create_group(comm::Comm, group::Group, tag::Integer)` and perhaps even `Comm_group(comm::Comm)` and `Comm_remote_group(comm::Comm)` could also stay in in `comm.jl`. What do you think?


src/group.jl
```diff
@@ -0,0 +1,134 @@
...
+    newgroup = Group()
+    @mpichk ccall((:MPI_Group_incl, libmpi), Cint,
+        (MPI_Group, Cint, Ptr{Cint}, Ptr{MPI_Group}), group, length(ranks), ranks, newgroup)
+    finalizer(free, newgroup)
+    return newgroup
+end
+
+# int MPI_Group_range_excl(MPI_Group group, int n, int ranges[][3], MPI_Group *newgroup)
+# int MPI_Group_range_incl(MPI_Group group, int n, int ranges[][3], MPI_Group *newgroup)
+# int MPI_Group_translate_ranks(MPI_Group group1, int n, const int ranks1[], MPI_Group group2, int ranks2[])
```
simonbyrne, line 134: You can either get rid of these, or move them above the `ccall`s

src/group.jl
```diff
@@ -0,0 +1,134 @@
...
+    return nothing
+end
+
+"""
+    Group_rank(group::Group)
+
+The rank of the process in the particular group.
+
+Returns an integer in the range `0:MPI.Group_size()-1`.
+"""
```
simonbyrne, line 21: These should be added to the docs (can put them in `docs/src/advanced.md`)

simonbyrne (changes requested):
Thanks! It still needs the following:
- [x] Open MPI constants: you can copy what we do for Comm objects, and find the fortran constant values here: https://github.com/open-mpi/ompi/blob/master/ompi/include/mpif-values.pl
- [ ] Some basic tests

carlocastoldi:
I took the liberty to move the 4 functions above mentioned into `comm.jl` and fixing some typos in `io.jl` docs.
Additionally: I have yet to see if this assumption is correct or not. I guess we will see it once the tests are run hahaha
```julia
@test MPI.Group_compare(grp, MPI.GROUP_NULL) === MPI.IDENT
```

carlocastoldi:
I finally managed to run `test_group.jl`. Sorry about it. Now I am sure that with the latest version the tests pass _on my system_

carlocastoldi:
From the pipeline i see that it fails due to this:
```
PMPI_Group_compare(172): MPI_Group_compare(MPI_GROUP_NULL, MPI_GROUP_NULL, result=0x7ffcf7f605c8) failed
PMPI_Group_compare(128): Null group
```

With my available implementation of MPI it worked fine, but after looking up for [MPI_Comm_compare](https://www.mpich.org/static/docs/latest/www3/MPI_Comm_compare.html) (not [MPI_Group_compare](https://www.mpich.org/static/docs/latest/www3/MPI_Group_free.html)!) i found out that:
> It is an error to use MPI_COMM_NULL as one of the arguments to MPI_Comm_compare.

I guess it's the same for MPI_Group_compare, then. I'll drop that test case after all...

deps/gen_consts.jl
```diff
@@ -107,6 +107,9 @@ MPI_handle = [
     ],
     :MPI_File => [
         :MPI_FILE_NULL,
+    :MPI_Group => [
```
simonbyrne, line 110:
```suggestion
    ],
    :MPI_Group => [
```

simonbyrne:
Thank you!

carlocastoldi:
Thank you for merging!

When can we expect the next release, if I may ask? I would like my internship project to depend on the official MPI.jl package rather than on my own patched version 😁

simonbyrne:
When do you need it by? There are a few more breaking changes I would like to get in before the next release, but since this isn't breaking, I could backport it to the 0.19 branch?

carlocastoldi:
I finish my internship at the end of the year.
Surely if it's not too big of a problem for you, backporting sounds like a good solution to me. But if it was an hassle for you, I think i can still go with referencing my own repository as a dependency.
Having it backported would allow the framework I am building now to be technically easily updatable in the future, tho

Whatever it's your decision, thank you a lot!

simonbyrne:
I've backported and tagged v0.19.2, so if you update and set that as the minimum version you'll be all set.

Thanks again for the contribution!

carlocastoldi:
I updated my framework to the latest version of MPI.jl
It all works as it should! Thanks a ton for very fast help you provided me with ^^

Issue #483, opened 2021-07-13
Don't download any artifacts if JULIA_MPI_BINARY=system


crstnbr:
MPI.jl seems to always download the MPICH and OpenMPI artifacts despite having set `JULIA_MPI_BINARY=system`. It is using the correct system MPI eventually but ideally it shouldn’t download the unused artifacts in the first place.

I guess / fear that this is a limitation of Pkg and the lack of optional dependencies. But I wanted to voice and archive this point somewhere and thought here would be the right place.

simonbyrne:
I guess we could make them lazy, and download the appropriate ones at package build time?

simonbyrne:
There are some new features in 1.6 for handling lazy artifacts, i haven't looked into it though. If someone wants to try it, i'd certainly be interested.

crstnbr:
MKL_jll seems to be lazy, see discussion over at https://github.com/JuliaLang/Pkg.jl/issues/2664. However, don't know what that laziness means exactly. Does it also allow optional download without an Override.toml?

simonbyrne:
We don't support Overrrides.toml, as it isn't flexible enough for our needs (e.g. Overrides.toml requires the library version number to be the same, which means we can't support different MPI library versions, let alone different implementations).

We should also look into using [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl) instead of our current homebrew approach. I don't have time to tackle this at the moment, but am happy to review PRs.

sloede:
> However, don't know what that laziness means exactly. Does it also allow optional download without an Override.toml?

`lazy` means that the artifact is downloaded at first use. That is, the actual artifact is not downloaded until you use something like `artifact"XYZ"` to get the path, at which point all lazy artifacts will be downloaded.

From https://pkgdocs.julialang.org/v1/artifacts/#Using-Artifacts
> For the specific use case of using artifacts that were previously bound, we have the shorthand notation artifact"name" which will automatically search for the Artifacts.toml file contained within the current package, look up the given artifact by name, install it if it is not yet installed, then return the path to that given artifact.

[...]

PR #524, opened 2021-12-05
Try using Preferences.jl instead of Pkg.build

simonbyrne:
This is an attempt to switch to using Preferences.jl instead of relying on environment variables. It is somewhat orthogonal to #513, but there are some design decisions that will necessarily interact.

This is based on discussion #483.

Instead of using environment variables, users would now call
```
julia -e 'using MPI; MPI.use_system_binary()'
```

There are a couple of pain points that I think would need to be addressed:
- We can't query or set preferences at Pkg.build time. This means that things we used to do (e.g. query environment variables, or run the code generator for unknown ABIs) will no longer work
- As such, we can only support known ABIs at this time.
- Setting the preferences requires installing and precompiling MPI, running the above code, then precompiling again. This is not an ideal workflow.

cc: @carstenbauer @giordano @staticfloat @vchuravy


staticfloat:
> We can't query or set preferences at Pkg.build time. This means that things we used to do (e.g. query environment variables, or run the code generator for unknown ABIs) will no longer work

Why can't you do this at precompile time instead?  What's special about `Pkg.build()` time?

simonbyrne:
Possibly? What happens if we modify the preferences? Is there a way to trigger a second round of precompilation?

staticfloat:
That’s the major feature of preferences: if you use a preference value at
compile-time, it gets logged, and if that value changes, the next time you
load the package it will get recompiled.

simonbyrne:
Ah. So to clarify, the hash is based on the state of the preferences _after_ precompilation, not before? i.e. so if I call `@set_preferences!` during precompilation, this won't trigger another round of precompilation when I load the package again?

staticfloat:
> So to clarify, the hash is based on the state of the preferences after precompilation, not before?

That is correct. Every time you call `@load_preference(key)`, the macro will check to see if you are currently compiling.  If you are, it will [mark that preference name as a compile-time preference](https://github.com/JuliaPackaging/Preferences.jl/blob/master/src/Preferences.jl#L30-L32).  Then, after the whole compilation is finished, it will write out your `.ji` file and embed a hash of the preferences [into the header of the `.ji` file](https://github.com/JuliaLang/julia/blob/3d4b213a6d751e07fd9becbc17221868af638117/src/dump.c#L1223-L1230).  Then, when searching `.ji` files, [we reject any previously-compiled `.ji` files](https://github.com/JuliaLang/julia/blob/2e9bbda5bd5bd6c04d019aa565ea60efab7c10d4/base/loading.jl#L1939-L1943) if their preferences hash doesn't match what was used when it was last compiled.

All that being said, you're absolutely responsible for ensuring that the state of your package is consistent, and it can get really confusing if a top-level definition changes halfway through precompilation.  Example:

```
module PrefsExample
using Preferences

const provider = something(
    @load_preference("provider", nothing),
    "default",
)::String

function do_work()
    @static if provider == "default"
        @info("do default thing")
    else
        @info("do something else")
    end
end

# Set a DIFFERENT preference value
@set_preferences!("provider" => "special")

end # module PrefsExample
```

If you compile this once, `do_work()` will print `do default thing`, and it won't recompile itself automatically.  But if you then do something else that causes  recompilation (even something as simple as a `.ji` file falling out of the LRU cache due to having many other environments that use this package) suddenly the behavior will change.  So my point is to make things as simple and straightforward as possible; if you can, only read the preference once, and only write it once, at well-defined points in your control flow.  I know some people like to allow for environment variables to influence these things, perhaps by doing something like:

```
const provider = something(
    @load_preference("provider", nothing),
    get(ENV, "PROVIDER", nothing),
    "default",
)::String
```

But I really dislike this for the following reasons:

- If you set this manually once during build time, if it needs to be recompiled at some time in the future and you neglect to set it, the result will be wrong.
- It's global, affecting all environments.  Preferences should be per-environment.
- It's not easily reproducible across machines.

It's much better to provide the user a `Package.set_provider()` method that tells the user "Preference set; please restart Julia" or something, if it's a compile-time preference.

simonbyrne:
Thanks @staticfloat. One of the challenges I'm trying to solve is figuring out the ABI of the user-provided MPI library: this should be stored in the Preferences.jl, since we want to invalidate the precompile cache if it changes. So I want to do something like:
```
const abi = begin
    _abi = @load_preference("abi", nothing)
    if isnothing(_abi)
       _abi = find_abi()
       @set_preference!("abi" => _abi)
    end
    _abi
end
```
Is that reasonable usage?

Also, if the ABI is not known, then we need to generate a .jl file with the constants by compiling and running a C program. Previously we stored this information in `deps`; should we now use [Scratch.jl](https://github.com/JuliaPackaging/Scratch.jl) for this? If so, does this mean that we have to put the ```julia
include(joinpath(@get_scratch!("constants"), "consts.jl"))
```
statement inside the `__init__()` function?

staticfloat:
> Is that reasonable usage?

Yes, as long as you don't expect the ABI to change out from underneath the user.  Your module only calls `find_abi()` at compile-time, so unless there's some other piece that changes the preference when the ABI changes, it will never recompile.  You could do something like do a sanity check `abi == find_abi()` within your `__init__()`, set the preference again, then tell the user to restart Julia, I suppose.


> should we now use Scratch.jl for this?

Yes, that's a great idea.  Using `Scratch` doesn't require any real changes to _when_ you do the work; you can do that in `__init__()` or at the top-level, either is fine.

PR #529, opened 2021-12-17
Add a lightweight MPIABI sub-package that controls the MPI ABI

vchuravy:
The goal is to have a central source of truth to query the MPI ABI used by Julia and by binary dependencies.

With this we can change the ABI, and after a restart the user will download the appropriate binaries/complain that they are incompatible.

With this I can change https://github.com/JuliaPackaging/Yggdrasil/pull/4073 to contain

```
augmented_platform_block = """
    using Base.BinaryPlatforms
    using MPIABI
    function augmented_platform(platform)
        abi = MPI.abi
        BinaryPlatforms.add_tag!(platform.tags, "mpi", string(abi))
        return platform
    end
"""
```

and have the LAMMPS_jll use the right ABI. 

Open questions are:
- ABI vs System and Library? E.g. Can we assume that binaries built against OpenMPI_jll are compatible with a system OpenMPI? I think so, so switching on the ABI and not the library used "should" work.
- Do we need to have a separate package? I think so, we want the choice to be as lightweight as possible, and MPI.jl might use MPIABI to what it includes itself.

cc: @eschnett @staticfloat @simonbyrne

### DIFF
--- a/lib/MPIABI/Project.toml
+++ b/lib/MPIABI/Project.toml
@@ -1,0 +1,7 @@
+name = "MPIABI"
+uuid = "3da0fdf6-3ccc-4f1b-acd9-58baa6c99267"
+authors = []
+version = "0.1.0"
+
+[deps]
+Preferences = "21216c6a-2e73-6563-6e65-726566657250"
--- a/lib/MPIABI/src/MPIABI.jl
+++ b/lib/MPIABI/src/MPIABI.jl
@@ -1,0 +1,79 @@
+module MPIABI
+
+using Preferences
+
+function known_abis()
+    return (:MicrosoftMPI, :MPICH, :OpenMPI, :MPItrampoline)
+end
+
+# const abi = begin
+#     _abi = @load_preference("abi", nothing)
+#     if isnothing(_abi)
+#        _abi = find_abi()
+#        @set_preference!("abi" => _abi)
+#     end
+#     _abi
+# end
+
+const abi = @load_preference("abi", Sys.iswindows() ? :MicrosoftMPI : :MPICH)
+
+function __init__()
+    if Sys.iswindows()
+        if abi != :MicrosoftMPI
+            @error """
+                The MPI ABI is not compatible with the current platform.
+                Please set the MPI ABI to :MicrosoftMPI.
+            """
+        end
+    end
+end
+
+function set_abi(abi)
+    if abi ∉ known_abis()
+        error("""
+            The MPI ABI $abi is not supported.
+            Please set the MPI ABI to one of the following:
+            $(known_abis())
+        """)
+    end
+    @set_preferences!("abi" => abi)
+    @warn "The MPI abi has changed, you will need to restart Julia for the change to take effect" abi
+
+    if VERSION <= v"1.6.5" || VERSION == v"1.7.0"
+        @warn """
+        Due to a bug in Julia (until 1.6.5 and 1.7.1), setting preferences in transitive dependencies
+        is broken (https://github.com/JuliaPackaging/Preferences.jl/issues/24). As a workaround, we 
+        will now add MPIABI as a direct dependency in your current project.
+        """
+
+        uuid = Preferences.get_uuid(@__MODULE__)
+        project_toml, pkg_name = Preferences.find_first_project_with_uuid(uuid)
+
+        if project_toml === nothing && pkg_name === nothing
+            # If we couldn't find one, we're going to use `active_project()`
+            project_toml = Base.active_project()
+
+            # And we're going to need to add this UUID as an "extras" dependency:
+            # We're going to assume you want to name this this dependency in the
+            # same way as it's been loaded:
+            pkg_uuid_matches = filter(d -> d.uuid == uuid, keys(Base.loaded_modules))
+            if isempty(pkg_uuid_matches)
+                error("Cannot set preferences of an unknown package that is not loaded!")
+            end
+            pkg_name = first(pkg_uuid_matches).name
+
+            # Read in the project, add the deps, write it back out!
+            project = Base.parsed_toml(project_toml)
+            if !haskey(project, "deps")
+                project["deps"] = Dict{String,Any}()
+            end
+            project["deps"][pkg_name] = string(uuid)
+            open(project_toml, "w") do io
+                TOML.print(io, project; sorted=true)
+            end
+        end
+    end
+    return abi
+end
+
+end # module
--- a/lib/MPIABI/UNLICENSE
+++ b/lib/MPIABI/UNLICENSE
@@ -1,0 +1,25 @@
+This is free and unencumbered software released into the public
+domain.
+
+Anyone is free to copy, modify, publish, use, compile, sell, or
+distribute this software, either in source code form or as a compiled
+binary, for any purpose, commercial or non-commercial, and by any
+means.
+
+In jurisdictions that recognize copyright laws, the author or authors
+of this software dedicate any and all copyright interest in the
+software to the public domain. We make this dedication for the benefit
+of the public at large and to the detriment of our heirs and
+successors. We intend this dedication to be an overt act of
+relinquishment in perpetuity of all present and future rights to this
+software under copyright law.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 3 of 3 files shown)


simonbyrne:
> Can we assume that binaries built against OpenMPI_jll are compatible with a system OpenMPI?

I seem to recall that the .so version number is baked into the binaries, so maybe not.

lib/MPIABI/src/MPIABI.jl
```diff
@@ -0,0 +1,79 @@
+module MPIABI
+
+using Preferences
+
+function known_abis()
+    return (:MicrosoftMPI, :MPICH, :OpenMPI, :MPItrampoline)
```
simonbyrne, line 6: should we make these all lowercase?

lib/MPIABI/src/MPIABI.jl
```diff
@@ -0,0 +1,79 @@
...
+
+function set_abi(abi)
+    if abi ∉ known_abis()
+        error("""
+            The MPI ABI $abi is not supported.
+            Please set the MPI ABI to one of the following:
+            $(known_abis())
+        """)
+    end
+    @set_preferences!("abi" => abi)
```
simonbyrne, line 39: In #524, I suggested the following:
```
#  binary   = "" (default) | "system" | "MPICH_jll" | "OpenMPI_jll" | "MicrosoftMPI_jll"
#  path     = "" (default) | top-level directory location
#  library  = "" (default) | library name/path | list of library names/paths
#  abi      = "" (default) | "MPICH" | "OpenMPI" | "MicrosoftMPI" | "unknown"
#  mpiexec  = "" (default) | executable name/path
```

lib/MPIABI/src/MPIABI.jl
```diff
@@ -0,0 +1,79 @@
+module MPIABI
+
+using Preferences
+
+function known_abis()
```
giordano, line 5: Out of curiosity, why a function and not a constant?

lib/MPIABI/src/MPIABI.jl
```diff
@@ -0,0 +1,79 @@
...
+#        _abi = find_abi()
+#        @set_preference!("abi" => _abi)
+#     end
+#     _abi
+# end
+
+const abi = @load_preference("abi", Sys.iswindows() ? :MicrosoftMPI : :MPICH)
+
+function __init__()
+    if Sys.iswindows()
```
giordano, line 21:
```suggestion
    @static if Sys.iswindows()
```
?

lib/MPIABI/src/MPIABI.jl
```diff
@@ -0,0 +1,79 @@
...
+        error("""
+            The MPI ABI $abi is not supported.
+            Please set the MPI ABI to one of the following:
+            $(known_abis())
+        """)
+    end
+    @set_preferences!("abi" => abi)
+    @warn "The MPI abi has changed, you will need to restart Julia for the change to take effect" abi
+
+    if VERSION <= v"1.6.5" || VERSION == v"1.7.0"
```
giordano, line 42: Maybe `@static` also here?

lib/MPIABI/Project.toml
```diff
@@ -0,0 +1,7 @@
+name = "MPIABI"
+uuid = "3da0fdf6-3ccc-4f1b-acd9-58baa6c99267"
+authors = []
+version = "0.1.0"
+
+[deps]
+Preferences = "21216c6a-2e73-6563-6e65-726566657250"
```
giordano, line 7: Compat for Preferences.jl and Julia

lib/MPIABI/Project.toml
```diff
@@ -0,0 +1,7 @@
+name = "MPIABI"
+uuid = "3da0fdf6-3ccc-4f1b-acd9-58baa6c99267"
+authors = []
```
giordano, line 3: Poor package, no one wants to claim authorship 😔

lib/MPIABI/src/MPIABI.jl
```diff
@@ -0,0 +1,79 @@
...
+        @warn """
+        Due to a bug in Julia (until 1.6.5 and 1.7.1), setting preferences in transitive dependencies
+        is broken (https://github.com/JuliaPackaging/Preferences.jl/issues/24). As a workaround, we 
+        will now add MPIABI as a direct dependency in your current project.
+        """
+
+        uuid = Preferences.get_uuid(@__MODULE__)
+        project_toml, pkg_name = Preferences.find_first_project_with_uuid(uuid)
```
giordano, line 50: Wouldn't `Pkg.add` but much easier? But you probably don't want to require `Pkg` just for a workaround?

lib/MPIABI/src/MPIABI.jl
vchuravy: `path`, `library`, `mpiexec` are dependent on which actual MPI installation is being loaded and the intention is for this to be used to choose which MPI installation to be loaded. 

The question is: Do we bifurcate along the `_jll` package used or along the ABI used. Thinking about the library versioning issue I think we need to make this mean with `_jll` is being used by MPI.jl (which is a shame) I would like it to be the ABI. 

Right now if the ABI is set to `OpenMPI` we will load `OpenMPI_jll` as part of `LAMMPS_jll`, so that would not be compatible with `binary==system && abi==openmpi` IIUC, except if we also set `libmpi_path` for `OpenMPI_jll`?

But then we could still run into a library version mismatch?

lib/MPIABI/src/MPIABI.jl
vchuravy: I am tempted to fix this in Preferences.jl, which is where I stole this code from.

vchuravy: If version 1.6.5 was already released I would just make that the compat xD

lib/MPIABI/src/MPIABI.jl
simonbyrne: How do you figure out the ABI without knowing the library etc?

lib/MPIABI/src/MPIABI.jl
giordano: 1.6.5 should be very close: https://discourse.julialang.org/t/julia-1-6-5-testing-period/73068

lib/MPIABI/src/MPIABI.jl
vchuravy: That's why this package doesn't do it. It is a communication channel for ABI aware `JLL` packages.

I was thinking MPI.jl would figure out the details and then set it here.

lib/MPIABI/src/MPIABI.jl
vchuravy: I think JLLWrapper&Co normalize these to lower-case. I don't have a strong preference.

lib/MPIABI/src/MPIABI.jl
simonbyrne: There are some advantages to setting them all in the same package:
- it's easier to ensure that they stay in sync (e.g. that the preferences are set in the same location)
- it might help avoid an unnecessary round of precompilation

vchuravy: Yeah I making this up here as we go along :) 
I don't think we have a good idea of what the best practices should be.

The other extreme is to just use MPI.jl And make all the JLL packages depend on it

simonbyrne: I like the idea of a subpackage, but if e.g. Yggdrasil binaries built against MPICH are only valid for MPICH_jll and not the system MPICH, it would make sense to have `binary` in the subpackage.

simonbyrne:
I've been thinking about this a bit more, and prompted by @staticfloat's comment here: https://github.com/JuliaInterop/CxxWrap.jl/issues/309#issuecomment-1002338843, could we somehow encapsulate all this (and https://github.com/JuliaPackaging/Yggdrasil/pull/4073) by creating a new MPI_jll package?

What I was vaguely thinking is that this would act like an ordinary jll package, but with some extra constants (which could be set by Preferences):
- `binary`: specify which jll package to use, or `"system"` if using a non-jll binary.
- `abi`: as defined in this PR
- `libmpi`, `mpiexec`, etc: these would import the appropriate jll package bindings, or if `binary == "system"`, would be set by `libmpi_path`, etc.

simonbyrne:
Ideally this would also handle the platform PackageSpec stuff (i.e. adding MPI_jll as a PackageSpec would automatically expand the platforms)

vchuravy:
Closed in favor of  #541

PR #530, opened 2021-12-18
Run CI only when relevant directories are modified

giordano:
This avoids spawning dozens of jobs when modifying unrelated files (like README,
documentation, etc...)

This is related to #529, as we can have a CI job to test only MPIABI.jl,
although we'll probably want to run downstream MPI.jl tests when MPIABI.jl is
modified.

### DIFF
--- a/.github/workflows/UnitTests.yml
+++ b/.github/workflows/UnitTests.yml
@@ -1,11 +1,20 @@
-name: Unit Tests
+name: MPI.jl Unit Tests
 
 on:
   pull_request:
-
+    paths:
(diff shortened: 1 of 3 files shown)


.github/workflows/UnitTests.yml
```diff
@@ -1,11 +1,20 @@
-name: Unit Tests
+name: MPI.jl Unit Tests
 
 on:
   pull_request:
-
+    paths:
```
giordano, line 5: We should probably also include this file in the list, so that when this is modified these tests are also run?

PR #536, opened 2022-01-26
clean up Cartesian topology functions

### LINKED ISSUE #533: `MPI.Cart_create` does not accept `Bool` arguments for `periodic`.
`MPI.Cart_create` triggers an error if booleans are provided for `periodic` argument, but accepts `Int`. Wouldn't it be better to cast booleans to `int` in the `ccall` in `MPI.jl`?
```
npx = 2; npy = 2

using MPI

MPI.Init()

dims = [npy, npx]; periodic = [true, true]; reorder = false
# dims = [npy, npx]; periodic = [1, 1]; reorder = false

commxy = MPI.Cart_create(MPI.COMM_WORLD, dims, periodic, reorder)
commx = MPI.Cart_sub(commxy, [false, true])
commy = MPI.Cart_sub(commxy, [true, false])

MPI.Finalize()
```

simonbyrne:
Fixes #533, also gets rid of mutating functions: they're not really that useful, and the make the API confusing.

### DIFF
--- a/src/topology.jl
+++ b/src/topology.jl
@@ -1,62 +1,78 @@
 """
-    Dims_create!(nnodes::Integer, [ndims::Integer], dims)
+    newdims = Dims_create(nnodes::Integer, dims)
 
-Create a division of `nnodes` processes in a Cartesian grid.
+A convenience function for selecting a balanced Cartesian grid of a total of `nnodes`
+nodes, for example to use with [`MPI.Cart_create`](@ref).
+
+`dims` is an array or tuple of integers specifying the number of nodes in each dimension.
+The function returns an array `newdims` of the same length, such that if `newdims[i] =
+dims[i]` if `dims[i]` is non-zero, and `prod(newdims) == nnodes`, and values `newdims` are
+as close to each other as possible.
+
+`nnodes` should be divisible by the product of the non-zero entries of `dims`.
 
 # External links
 $(_doc_external("MPI_Dims_create"))
 """
-function Dims_create!(nnodes::Integer, ndims::Integer, dims::MPIBuffertype{T}) where {T <: Integer}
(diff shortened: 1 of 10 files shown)


simonbyrne:
@jipolanco and @Chiil: any thoughts on this?

Chiil:
Great. I like consistency, now we can specify both the `periodic` and the `reorder` as boolean.

src/topology.jl
```diff
@@ -1,62 +1,78 @@
...
-function Dims_create!(nnodes::Integer, dims::AbstractArray{T,N}) where {T<:Integer, N}
-    cdims = Cint.(dims[:])
-    ndims = length(cdims)
-    Dims_create!(nnodes, ndims, cdims)
-    dims[:] .= cdims
-end
 
 """
-    comm_cart = Cart_create(comm_old::Comm, [ndims::Integer], dims, periods, reorder)
+    comm_cart = Cart_create(comm::Comm, dims; periodic=map(_->false, dims), reorder=false)
```
simonbyrne, line 26: Any thoughts on what should be the defaults? Or if we should have them at all?

jipolanco: I agree with both `periodic` and `reorder` being optional.

`periodic` being `false` along all dimensions seems like a reasonable default to me. I don't know if it matters, but this argument is called `periods` in the OpenMPI and MPICH documentations. I do prefer `periodic` though.

As for `reorder`, I'd say leaving it as `false` by default also makes sense, as it would avoid any surprises (reordered ranks) to potentials users who wouldn't want to bother with optional arguments.

jipolanco:
Everything looks good to me, thanks. I also agree with removing the mutating functions.

simonbyrne:
Thank you for the feedback!

PR #540, opened 2022-02-15
Silence warning about deprecated function when running tests

giordano:
I was unsure whether using `@test_deprecated` or just using the new syntax, but
I thought it may make sense to still test the deprecated function while we keep
it, to make sure it isn't somehow accidentally broken.

### DIFF
--- a/test/test_shared_win.jl
+++ b/test/test_shared_win.jl
@@ -51,7 +51,7 @@ function main()
         @test all(shared_arr[:, 1] .== 1:100)
         @test all(shared_arr[:, 2] .== 901:1000)
         if node_rank <= 1
-            len, elsize_bytes, baseptr = MPI.Win_shared_query(win, owner_rank)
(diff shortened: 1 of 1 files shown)


giordano:
Side note: would it make sense to show some more information about the system at https://github.com/JuliaParallel/MPI.jl/blob/fa377d4d60df9b94e5c86605473e01bb2939808d/test/runtests.jl#L21 when running the tests?  I was thinking about `MPI.Get_library_version()`, to show which MPI we're using (there is something similar in [`FFTW.jl` tests](https://github.com/JuliaMath/FFTW.jl/blob/3b49c8076de38ff2c5ab86cf6592cf020103a7bc/test/runtests.jl#L8)), but I reckon the output of `MPI.Get_library_version()` can be quite verbose.  Any other relevant variable?

simonbyrne:
Thanks. I never know how to handle these things.

I'll merge this for now: I need to do a cleanup of the test suite at some point, so we can clean it up then.

PR #541, opened 2022-02-15
Add MPIPreferences subpackage to select binary

simonbyrne:
Based on #528 and #529.


simonbyrne:
Note that the spurious tests are passing due to https://github.com/JuliaLang/Pkg.jl/pull/2920

simonbyrne:
@vchuravy @staticfloat: until https://github.com/JuliaLang/Pkg.jl/pull/2920 makes it into a new release, is there a way to work around the problem for testing purposes?

.gitignore
```diff
@@ -5,3 +5,4 @@ libjuliampi.dylib
 *.mem
 *.so
 Manifest.toml
+LocalPreferences.toml
```
giordano, line 8: Missing newline
```suggestion
LocalPreferences.toml

```

src/MPI.jl
```diff
@@ -33,14 +33,49 @@ function _doc_external(fname)
...
 
-try
-    include(joinpath(dirname(@__DIR__), "deps", "deps.jl"))
-    include(joinpath(dirname(@__DIR__), "deps", "compile_time_mpi_constants.jl"))
-catch e
-    error("MPI.jl not properly configured, please run `Pkg.build(\"MPI\")`.")
+
+import MPIPreferences
+
+@static if MPIPreferences.binary == "MPICH_jll"
```
giordano, line 39: My understanding is that a top-level `@static` is redundant, the expression is executed at precompile time anyway.

src/consts/consts.jl
```diff
@@ -0,0 +1,46 @@
...
+    include("mpiwrapper.jl")
+else
+    error("Unknown MPI ABI $(MPIPreferences.abi)")
+end
+
+@eval function __init__()
+    $(Expr(:block, initexprs...))
+end
+
+end
```
giordano, line 46: Missing newline
```suggestion
end

```

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -0,0 +1,181 @@
...
+            version = VersionNumber(m.captures[1])
+        end
+
+    elseif occursin("CRAY MPICH", version_string)
+        impl = "CrayMPICH"
+        # "MPI VERSION    : CRAY MPICH version 7.7.10 (ANL base 3.2)\n"
+        if (m = match(r"CRAY MPICH version (\d+.\d+.\d+)", version_string)) !== nothing
+            version = VersionNumber(m.captures[1])
+        end
+    elseif startswith(version_string, "FUJITSU MPI")
```
giordano, line 146: Ok, this breaks the idea of #542 :smile:

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -0,0 +1,181 @@
...
+        "libmpi" => nothing,
+        "abi" => nothing,
+        "mpiexec" => nothing;
+        export_prefs=export_prefs,
+        force=force
+    )
+end
+
+function use_system_binary(;
+        library=["libmpi", "libmpi_ibm", "msmpi", "libmpich", "libmpitrampoline"],
```
giordano, line 44: Maybe `library_names` is more descriptive?

giordano (commented):
I think I missed how the constants have been generated.  It'd be good to have a reproducible way to update them, if needed.

Also, the [documentation](https://juliaparallel.github.io/MPI.jl/latest/configuration/) should also be updated, right?

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -0,0 +1,181 @@
+module MPIPreferences
+
+using Preferences, Libdl
+
+const binary = @load_preference("binary", Sys.iswindows() ? "MicrosoftMPI_jll" : "MPICH_jll")
```
simonbyrne, line 5: I wasn't sure if this should be a function or a `const` binding.

vchuravy: I think a const binding is fine, and will trigger recompilation if it changes.

src/MPI.jl
vchuravy: Indeed!

vchuravy:
@JuliaRegistrator register subdir=lib/MPIPreferences

lib/MPIPreferences/src/MPIPreferences.jl
vchuravy, line 44:
```suggestion
        library_names=["libmpi", "libmpi_ibm", "msmpi", "libmpich", "libmpitrampoline"],
```

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -0,0 +1,181 @@
...
+
+function use_system_binary(;
+        library=["libmpi", "libmpi_ibm", "msmpi", "libmpich", "libmpitrampoline"],
+        mpiexec="mpiexec",
+        abi=nothing,
+        export_prefs=false,
+        force=true,
+    )
+
+    libmpi = find_library(library)
```
vchuravy, line 51:
```suggestion
    libmpi = find_library(library_names)
```

src/MPI.jl
vchuravy, line 39:
```suggestion
if MPIPreferences.binary == "MPICH_jll"
```

src/consts/microsoftmpi.jl
```diff
@@ -0,0 +1,334 @@
...
+    MPI_SOURCE::Cint
+    MPI_TAG::Cint
+    MPI_ERROR::Cint
+end
+
+# Opaque handles:
+const MPI_Comm = UInt32
+const MPI_Datatype = UInt32
+const MPI_Errhandler = UInt32
+const MPI_File = UInt32
```
vchuravy, line 34:
```suggestion
const MPI_File = UInt
```

docs/src/knownissues.md
```diff
@@ -34,6 +34,12 @@ This can be worked around be either:
 
 2. Launching julia with the  `--compiled-modules=no` option. This can result in much longer package load times.
 
+## Open MPI
+
+### Segmentation fault
+
+When attempting to use a system-provided Open MPI implementation, you may encounter a segmentation fault. This can be fixed by setting the environment variable `ZES_ENABLE_SYSMAN=1`. See [Open MPI issue #10142](https://github.com/open-mpi/ompi/issues/10142).
```
giordano, line 41: This doesn't happen with `OpenMPI_jll`?

vchuravy: Yes because `OpenMPI_jll` uses an embedded HWloc instead of `HWloc_jll`

```
julia> dlopen("/home/vchuravy/.julia/artifacts/aae9ba66be9f6bccbfa324bbab4fa317b676b9d2/lib/libhwloc.so")
Ptr{Nothing} @0x0000000001ec63a0

julia> dlclose(ans)

````

has this exact problem.

vchuravy (approved):

docs/src/knownissues.md
simonbyrne: From https://github.com/open-mpi/ompi/issues/10142#issuecomment-1072367339:

> Open MPI embeds a copy of hwloc (which, for the purposes of this discussion, is just a library), but that embedded copy of hwloc is only used if hwloc is not already available on your system. The version of hwloc that is included in Open MPI v4.1.2 is hwloc v2.0.2, which is long before the ZES_ENABLE_SYSMAN issue (i.e., it won't crash because of this).

ggouaillardet: note that even if you are using a recent `hwloc` release that has this issue, the crash only occurs when `hwloc` is dlclose'd **and** the environment is accessed.

bgoglin: hwloc 2.7.1rc1 is available from https://www.open-mpi.org/software/hwloc/v2.7/
I'll release the final 2.7.1 in a couple days in you can confirm the issue is gone (it should only exist with hwloc >=2.5 and <=2.7.0).

Issue #539, opened 2022-02-14
Unknown MPI ABI when using Fujitsu MPI on Fugaku, and segmentation faults when running the tests with this implementation


giordano:
```
$ julia --project -e 'ENV["JULIA_MPI_BINARY"]="system"; using Pkg; Pkg.build("MPI"; verbose=true)'
    Building MPI → `~/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/d56a80d8cf8b9dc3050116346b3d83432b1912c0/build.log`
[ Info: using system MPI                             ]  0/1
┌ Info: Using implementation
│   libmpi = "libmpi"
│   mpiexec_cmd = `mpiexec`
└   MPI_LIBRARY_VERSION_STRING = "FUJITSU MPI Library 4.0.0 (4.0.1fj4.0.0)\0"
┌ Info: MPI implementation detected
│   impl = UnknownMPI::MPIImpl = 0
│   version = v"0.0.0"
└   abi = "unknown"
[ Info: Unknown MPI ABI: building constants file
ERROR: LoadError: IOError: could not spawn `mpicc gen_consts.c -o gen_consts -lmpi -L/opt/FJSVxtclanga/tcsds-1.2.34/lib64`: no such file or directory (ENOENT)
Stacktrace:
  [1] _spawn_primitive(file::String, cmd::Cmd, stdio::Vector{Any})
    @ Base ./process.jl:100
  [2] #690
    @ ./process.jl:113 [inlined]
  [3] setup_stdios(f::Base.var"#690#691"{Cmd}, stdios::Vector{Any})
    @ Base ./process.jl:197
  [4] _spawn
    @ ./process.jl:112 [inlined]
  [5] run(::Cmd; wait::Bool)
    @ Base ./process.jl:445
  [6] run(::Cmd)
    @ Base ./process.jl:444
  [7] top-level scope
    @ ~/.julia/packages/MPI/08SPr/deps/gen_consts.jl:231
  [8] include(fname::String)
    @ Base.MainInclude ./client.jl:451
  [9] top-level scope
    @ ~/.julia/packages/MPI/08SPr/deps/build.jl:114
 [10] include(fname::String)
    @ Base.MainInclude ./client.jl:451
 [11] top-level scope
    @ none:5
in expression starting at /home/ra000019/a04463/.julia/packages/MPI/08SPr/deps/gen_consts.jl:231
[...]

PR #542, opened 2022-02-15
Do not hard fail in `test_errorhandler.jl` with unknown implementations

giordano:
Ref: https://github.com/JuliaParallel/MPI.jl/issues/539#issuecomment-1039798163.  The idea is to not make tests hard-fail on custom error handlers with unknown ABIs, since it's known that they may not be implemented.  @simonbyrne what do you think about this?  With this change, tests of `MPI.jl` v0.19.2 are successful on Fugaku with system (Fujitsu) MPI, albeit with a broken test.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -50,7 +50,9 @@ testfiles = sort(filter(istest, readdir(testdir)))
         elseif f == "test_error.jl"
             r = run(ignorestatus(cmd))
             @test !success(r)
-        elseif f == "test_errorhandler.jl" && MPI.identify_implementation()[1] == MPI.UnknownMPI
+        elseif f == "test_errorhandler.jl" && (MPI.identify_implementation()[1] == MPI.UnknownMPI ||
+            # Fujitsu MPI is known to not work with custom error handlers
+            startswith(MPI.MPI_LIBRARY_VERSION_STRING, "FUJITSU MPI"))
             try
                 run(cmd)
             catch e
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -36,16 +36,30 @@ istest(f) = endswith(f, ".jl") && startswith(f, "test_")
 testfiles = sort(filter(istest, readdir(testdir)))
 
 @testset "$f" for f in testfiles
-    mpiexec() do cmd
+    mpiexec() do mpirun
+        cmd = `$mpirun -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`
         if f == "test_spawn.jl"
-            run(`$cmd -n 1 $(Base.julia_cmd()) $(joinpath(testdir, f))`)
+            # Some command as the others, but always use a single process
+            _cmd = deepcopy(cmd)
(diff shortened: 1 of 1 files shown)


test/runtests.jl
```diff
@@ -25,24 +25,38 @@ istest(f) = endswith(f, ".jl") && startswith(f, "test_")
...
         elseif f == "test_threads.jl"
             withenv("JULIA_NUM_THREADS" => "4") do
-                run(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`)
+                run(cmd)
             end
         elseif f == "test_error.jl"
-            r = run(ignorestatus(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`))
+            r = run(ignorestatus(cmd))
             @test !success(r)
+        elseif f == "test_errorhandler.jl" && MPI.identify_implementation()[1] == MPI.UnknownMPI
```
simonbyrne, line 42: We should provide some more generic mechanism for excluding specific tests (or allowing them to fail)

vchuravy (approved):

test/runtests.jl
```diff
@@ -34,24 +34,40 @@ istest(f) = endswith(f, ".jl") && startswith(f, "test_")
...
-                run(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`)
+                run(cmd)
             end
         elseif f == "test_error.jl"
-            r = run(ignorestatus(`$cmd -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`))
+            r = run(ignorestatus(cmd))
             @test !success(r)
+        elseif f == "test_errorhandler.jl" && (MPI.identify_implementation()[1] == MPI.UnknownMPI ||
+            # Fujitsu MPI is known to not work with custom error handlers
+            startswith(MPI.MPI_LIBRARY_VERSION_STRING, "FUJITSU MPI"))
```
giordano, line 53: Would this be OK for the time being?  We allow failures for unknown MPIs and Fujitsu MPI, but the error message is still shown.

test/runtests.jl
```diff
@@ -34,24 +34,40 @@ istest(f) = endswith(f, ".jl") && startswith(f, "test_")
 testfiles = sort(filter(istest, readdir(testdir)))
 
 @testset "$f" for f in testfiles
-    mpiexec() do cmd
+    mpiexec() do mpirun
+        cmd = `$mpirun -n $nprocs $(Base.julia_cmd()) $(joinpath(testdir, f))`
```
vchuravy, line 38:
```suggestion
        cmd(n=nprocs) = `$mpirun -n $n $(Base.julia_cmd()) $(joinpath(testdir, f))`
```

test/runtests.jl
vchuravy: Yeah I am okay with this!

test/runtests.jl
giordano: Oh wow, that was so obvious in hindsight :smile:  Thanks!

PR #553, opened 2022-04-14
Add RMA examples to docs 

carstenbauer:
Adds one-sided communication (active and passive RMA) examples to the docs. Building the docs worked locally. Feel free to suggest improvements to the examples or stylistic changes.

PS: I have more instructive examples like sending a subpart of a matrix via 1) views 2) explicitly using derived types etc. Should I prepare / add those as well? What do you think?


simonbyrne:
Thanks!

> PS: I have more instructive examples like sending a subpart of a matrix via 1) views 2) explicitly using derived types etc. Should I prepare / add those as well? What do you think?

Certainly!

Issue #543, opened 2022-02-19
Reorganise interface of `MPI.mpiexec`


giordano:
Prompted by an issue faced yesterday by @vchuravy, we should change the interface of `MPI.mpiexec` to more closely resemble the new interface of [`ExecutableProduct`](https://docs.binarybuilder.org/stable/jll/#ExecutableProduct)s in JLL packages.  In particular:

* the syntax to run the executable should be
  ```julia
  run(`$(mpiexec()) $arguments`)
  ```
  instead of
  ```julia
  mpiexec() do mpiexec
      run(`$mpiexec $arguments`)
  end
  ```
  The new syntax has the advantage of being (a) much shorter and (b) thread-safe (although usually you wouldn't run MPI processes within multiple threads)
* add the `adjust_LIBPATH::Bool` keyword argument, like the [`JLLWrappers`](https://github.com/JuliaPackaging/JLLWrappers.jl/blob/bade7d4b7af3e1c0c590bfd383a3df5ed10852f8/src/products/executable_generators.jl#L6) do.  This would solve Valentin's issue because when using the default MPICH from JLLs, the wrapper would set by default `LD_LIBRARY_PATH`, which screwed up some libraries.  This argument can perhaps remain undocumented if we want.

Note that this is a breaking change.

simonbyrne:
Does it need to be breaking? We don't define a no-arg method

sloede:
I assume all packages that use MPI during CI testing use it (like we do). Thus if there are no strong arguments against keeping the old syntax around, it would be nice to make it non breaking.

simonbyrne:
my plan in #541 is to import `mpiexec` directly, so that should fix this.

is there a link to @vchuravy's issue?

giordano:
Not on slack, just a (long) [thread on Slack](https://julialang.slack.com/archives/C674ELDNX/p1645131463952399).

The gist is that due to `LD_LIBRARY_PATH` being set by `MPICH_jll.mpiexec(;adjust_LIBPATH=true)`, `/usr/lib` was in that environment variable, and `using Pango_jll` would mix system libraries and JLL libraries, wrecking havoc.

simonbyrne:
This should now be resolved. I've opened #556 so that the when using the system MPI the arguments are compatible.

PR #556, opened 2022-04-15
Add adjust_PATH and adjust_LIBPATH to System mpiexec

simonbyrne:
Makes it compatible with JLLWrappers (see #543)


vchuravy:
For CI to work we need #552

vchuravy (approved):

PR #557, opened 2022-04-15
Enable GPU CI

vchuravy:
- Improve CUDA-aware MPI support and add CI
- test hypothesis


src/cuda.jl
```diff
@@ -1,6 +1,7 @@
 import .CUDA
 
 function Base.cconvert(::Type{MPIPtr}, buf::CUDA.CuArray{T}) where T
+    CUDA.synchronize(CUDA.stream())
```
vchuravy, line 4:
```suggestion
```

PR #565, opened 2022-04-16
openmpi needs to be opened with RTLD_GLOBAL

### LINKED ISSUE #560: MPIPreferences segfaults
Running `MPIPreferences.use_system_binary()` segfaults if `using MPI` was not executed before. This behaviour is observed on MPI 0.20-dev (master) branch with Julia 1.7.1 and Julia 1.7.2 on Linux (Ubuntu 20.04).

vchuravy:
May fix #560

### DIFF
--- a/lib/MPIPreferences/src/MPIPreferences.jl
+++ b/lib/MPIPreferences/src/MPIPreferences.jl
@@ -82,9 +82,21 @@ function identify_abi(libmpi)
     # implementations.
     buf = Array{UInt8}(undef, 8192)
     buflen = Ref{Cint}()
-    dlopen(libmpi) do hdl
+
+    @static if Sys.isunix()
+        # need to open libmpi with RTLD_GLOBAL flag for Linux, before
+        # any ccall cannot use RTLD_DEEPBIND; this leads to segfaults
+        # at least on Ubuntu 15.10
+        hndl = Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
+    else
+        hndl = Libdl.dlopen(libmpi)
+    end
+
+    try
         ptr = dlsym(hdl, :MPI_Get_library_version)
         ccall(ptr, Cint, (Ptr{UInt8}, Ref{Cint}), buf, buflen)
+    finally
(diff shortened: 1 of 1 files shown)


vchuravy:
@luraess could you verify that this works for you?

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -82,9 +82,21 @@ function identify_abi(libmpi)
...
+        # need to open libmpi with RTLD_GLOBAL flag for Linux, before
+        # any ccall cannot use RTLD_DEEPBIND; this leads to segfaults
+        # at least on Ubuntu 15.10
+        hndl = Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
+    else
+        hndl = Libdl.dlopen(libmpi)
+    end
+
+    try
         ptr = dlsym(hdl, :MPI_Get_library_version)
```
vchuravy, line 96:
```suggestion
        ptr = dlsym(hndl, :MPI_Get_library_version)
```

luraess:
Testing with update still segfaults with same behaviour and versions, as described in #560.

simonbyrne:
Ah, this could be https://github.com/open-mpi/ompi/issues/10142

Try setting the environment variable
```
export ZES_ENABLE_SYSMAN=1
```

luraess:
Seems to be the case @simonbyrne . Your suggestion avoids segfault to occur. I guess this could/should be documented (@vchuravy maybe in your refactored docs?). Is  MPI.jl fix possible or is this purely OMPI issue?

simonbyrne:
> Is MPI.jl fix possible or is this purely OMPI issue?

No, unfortunately it's not fixable on our side. We should document it though.

luraess:
Thanks for the input!

> We should document it though.

Absolutely-

giordano:
That's already documented, isn't it: https://juliaparallel.org/MPI.jl/latest/knownissues/#Segmentation-fault?

simonbyrne:
Haha, I had forgotten I wrote that.

Issue #561, opened 2022-04-16
MPI test do not select the correct (system) MPI implementation


luraess:
Testing MPI v0.20-dev (master) branch on Linux (Ubuntu 20.04) with system MPI 
```
julia> MPIPreferences.use_system_binary()
┌ Info: MPI implementation
│   libmpi = "libmpi"
│   version_string = "Open MPI v4.1.2, package: Open MPI luraess@superzack Distribution, ident: 4.1.2, repo rev: v4.1.2, Nov 24, 2021\0"
│   impl = "OpenMPI"
│   version = v"4.1.2"
└   abi = "OpenMPI"
```
does not pick up the correct MPI implementation although `LocalPreferences.toml` contains the above info. Tests would always pick up MPICH_jll
```
┌ Info: Running MPI tests
│   ArrayType = Array
│   nprocs = 1
│   MPIPreferences.abi = "MPICH"
└   MPIPreferences.binary = "MPICH_jll"
```
Copying manually `LocalPreferences.toml` does not solve the issue.

Note that MPI selects the correct implementation outside of test
```
julia> MPI.MPI_LIBRARY_VERSION_STRING
"Open MPI v4.1.2, package: Open MPI luraess@superzack Distribution, ident: 4.1.2, repo rev: v4.1.2, Nov 24, 2021\0"
```

vchuravy:
What version of Julia are you using?

luraess:
```
julia> versioninfo()
Julia Version 1.7.2
Commit bf53498635 (2022-02-06 15:21 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: AMD EPYC 7282 16-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, znver2)
  ```
  Same issue with Spack-built v1.7.1

simonbyrne:
Is this worth fixing, or should we just add a section to the docs on how to run the tests (copy the LocalPreferences.jl into your test directory)

luraess:
Is this now fixed by #564? Seems that https://github.com/JuliaPackaging/Preferences.jl/pull/37#event-6460508912 and https://github.com/JuliaLang/Pkg.jl/pull/3061, as flagged some months ago to me by @vchuravy got merged (don't know the up to date status though @simonbyrne).

PR #566, opened 2022-04-16
Rewrite configuration docs for MPIPreferences

vchuravy:
- Re-export use_jll_binary and use_system_binary
- Rewrite configuration docs for MPIPreferences

### DIFF
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -86,11 +86,11 @@ a file called `LocalPreferences.toml` adjacent to the current `Project.toml`.
 julia --project -e 'using MPI; MPI.use_system_binary()'
 ```
 !!! note
-    You can copy `LocalPreferences.toml` to a different project, but you must list
-    `MPIPreferences` in the Project.toml either in `[deps]` or `[extras]` for the settings
+    You can copy `LocalPreferences.toml` to a different project folder, but you must list
+    `MPIPreferences` in the `[extras]` section of the `Project.toml` for the settings
     to take effect. Due to a bug in Julia (until `v1.6.5` and `v1.7.1`), getting preferences
-    gtom transitive dependencies is broken (https://github.com/JuliaPackaging/Preferences.jl/issues/24).
-    To fix this either update your version of Julia, or add `MPIPreferences` as a direct dependency to your project.
+    from transitive dependencies is broken (https://github.com/JuliaPackaging/Preferences.jl/issues/24).
+    To fix this update your version of Julia, or add `MPIPreferences` as a direct dependency to your project.
 
 The MPI standard doesn't specify the exact application binary interface (ABI).
 
@@ -107,7 +107,7 @@ The following MPI implementations should work out-of-the-box with MPI.jl:
 ```@doc
 MPI.use_system_binary
 ```
-You can use the argument `mpiexec` to provide the the name (or full path) of the MPI launcher executable. The default is
+You can use the argument `mpiexec` to provide the name (or full path) of the MPI launcher executable. The default is
 `mpiexec`, but some clusters require using the scheduler launcher interface (e.g. `srun`
 on Slurm, `aprun` on PBS). If the MPI library has an uncommon name you can provide it in `library_names`.
 In case ABI detection fails you can provide a manual choice,
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -92,7 +92,6 @@ julia --project -e 'using MPI; MPI.use_system_binary()'
     from transitive dependencies is broken (https://github.com/JuliaPackaging/Preferences.jl/issues/24).
     To fix this update your version of Julia, or add `MPIPreferences` as a direct dependency to your project.
 
-The MPI standard doesn't specify the exact application binary interface (ABI).
 
 The following MPI implementations should work out-of-the-box with MPI.jl:
 
@@ -110,6 +109,7 @@ MPI.use_system_binary
 You can use the argument `mpiexec` to provide the name (or full path) of the MPI launcher executable. The default is
 `mpiexec`, but some clusters require using the scheduler launcher interface (e.g. `srun`
 on Slurm, `aprun` on PBS). If the MPI library has an uncommon name you can provide it in `library_names`.
+The MPI standard does not specify the exact application binary interface (ABI).
 In case ABI detection fails you can provide a manual choice,
 but also open an issue so that the auto-detection can be improved. `export_prefs=true` can be used to copy the preferences into the `Project.toml` instead of creating a
 `LocalPreferences.toml` file to hold them.
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -110,8 +110,9 @@ You can use the argument `mpiexec` to provide the name (or full path) of the MPI
 `mpiexec`, but some clusters require using the scheduler launcher interface (e.g. `srun`
 on Slurm, `aprun` on PBS). If the MPI library has an uncommon name you can provide it in `library_names`.
 The MPI standard does not specify the exact application binary interface (ABI).
-In case ABI detection fails you can provide a manual choice,
-but also open an issue so that the auto-detection can be improved. `export_prefs=true` can be used to copy the preferences into the `Project.toml` instead of creating a
+In case ABI detection fails you can provide a manual choice (either `MPICH`, `MPItrampoline`, `OpenMPI`, or `MicrosoftMPI`),
+but also open an issue such that the automatic detection can be improved.
+`export_prefs=true` can be used to copy the preferences into the `Project.toml` instead of creating a
 `LocalPreferences.toml` file to hold them.
 
 #### Notes to HPC cluster adminstators
@@ -158,5 +159,5 @@ The test suite can also be modified by the following variables:
 - `JULIA_MPI_TEST_NPROCS`: How many ranks to use within the tests
 - `JULIA_MPI_TEST_ARRAYTYPE`: Set to `CuArray` to test the CUDA-aware interface with
   [`CUDA.CuArray](https://github.com/JuliaGPU/CUDA.jl) buffers.
-- `JULIA_MPI_TEST_BINARY`: Check that the correct MPI binary is used for the tests
-- `JULIA_MPI_TEST_ABI`: Check that the correct MPI ABI is used for the tests
+- `JULIA_MPI_TEST_BINARY`: Check that the specified MPI binary is used for the tests
+- `JULIA_MPI_TEST_ABI`: Check that the specified MPI ABI is used for the tests
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -9,10 +9,6 @@ clusters or multi-GPU machines, you will probably want to configure against a
 system-provided MPI implementation in order to exploit features such as fast network
 interfaces and CUDA-aware MPI interfaces.
 
-MPI.jl will attempt to detect when you are running on a HPC cluster, and warn the user
-about this. To disable this warning, set the environment variable
-`JULIA_MPI_CLUSTER_WARN=n`.
-
 ## Julia wrapper for `mpiexec`
 
 Since you can configure `MPI.jl` to use one of several MPI implementations, you
@@ -52,26 +48,51 @@ with:
 $ mpiexecjl --project=/path/to/project -n 20 julia script.jl
 ```
 
-## Using a system-provided MPI
+## Using MPIPreferences.jl
 
-### Requirements
+MPI.jl uses [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl) to
+allow the user to choose which MPI implementation to use for a project. This provides
+a single source of truth that can be used for `JLL` (Julia packages providing C libraries)
+that link against MPI, as well as localizing the choice of MPI implementation to a project.
+
+Users can use the provided [`use_system_binary`](@ref) or [`use_jll_binary`](@ref)
+to switch MPI implementations.
+
+### Migration from MPI.jl `v0.19`
+
+Prior to MPI.jl `v0.20` environment variables were used to configure which MPI
+library to use. These have now been removed.
+
+- `JULIA_MPI_BINARY`
+- `JULIA_MPIEXEC`
+- `JULIA_MPI_INCLUDE_PATH`
+- `JULIA_MPI_CFLAGS`
+- `JULIA_MPICC`
+
+### Using a system-provided MPI
+
+#### Requirements
 
 MPI.jl requires a shared library installation of a C MPI library, supporting the MPI 3.0
 standard or later.
 
-### Building
+### Configuration
 
-To use the the system MPI, set the environment variable `JULIA_MPI_BINARY=system` and run
-`Pkg.build("MPI")`. This can be done by:
+To use the the system MPI, run `MPI.use_system_binary()`.
+This will attempt find and identify any available MPI implementation, and create
+a file called `LocalPreferences.toml` adjacent to the current `Project.toml`.
+
+```sh
+julia --project -e 'using MPI; MPI.use_system_binary()'
 ```
-julia --project -e 'ENV["JULIA_MPI_BINARY"]="system"; using Pkg; Pkg.build("MPI"; verbose=true)'
-```
-This will attempt find and identify any available MPI implementation.
+!!! note
+    You can copy `LocalPreferences.toml` to a different project, but you must list
+    `MPIPreferences` in the Project.toml either in `[deps]` or `[extras]` for the settings
+    to take effect. Due to a bug in Julia (until `v1.6.5` and `v1.7.1`), getting preferences
+    gtom transitive dependencies is broken (https://github.com/JuliaPackaging/Preferences.jl/issues/24).
+    To fix this either update your version of Julia, or add `MPIPreferences` as a direct dependency to your project.
 
 The MPI standard doesn't specify the exact application binary interface (ABI).
-The build script will attempt to build a small C program to
-determine the appropriate type definitions and constants. This requires a compatible C
-compiler (`mpicc` by default).
 
 The following implementations should work:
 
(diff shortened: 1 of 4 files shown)


docs/src/configuration.md
```diff
@@ -52,26 +48,51 @@ with:
...
+a file called `LocalPreferences.toml` adjacent to the current `Project.toml`.
+
+```sh
+julia --project -e 'using MPI; MPI.use_system_binary()'
 ```
-This will attempt find and identify any available MPI implementation.
+!!! note
+    You can copy `LocalPreferences.toml` to a different project, but you must list
+    `MPIPreferences` in the Project.toml either in `[deps]` or `[extras]` for the settings
+    to take effect. Due to a bug in Julia (until `v1.6.5` and `v1.7.1`), getting preferences
```
sloede, line 91: Is `MPIPreferences` added automatically if `use_system_binary()` is called? If yes, this should be stated somewhere explicitly. Otherwise, it should me mentioned somewhere that this needs to be added manually.

It should also be stated somewhere that when switching back to the JLL binaries are used again, one can also safely delete the `LocalPreferences.jl` file. 

Also, be explicit about where it should be added: Pick either `[deps]` or `[extras]` and remove the burden of choice from the user. If it's all the same, I'd suggest `[deps]` since it's what is more well-known; one should only propagate adding it to `[extras]` if this is conceptually the "better" place.

docs/src/configuration.md
```diff
@@ -52,26 +48,51 @@ with:
...
+    gtom transitive dependencies is broken (https://github.com/JuliaPackaging/Preferences.jl/issues/24).
+    To fix this either update your version of Julia, or add `MPIPreferences` as a direct dependency to your project.
```
sloede, line 93: I don't understand the logic here. So do I always have to add `MPIPreferences` or only sometimes? This somehow does not align with the first part of the note.

docs/src/configuration.md
```diff
@@ -52,26 +48,51 @@ with:
...
 
 The MPI standard doesn't specify the exact application binary interface (ABI).
```
sloede, line 95:
```suggestion
```
Is this sentence still useful/necessary here?

docs/src/configuration.md
```diff
@@ -52,26 +48,51 @@ with:
...
-The build script will attempt to build a small C program to
-determine the appropriate type definitions and constants. This requires a compatible C
-compiler (`mpicc` by default).
 
 The following implementations should work:
```
sloede, line 97:
```suggestion

The following MPI implementations should work out-of-the-box with MPI.jl:
```

docs/src/configuration.md
```diff
@@ -81,35 +102,59 @@ The following implementations should work:
...
 
-### [Environment variables](@id environment_variables)
+```@doc
+MPI.use_system_binary
+```
+You can use the argument `mpiexec` to provide the the name (or full path) of the MPI launcher executable. The default is
+`mpiexec`, but some clusters require using the scheduler launcher interface (e.g. `srun`
+on Slurm, `aprun` on PBS). If the MPI library has an uncommon name you can provide it in `library_names`.
+In case ABI detection fails you can provide a manual choice,
+but also open an issue so that the auto-detection can be improved. `export_prefs=true` can be used to copy the preferences into the `Project.toml` instead of creating a
```
sloede, line 114:
```suggestion
In case ABI detection fails you can provide a manual choice (either `MPICH` or `OpenMPI` or `MicrosoftMPI`),
but also open an issue such that the automatic detection can be improved.
`export_prefs=true` can be used to copy the preferences into the `Project.toml` instead of creating a
```
I think these three are the only ones supported, right? Further, why `export_prefs` and not `export_preferences`? Ok, that's probably a personal style question, I try to avoid abbreviations where not necessary. Finally, I think we should state a reason why anyone would want to use `export_prefs`, i.e., what are the possible advantages.

docs/src/configuration.md
```diff
@@ -81,35 +102,59 @@ The following implementations should work:
...
+`LocalPreferences.toml` file to hold them.
+
+#### Notes to HPC cluster adminstators
+
+Preferences are merged across the Julia load path, so it is feasible to provide a module file that appends a path to
+`JULIA_LOAD_PATH` variable that contains system wide preferences.
```
sloede, line 120:
```suggestion
Preferences are merged across the Julia load path, such that it is feasible to provide a module file that appends a path to
`JULIA_LOAD_PATH` variable that contains system-wide preferences.
```

docs/src/configuration.md
```diff
@@ -81,35 +102,59 @@ The following implementations should work:
...
-- `JULIA_MPI_PATH`: the top-level installation directory of MPI. i.e. the library should
-  be located in `${JULIA_MPI_PATH}/lib` and `mpiexec` in `${JULIA_MPI_PATH}/bin`
-- `JULIA_MPI_LIBRARY`: the library name or full path of the MPI shared library. By
-  default, it will attempt to look for common MPI library names in the standard library
-  paths (e.g. `libmpi`, `libmpich`, `msmpi`).
-- `JULIA_MPIEXEC`: the name (or full path) of the MPI launcher executable. The default is
-  `mpiexec`, but some clusters require using the scheduler launcher interface (e.g. `srun`
-  on Slurm, `aprun` on PBS).
-- `JULIA_MPIEXEC_ARGS`: Additional arguments to be passed to MPI launcher.
+```
```
sloede, line 124:
~~~suggestion
```toml
~~~

File: docs/src/configuration.md [hunk 8]
```diff
@@ -81,35 +102,59 @@ The following implementations should work:
...
+#### Notes to HPC cluster adminstators
+
+Preferences are merged across the Julia load path, so it is feasible to provide a module file that appends a path to
+`JULIA_LOAD_PATH` variable that contains system wide preferences.
 
-The following optional environment variables can be used to control certain aspects of the
-build script and other library behaviour. The results of these will be cached in a
-configuration file located at `~/.julia/prefs/MPI.toml` and so can be used for subsequent
-MPI builds.
+As an example you can use [`MPI.use_system_binary(;export_prefs)`](@ref) to create a file `Project.toml` containing:
```
sloede, line 122: Why the `export_prefs`? Why can't I just copy the `LocalPreferences.toml` to the central location? If I have to use `export_prefs=true`, it should be stated explicitly as such.

docs/src/configuration.md
```diff
@@ -81,35 +102,59 @@ The following implementations should work:
...
+MPIPreferences = "3da0fdf6-3ccc-4f1b-acd9-58baa6c99267"
+
+[preferences.MPIPreferences]
+abi = "OpenMPI"
+binary = "system"
+libmpi = "/software/mpi/lib/libmpi.so"
+mpiexec = "/software/mpi/bin/mpiexec"
+```
+copying that to a central location like: `/software/mpi/julia` and setting `JULIA_LOAD_PATH=":/software/mpi/julia"` (note the `:` before the path) in the coressponding module file (preferrably the MPI installation, or the Julia module)
+, will cause the user to default to your cluster MPI installation.
```
sloede, line 135:
```suggestion
Copying this `Project.toml` to a central location such as `/software/mpi/julia` and setting the environment
variable `JULIA_LOAD_PATH=":/software/mpi/julia"` (note the `:` before the path) in the corresponding
module file (preferably the module file for the MPI installation or for Julia),
will cause the user to default to your cluster MPI installation.
```

docs/src/configuration.md
```diff
@@ -81,35 +102,59 @@ The following implementations should work:
...
 
-The following variables are also queried:
+The user can still create a differing per Julia project that
+will take precedent.
```
sloede, line 138:
```suggestion
The user can still provide differing MPI configurations for each Julia project that
will take precedent by modifying the local `Project.toml` or by providing a `LocalPreferences.toml` file.
```
For clarity?

docs/src/configuration.md
```diff
@@ -81,35 +102,59 @@ The following implementations should work:
...
 
-- `JULIA_MPI_INCLUDE_PATH`: the directory containing the MPI header files.
-- `JULIA_MPI_CFLAGS`: C flags passed to the constant generation build (default: `-lmpi`)
-- `JULIA_MPICC`: MPI C compiler (default: `mpicc`)
+### Using a different JLL provided MPI library
+
+The following JLL implementations are provided:
```
sloede, line 142:
```suggestion
The following MPI implementations are provided as JLL packages and automatically obtained when installing MPI.jl:
```

docs/src/configuration.md
```diff
@@ -81,35 +102,59 @@ The following implementations should work:
...
+
+- `MicrosoftMPI_jll`: Default for Windows
+- `MPICH_jll`: Default on all Unix systems
```
sloede, line 145:
```suggestion
- `MPICH_jll`: Default for all Unix-like systems
```

docs/src/configuration.md
```diff
@@ -81,35 +102,59 @@ The following implementations should work:
...
 
+## Environment variables for the testsuite
 The test suite can also be modified by the following variables:
 
-- `JULIA_MPIEXEC_TEST_ARGS`: Additional arguments to be passed to the MPI launcher for the tests only.
+- `JULIA_MPI_TEST_NPROCS`: How many ranks to use within the tests
 - `JULIA_MPI_TEST_ARRAYTYPE`: Set to `CuArray` to test the CUDA-aware interface with
   [`CUDA.CuArray](https://github.com/JuliaGPU/CUDA.jl) buffers.
+- `JULIA_MPI_TEST_BINARY`: Check that the correct MPI binary is used for the tests
+- `JULIA_MPI_TEST_ABI`: Check that the correct MPI ABI is used for the tests
```
sloede, line 160: From the docs it is unclear what I need to do with these variables. Set them to `true`?

sloede (commented):
Thanks a lot for the documentation - I think this is making great progress towards usability of the new preferences system! I have left a few comments where I think the current draft can be improved. Please take everything as a suggestion and feel free to ask if I should clarify something.

docs/src/configuration.md
```diff
@@ -52,26 +48,51 @@ with:
...
-### Building
+### Configuration
 
-To use the the system MPI, set the environment variable `JULIA_MPI_BINARY=system` and run
-`Pkg.build("MPI")`. This can be done by:
-```
-julia --project -e 'ENV["JULIA_MPI_BINARY"]="system"; using Pkg; Pkg.build("MPI"; verbose=true)'
+To use the the system MPI, run `MPI.use_system_binary()`.
+This will attempt find and identify any available MPI implementation, and create
+a file called `LocalPreferences.toml` adjacent to the current `Project.toml`.
```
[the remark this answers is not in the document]
vchuravy: You are always using a project :), but yes it would be created in `.julia/environment/v1.7` if you are using the default project.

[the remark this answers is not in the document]
sloede: I know, but it took me a while to understand this fact. I just thought it might help users to avoid the "huh, where is the newly created preferences file?" moment. But it is also sort of redundant to the basic Julia docs, so feel free to ignore this suggestion if you do not like it.

docs/src/configuration.md
vchuravy: Yes MPIPreferences is added automatically to the `[extras]` section. Entries in `[deps]` need corresponding entries in `[compat]` (if one wants to register the package) and makes MPIPreferences implicitly loadable, so from my perspective `[extra]` is preferred.

docs/src/configuration.md
vchuravy: Yeah this is a wart... On older Julia versions the extra section didn't work properly (Need to double check this as well)

docs/src/configuration.md
vchuravy: I think so, at least to explain the word ABI

docs/src/configuration.md
vchuravy: You are missing MPI trampoline.

`export_prefs` is inherited from Preferences.jl we just pass through the keyword argument.

Export preferences puts it into the Project.toml so it is one less file to copy around. Other than that there shouldn't be a difference

docs/src/configuration.md
sloede: OK, in that case I suggest to rewrite this note to include this information, e.g.,
```suggestion
    You can copy `LocalPreferences.toml` to a different project, but you must list
    `MPIPreferences` in the Project.toml either in the `[extra]` section for the settings
    to take effect. Due to a bug in Julia (until `v1.6.5` and `v1.7.1`), getting preferences
```
but also state somewhere that `MPIPreferences` are automatically added to the Project.toml when `use_system_binary()` is called.

docs/src/configuration.md
vchuravy: I only wanted to show one file instead of two, but you are probably right. I personally prefer the LocalPreferences.toml

docs/src/configuration.md
vchuravy: Need to be set to what you expect the variable to be.

E.g. `JULIA_MPI_TEST_BINARY=binary` and the test will fail if you are using a jll

docs/src/configuration.md
sloede: I have updated the suggestion

docs/src/configuration.md
sloede: If `LocalPreferences.toml` works (and can also put on the load path), and if everything else is all the same, I'd probably just describe this approach. Users familiar with Preferences.jl will know that the other option exists, while users who don't are not overwhelmed by more choices.

Another advantage of `LocalPreferences.jl` is that it is imho *much* less ugly if you have a local project in a git repo, since you can just add `LocalPreferences.jl` to `.gitignore`, while that is not possible for `Project.toml`.

Thus I suggest to reformulate this as more "do this" instead of "this is one way of doing it" :-)

docs/src/configuration.md
sloede:
```suggestion
- `JULIA_MPI_TEST_BINARY`: Check that the specified MPI binary is used for the tests
- `JULIA_MPI_TEST_ABI`: Check that the specified MPI ABI is used for the tests
```
Maybe like this?

docs/src/configuration.md
sloede: I would still remove it here then and put it next to where ABI is really used (see other comment).

docs/src/configuration.md
```diff
@@ -52,64 +48,115 @@ with:
...
+#### Advanced options
 
-### [Environment variables](@id environment_variables)
+```@doc
+MPI.use_system_binary
+```
+You can use the argument `mpiexec` to provide the the name (or full path) of the MPI launcher executable. The default is
+`mpiexec`, but some clusters require using the scheduler launcher interface (e.g. `srun`
+on Slurm, `aprun` on PBS). If the MPI library has an uncommon name you can provide it in `library_names`.
+In case ABI detection fails you can provide a manual choice,
```
sloede, line 113:
```suggestion
The MPI standard does not specify the exact application binary interface (ABI).
In case ABI detection fails you can provide a manual choice,
```

docs/src/configuration.md
```diff
@@ -52,64 +48,115 @@ with:
...
 standard or later.
 
-### Building
+### Configuration
 
-To use the the system MPI, set the environment variable `JULIA_MPI_BINARY=system` and run
-`Pkg.build("MPI")`. This can be done by:
-```
-julia --project -e 'ENV["JULIA_MPI_BINARY"]="system"; using Pkg; Pkg.build("MPI"; verbose=true)'
+To use the system MPI library, run `MPI.use_system_binary()`.
```
giordano, line 81:
```julia
julia> MPI.use_system_binary()
ERROR: UndefVarError: use_system_binary not defined
Stacktrace:
 [1] getproperty(x::Module, f::Symbol)
   @ Base ./Base.jl:35
 [2] top-level scope
   @ REPL[4]:1
```
I guess you either want to import the function from `MPIPreferences` (and maybe do `const use_system_binary = MPIPreferences.use_system_binary` to enable REPL completion), or mention here `MPIPreferences.use_system_binary`

docs/src/configuration.md
```diff
@@ -52,64 +48,115 @@ with:
...
+
+```sh
+julia --project -e 'using MPI; MPI.use_system_binary()'
 ```
-This will attempt find and identify any available MPI implementation.
+!!! note
+    You can copy `LocalPreferences.toml` to a different project, but you must list
+    `MPIPreferences` in the Project.toml either in `[deps]` or `[extras]` for the settings
+    to take effect. Due to a bug in Julia (until `v1.6.5` and `v1.7.1`), getting preferences
+    gtom transitive dependencies is broken (https://github.com/JuliaPackaging/Preferences.jl/issues/24).
```
giordano, line 92: I think "gtom" should be "from"?
```suggestion
    from transitive dependencies is broken (https://github.com/JuliaPackaging/Preferences.jl/issues/24).
```

docs/src/configuration.md
luraess: ☝️ good point.

docs/src/configuration.md
```diff
@@ -52,64 +48,115 @@ with:
...
 - [IBM Spectrum MPI](https://www.ibm.com/us-en/marketplace/spectrum-mpi)
 
-If the implementation is changed, you will need to re-run `Pkg.build("MPI")`.
+#### Advanced options
 
-### [Environment variables](@id environment_variables)
+```@doc
+MPI.use_system_binary
+```
+You can use the argument `mpiexec` to provide the the name (or full path) of the MPI launcher executable. The default is
```
luraess, line 110: Twice `the` should be removed

`to provide the the name`

docs/src/configuration.md
giordano: Uh, I missed in this PR you're importing the names from MPIPreferences, but I think it'd still be good to make REPL completion work, it's a small quality-of-life improvement

src/MPI.jl
```diff
@@ -35,6 +35,8 @@ end
 
 
 import MPIPreferences
+import MPIPreferences: use_jll_binary, use_system_binary
```
vchuravy, line 38:
```suggestion
const use_jll_binary = MPIPreferences.use_jll_binary
const use_system_binary =MPIPreferences.use_system_binary
```

@giordano like so?

docs/src/configuration.md
vchuravy, line 110:
```suggestion
You can use the argument `mpiexec` to provide the name (or full path) of the MPI launcher executable. The default is
```

src/MPI.jl
```diff
@@ -35,6 +35,8 @@ end
...
+export use_jll_binary, use_system_binary
```
vchuravy, line 39:
```suggestion
```

src/MPI.jl
giordano: Yup, perfect!

giordano (approved):
Looks good to me, thanks!  I like the section about migration from v0.19, very useful for current users.

My only problem with the new preference system is how to run tests (#561), but it's unrelated to this PR and we'll probably need to interact with Pkg devs.

PR #569, opened 2022-04-18
[docs] Fix broken links and remove references to non-existing docstrings

giordano:


docs/src/environment.md
```diff
@@ -23,5 +23,4 @@ MPI.Is_thread_main
 MPI.Initialized
 MPI.Finalize
 MPI.Finalized
-MPI.universe_size
```
giordano, line 26: This is a duplicate of https://github.com/JuliaParallel/MPI.jl/blob/5cf57d8a289a80f7e550a165863ef0dd40667370/docs/src/comm.md#miscellaneous, I feel like this function is better placed in the communication section, but I'm happy to swap it if you think it's better.

docs/src/group.md
```diff
@@ -11,29 +11,11 @@ MPI.Group
 MPI.Comparison
```
giordano, line 11: All docstrings I'm removing from this file were added in #521, but these functions/constants never had docstrings.

vchuravy (approved):
Can we also get Documenter to yell at us for missing targets?

giordano:
There are warnings, see for example https://github.com/JuliaParallel/MPI.jl/runs/6066841377?check_suite_focus=true.  In this PR I'm enabling colours to make those a bit more evident, if someone actually reads the logs.  But you probably want to make test fails with missing docstrings?  I'll see if there is this option, it'd indeed be useful, it isn't the first time I do a PR like this (#520) :smile:

PR #571, opened 2022-04-19
[MPIPreferences] Always set `ZES_ENABLE_SYSMAN=1` when dlopening libmpi

giordano:
Diff easier to read if disabling whitespace changes.

Ref: https://github.com/JuliaParallel/MPI.jl/issues/560#issuecomment-1102221341.

### DIFF
--- a/lib/MPIPreferences/src/MPIPreferences.jl
+++ b/lib/MPIPreferences/src/MPIPreferences.jl
@@ -157,96 +157,105 @@ function identify_abi(libmpi)
         ptr = dlsym(hndl, :MPI_Get_library_version)
         ccall(ptr, Cint, (Ptr{UInt8}, Ref{Cint}), buf, buflen)
     finally
-        Libdl.dlclose(hndl)
-    end
+        @assert buflen[] < 8192
+        resize!(buf, buflen[])
+        version_string = String(buf)
 
-    @assert buflen[] < 8192
-    resize!(buf, buflen[])
-    version_string = String(buf)
+        # 2) try to identify the MPI implementation
+        impl = "unknown"
+        version = v"0"
 
-    # 2) try to identify the MPI implementation
-    impl = "unknown"
-    version = v"0"
-
-    if startswith(version_string, "MPICH")
-        impl = "MPICH"
-        # "MPICH Version:\t%s\n" /  "MPICH2 Version:\t%s\n"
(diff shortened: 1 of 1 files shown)


lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -157,96 +157,105 @@ function identify_abi(libmpi)
...
-            version = VersionNumber(m.captures[1])
-        end
-    elseif startswith(version_string, "MPIwrapper")
-        impl = "MPIwrapper"
-        # MPIwrapper 2.2.2
-        if (m = match(r"^MPIwrapper Version:\t(\d+.\d+.\d+\w*)", version_string)) !== nothing
-            version = VersionNumber(m.captures[1])
+        @info "MPI implementation" libmpi version_string impl version abi
+
+        if impl == "OpenMPI" && v"4.1.2" ≤ version < v"5"
```
giordano, line 248: I don't know the exact version numbers affected.

Also, I wanted to check the value of `ZES_ENABLE_SYSMAN` here, but it appears that hwloc/openmpi change it after dlopening the library, and trying to access it _before_ dlopening the library also causes a segmentation fault, too :confused:

vchuravy: Before `dlopen` it should not cause a segmentation, after dlopen and before dlclose maybe?

giordano: I'll try to build a reproducer later today.

giordano: BTW, after dlopen and before dlcose it doesn't cause segmentation fault but `ZES_ENABLE_SYSMAN` is always set to 1, so not very useful to look at.  I tried to access it before dlopen with `get(ENV, "ZES_ENABLE_SYSMAN", "")` but I got the segmentation fault upon dlopen :confused:

simonbyrne: Yup....

simonbyrne:
I don't know what `ZES_ENABLE_SYSMAN=1` does, but should we just set it by default when calling `find_library`?

giordano:
Yeah, good call, I simplified the PR to always wrap dlopening of libmpi with `ZES_ENABLE_SYSMAN=1`.  I guess that's not going to cause any problems for a quick dlopen/dlclose cycle.

simonbyrne (approved):

PR #577, opened 2022-04-25
build UCX as well on CI

vchuravy:


.buildkite/pipeline.yml
```diff
@@ -6,39 +6,59 @@ steps:
...
+      make install
+      popd
+
+      tar -zcvf mpi-prefix.tar.gz mpi-prefix/
+    artifact_paths:
+      - "mpi-prefix.tar.gz"
+  - label: "Build OpenMPI -- ROCM"
+    agents:
+      queue: "juliagpu"
+      rocm: "*" # todo fix ROCM version
```
vchuravy, line 35: @jpsamaroo what should the value here be

.buildkite/pipeline.yml
```diff
@@ -6,39 +6,59 @@ steps:
...
       OPENMPI_VER_FULL: "4.0.3"
+      UCX_VER: "1.12.1"
     commands: |
       apt-get install --yes --no-install-recommends curl
-      curl https://download.open-mpi.org/release/open-mpi/v$${OPENMPI_VER}/openmpi-$${OPENMPI_VER_FULL}.tar.gz --output openmpi.tar.gz
+
+      curl -L https://github.com/openucx/ucx/releases/download/v$${UCX_VER}/ucx-$${UCX_VER}.tar.gz --output ucx.tar.gz
+      tar -zxvf ucx.tar.gz
+      pushd ucx-*
+      ./configure --with-cuda=/usr/share/local/cuda --prefix=$$(realpath ../mpi-prefix)
```
vchuravy, line 16: @maleadt that seems to not be the right path.

.buildkite/pipeline.yml
```diff
@@ -13,18 +13,19 @@ steps:
       curl -L https://github.com/openucx/ucx/releases/download/v$${UCX_VER}/ucx-$${UCX_VER}.tar.gz --output ucx.tar.gz
       tar -zxvf ucx.tar.gz
       pushd ucx-*
-      ./configure --with-cuda --prefix=$$(realpath ../mpi-prefix)
+      ./configure --with-cuda=/usr/share/local/cuda --prefix=$$(realpath ../mpi-prefix)
```
vchuravy, line 16:
```suggestion
      ./configure --with-cuda=/usr/local/share/cuda --prefix=$$(realpath ../mpi-prefix)
```

.buildkite/pipeline.yml
jpsamaroo: We don't have versions specified in the BK agents for `rocm`, although right now we're on 4.2.0. Because we can't easily control/select the version of ROCm in the CI containers, I would keep using `"*"`.

vchuravy: If we ever support more than one we will run into issues

PR #584, opened 2022-04-28
add SplitType

### LINKED ISSUE #579: MPI_Comm_split_type type argument
`MPI_Comm_split_type` takes a `type` argument. Currently we only expose `MPI_COMM_TYPE_SHARED`, which is kind of cumbersome as you have to do `MPI.Consts.MPI_COMM_TYPE_SHARED[]`. Additionally, some implementations (e.g. Open MPI) provide additional ones: https://www.open-mpi.org/doc/v4.1/man3/MPI_Comm_split_type.3.php

We should copy the same approach we use for other "runtime enums", where we define `mutable struct` wrapper which can be updated at load time (and the undefined ones can be set to an invalid value).

cc: @omlins

simonbyrne:
Fixes #579.

Needs a test.

cc: @omlins

### DIFF
--- a/src/comm.jl
+++ b/src/comm.jl
@@ -166,10 +166,19 @@ end
 """
     Comm_split(comm::Comm, color::Integer, key::Integer)
 
+Partition the communicator `comm`, one for each value of `color`, returning a
+new communicator. Within each group, the processes are ranked in the order of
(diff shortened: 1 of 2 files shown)


simonbyrne:
Any suggestions on how we should handle the Open MPI-specific ones?

PR #586, opened 2022-04-29
Cleanup `.gitignore`

giordano:
Do not ignore files that were previously generated by the build step, ~and the
`LocalPreferences.toml` (we shouldn't have to manually set the preference inside
this environment)~.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,4 @@
-libjuliampi.so
-libjuliampi.dylib
 *.cov
-*.dSYM
 *.mem
-*.so
 Manifest.toml
 LocalPreferences.toml


.gitignore
```diff
@@ -1,8 +1,3 @@
-libjuliampi.so
-libjuliampi.dylib
 *.cov
-*.dSYM
 *.mem
-*.so
 Manifest.toml
-LocalPreferences.toml
```
vchuravy, line 8: Why not git ignore that? It can still occur during development?

giordano: Fine, I kept it :stuck_out_tongue:

PR #591, opened 2022-05-01
Add a dummy module to force dlopening libmpi before `Consts` module

### LINKED ISSUE #587: `libmpi` already loaded before call to MPI.jl's `__init__` function

giordano:
Alternative approach to #588 and #590 to try and fix #587: without changing anything else, this PR adds a dummy module to force dlopening `libmpi` before the `Consts` submodule.  This is much simpler than #590 and, contrary to that PR, this doesn't affect loading time of the package.

CC: @sloede

### DIFF
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -57,6 +57,16 @@ else
     error("Unknown MPI binary: $(MPIPreferences.binary)")
 end
 
+if Sys.isunix()
+    # The `DlopenHack` submodule is a hack to force `dlopen`ing system libmpi on Unix
+    # systems with the flags we want to use (`Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL` instead
+    # of the default flags `RTLD_LAZY | RTLD_DEEPBIND), _before_ getting to the `Consts`
+    # module which would otherwise automatically `dlopen` libmpi with the default flags as a
+    # side effect of using `cglobal`.
+    include("dlopen_hack.jl")
+    using .DlopenHack
+end
+
 include("consts/consts.jl")
 using .Consts
 
@@ -99,18 +109,6 @@ include("mpiexec_wrapper.jl")
 include("deprecated.jl")
 
 function __init__()
-
-    @static if Sys.isunix()
-        # dlopen the MPI library before any ccall:
-        # - RTLD_GLOBAL is required for Open MPI
-        #   <https://www.open-mpi.org/community/lists/users/2010/04/12803.php>
-        # - also allows us to ccall global symbols, which enables
-        #   profilers which use LD_PRELOAD
-        # - don't use RTLD_DEEPBIND; this leads to segfaults at least
-        #   on Ubuntu 15.10
-        #   <https://github.com/JuliaParallel/MPI.jl/pull/109>
-        Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
-    end
 
     # disable UCX memory cache, since it doesn't work correctly
     # https://github.com/openucx/ucx/issues/5061
--- a/src/dlopen_hack.jl
+++ b/src/dlopen_hack.jl
@@ -1,0 +1,27 @@
+module DlopenHack
+
+using Libdl: Libdl
+using MPIPreferences: MPIPreferences
+import ..libmpi
+
+function __init__()
+    @static if MPIPreferences.binary == "system" &&
+        (x = tryparse(Bool, get(ENV, "GITHUB_ACTIONS", "false")); something(x, false))
+        # If we're running in CI, make sure that at this point the library isn't loaded yet.
+        @assert all(!contains(libmpi), Libdl.dllist())
+    end
+
+    # dlopen the MPI library before any ccall:
+    # - RTLD_GLOBAL is required for Open MPI
+    #   <https://www.open-mpi.org/community/lists/users/2010/04/12803.php>
+    # - also allows us to ccall global symbols, which enables
+    #   profilers which use LD_PRELOAD
+    # - don't use RTLD_DEEPBIND; this leads to segfaults at least
+    #   on Ubuntu 15.10
+    #   <https://github.com/JuliaParallel/MPI.jl/pull/109>
+    #   and when using HPE's MPT MPI implementation
+    #   <https://github.com/JuliaParallel/MPI.jl/pull/580>
+    Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
+end
+
+end # module


src/load.jl
```diff
@@ -0,0 +1,24 @@
+module Load
```
giordano, line 1: The name of the module is up to debate, I'm not attached to it :slightly_smiling_face:

src/load.jl
```diff
@@ -0,0 +1,24 @@
...
+
+using Libdl: Libdl
+using MPIPreferences: MPIPreferences
+import ..libmpi
+
+function __init__()
+    @static if MPIPreferences.binary == "system" && (x = tryparse(Bool, get(ENV, "CI", "false")); something(x, false))
```
giordano, line 8: Like in #590, I wanted to have a way to actually check that we open libmpi where we want.  I tried to keep this check as little invasive as possible with the `@static if`, could also replace `CI` with a more specific environment variable like `GITHUB_ACTIONS`

src/MPI.jl
```diff
@@ -57,6 +57,16 @@ else
     error("Unknown MPI binary: $(MPIPreferences.binary)")
 end
 
+if Sys.isunix()
```
sloede, line 60:
```suggestion
@static if Sys.isunix()
```
Dunno if this is even possible at module scope, but it was used in the `__init__` function

src/load.jl
sloede: I would probably call it what it is: a `UnixDlopenHack` module, or, if that is too casual, maybe `UnixForceDlopen`. Just `Load` is IMHO really too generic.

src/load.jl
sloede: I like it!

src/load.jl
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    # dlopen the MPI library before any ccall:
+    # - RTLD_GLOBAL is required for Open MPI
+    #   <https://www.open-mpi.org/community/lists/users/2010/04/12803.php>
+    # - also allows us to ccall global symbols, which enables
+    #   profilers which use LD_PRELOAD
+    # - don't use RTLD_DEEPBIND; this leads to segfaults at least
+    #   on Ubuntu 15.10
+    #   <https://github.com/JuliaParallel/MPI.jl/pull/109>
```
sloede, line 20:
```suggestion
    # - don't use RTLD_DEEPBIND; this leads to segfaults at least
    #   on Ubuntu 15.10
    #   <https://github.com/JuliaParallel/MPI.jl/pull/109>
    #   and when using HPE's MPT MPI implementation
    #   <https://github.com/JuliaParallel/MPI.jl/pull/580>
```
for completeness? Otherwise it might give the impression to future us that not using `RTLD_DEEPBIND` was a one-off issue on an ancient Ubuntu distro.

sloede (commented):
Thanks for putting in the time to find another solution to #587! I like this better than my approach in #588, since it does not change existing logic but only adds to it, with - at least to me - no significant observable downsides.

I have left a few comments (most of them minor and naming related), but overall this already LGTM. If we go ahead with this, I think we should nevertheless open an issue to track the fact that this is a hacky hack for a weird problem and that we are not proud of having the need for something like this 😬

src/MPI.jl
giordano: Top-level expressions are evaluated at precompile time anyway, a top-level `@static` doesn't do anything special.

sloede: OK, got it, thanks. Feel free to resolve this then ;-)

vchuravy:
I don't know if the initialization order of sibling modules is documented to be in-order.

sloede:
> I don't know if the initialization order of sibling modules is documented to be in-order, but we can proabably get away with this.

What if we make it a submodule to the `Consts` module? Wouldn't it guarantee to be initialized before the `Consts` module then (just like `Consts` is initialized before `MPI`)? It's a hack anyways, so I don't see a problem with adding yet another layer of submodules...

giordano:
Superseded by #592

PR #592, opened 2022-05-02
Call constant init via invokelatest

simonbyrne:
Yet another potential solution to #587. This appears to avoid early dlopen-ing of the library.

Thoughts @giordano @vchuravy @sloede ?

### DIFF
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -99,18 +99,22 @@ include("mpiexec_wrapper.jl")
 include("deprecated.jl")
 
 function __init__()
-
     @static if Sys.isunix()
         # dlopen the MPI library before any ccall:
         # - RTLD_GLOBAL is required for Open MPI
-        #   <https://www.open-mpi.org/community/lists/users/2010/04/12803.php>
-        # - also allows us to ccall global symbols, which enables
-        #   profilers which use LD_PRELOAD
-        # - don't use RTLD_DEEPBIND; this leads to segfaults at least
-        #   on Ubuntu 15.10
-        #   <https://github.com/JuliaParallel/MPI.jl/pull/109>
+        #   https://www.open-mpi.org/community/lists/users/2010/04/12803.php
+        # - also allows us to ccall global symbols, which enables profilers
+        #   which use LD_PRELOAD
+        # - don't use RTLD_DEEPBIND; this leads to issues with multiple MPI
+        #   libraries:
+        #   https://github.com/JuliaParallel/MPI.jl/pull/109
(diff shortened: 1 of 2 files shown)


sloede (approved):
LGTM! I don't know the implications of using `Base.invokelatest`, but at least on my system it seems to work as advertised in that `libmpi.so` is properly `dlopen`ed from `MPI.__init__` and not before.

There is a small performance penalty for `using MPI` at least on my system (0.209948 seconds on `master` vs. 0.254341 this PR). However, from the three proposed solutions (#588, #591, and this #592), I still favor this one since it seems like the least intrusive one.

giordano:
Any intuition why `invokelatest` delays the evaluation of `cglobal`s?

I also observe a small increase in loading time, but allocations skyrocket: `master` (from https://github.com/JuliaParallel/MPI.jl/pull/590#issue-1222218167):
```julia
julia> @time using MPI
  0.300320 seconds (193.68 k allocations: 13.254 MiB, 91.97% compilation time)
```
this PR:
```julia
julia> @time using MPI
  0.357344 seconds (706.06 k allocations: 43.360 MiB, 3.46% gc time, 91.73% compilation time)
```

sloede:
Oops, I didn't check the allocations, but an increase in 30 MiB is not trivial 🙀. This *does* make #591 look more appealing again...

simonbyrne:
I've added a `precompile` statement that reduces the latency (I see no difference now to master)

sloede:
> I've added a `precompile` statement that reduces the latency (I see no difference now to master)

`master`:
```
julia -e '@time using MPI'
  0.209852 seconds (193.57 k allocations: 13.306 MiB, 91.07% compilation time)
```

This PR:
```
julia -e '@time using MPI'
  0.212014 seconds (193.73 k allocations: 13.331 MiB, 90.34% compilation time)
```
I concur: there's virtually any difference now 🎉

@giordano If you get the same results, I propose to merge this and move on to other topics...

vchuravy:
I don't think we need the `invokelatest`? But it also doesn't hurt. This simply moves the `__init__` call from before to after the dllopen, so this seems like the best solution!

giordano:
Yup, I also see a very marginal difference now :+1: :rocket:

vchuravy (approved):

sloede:
> I don't think we need the `invokelatest`? 

When I tried this branch locally, removing invokelatest caused the library to be already opened when entering the `__init__` function, thus I do think it's necessary somehow..

PR #594, opened 2022-05-05
revamp MPIPreferences docs

### LINKED ISSUE #581: Switching system binaries fails due to undefined symbols
Trying to switch system binaries after a different system binary has already been installed will fail due to missing symbols at MPI.jl load time.

On HLRS Hawk:
* Clone this repo and dev its packages:
  ```shell
  git clone git@github.com:JuliaParallel/MPI.jl.git
  export JULIA_DEPOT_PATH=$HOME/.julia-mpijl-master
  julia -e 'using Pkg; Pkg.develop(path="MPI.jl/lib/MPIPreferences"); Pkg.develop(path="MPI.jl")'
  ```
* Load OpenMPI and install as system binary
  ```shell
  module load openmpi
  julia -e 'using MPI; MPI.use_system_binary()'
  ```
  will output something like
  ```shell
  ┌ Info: MPI implementation
  │   libmpi = "libmpi"
  │   version_string = "Open MPI v4.0.5, package: Open MPI hpcoft14@hawk-login01 Distribution, ident: 4.0.5, repo rev: v4.0.5, Aug 26, 2020\0"
  │   impl = "OpenMPI"
  │   version = v"4.0.5"
  └   abi = "OpenMPI"
  ┌ Warning: The underlying MPI implementation has changed. You will need to restart Julia for this change to take effect
  │   libmpi = "libmpi"
  │   abi = "OpenMPI"
  │   mpiexec = "mpiexec"
  └ @ MPIPreferences ~/hackathon/MPI.jl/lib/MPIPreferences/src/MPIPreferences.jl:122
  ```
* Check MPI (works): `mpirun -n 2 julia -e 'using MPI; MPI.Init(); println("I am rank $(MPI.Comm_rank(MPI.COMM_WORLD)) of $(MPI.Comm_size(MPI.COMM_WORLD)) ranks")'`
* Switch to different MPI backend
  ```shell
  module load mpt
  ```
* Try to install different MPI backend in MPI.jl
  ```shell
  julia -e 'using MPI; MPI.use_system_binary()'
  ```
  Error output:
  ```shell
  ERROR: InitError: could not load symbol "ompi_mpi_comm_null":
  /opt/hlrs/non-spack/mpi/mpt/2.23/lib/libmpi.so: undefined symbol: ompi_mpi_comm_null
  Stacktrace:
   [1] __init__()
     @ MPI.Consts ~/hackathon/MPI.jl/src/consts/consts.jl:43
   [2] _include_from_serialized(path::String, depmods::Vector{Any})
     @ Base ./loading.jl:768
   [3] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String)
     @ Base ./loading.jl:854
   [4] _require(pkg::Base.PkgId)
     @ Base ./loading.jl:1097
   [5] require(uuidkey::Base.PkgId)
     @ Base ./loading.jl:1013
   [6] require(into::Module, mod::Symbol)
     @ Base ./loading.jl:997
  during initialization of module Consts
  ```

In my naive understanding of how this newfangled MPIPreferences/Consts works, it seems like the issue is that the constants are (lazily) loaded at initialization time, thus tripping over the non-existing symbols. Any way to work around this?

simonbyrne:
Fixes #581, and adds some more details.


giordano:
For reference, the preview is at https://juliaparallel.org/MPI.jl/previews/PR594/ (I don't know if the link appears somewhere)

docs/src/configuration.md
```diff
@@ -108,31 +31,30 @@ The following MPI implementations should work out-of-the-box with MPI.jl:
...
-`LocalPreferences.toml` file to hold them.
+If the implementation is changed, you will need to call this function again. See the [`MPIPreferences.use_system_binary`](@ref) documentation for specific options.
 
-#### Notes to HPC cluster adminstators
+!!! note
+    You can copy `LocalPreferences.toml` to a different project folder, but you must list
+    `MPIPreferences` in the `[extras]` section of the `Project.toml` for the settings
+    to take effect. Due to a bug in Julia (until `v1.6.5` and `v1.7.1`), getting preferences
+    from transitive dependencies is broken (https://github.com/JuliaPackaging/Preferences.jl/issues/24).
+    To fix this update your version of Julia, or add `MPIPreferences` as a direct dependency to your project.
```
giordano, line 49: This note was here also before, but I suspect as it's phrased now it may be a bit confusing.  As far as I understand, `MPIPreferences` should be in the active environment, which *for running tests* means that it should be in `[extras]`, right?  This sounds like it should *always* be in `[extras]`.  Or am I getting it wrong?

simonbyrne: I think it also works if it is in the `[deps]` section?

simonbyrne: I've updated this, let me know what you think.

docs/src/configuration.md
```diff
@@ -108,31 +31,33 @@ The following MPI implementations should work out-of-the-box with MPI.jl:
...
-#### Notes to HPC cluster adminstators
+### Notes to HPC cluster adminstators
 
 Preferences are merged across the Julia load path, such that it is feasible to provide a module file that appends a path to
-`JULIA_LOAD_PATH` variable that contains system-wide preferences.
+`JULIA_LOAD_PATH` variable that contains system-wide preferences. The steps are as follows:
 
-As an example you can use [`MPI.use_system_binary()`](@ref MPIPreferences.use_system_binary)
-to create a file `LocalPreferences.toml` containing:
+1. Run [`MPIPreferences.use_system_binary()`](@ref MPIPreferences.use_system_binary), which will generate a file `LocalPreferences.toml` containing something like the following:
```
giordano, line 60: Note that the numbered list always restarts from 1: https://juliaparallel.org/MPI.jl/previews/PR594/configuration/#Notes-to-HPC-cluster-adminstators I believe you have to indent all blocks within each entry with two spaces, including the code blocks.

docs/src/configuration.md
giordano: Yes, I think it's better now, thanks!

docs/src/configuration.md
```diff
@@ -9,94 +9,17 @@ clusters or multi-GPU machines, you will probably want to configure against a
...
-$ mpiexecjl --project=/path/to/project -n 20 julia script.jl
-```
-
-## Using MPIPreferences.jl
-
-MPI.jl uses [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl) to
+This can be configured using the MPIPreferences.jl package, which allows the user to choose their MPI implementation. This uses [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl) to
 allow the user to choose which MPI implementation to use for a project. This provides
 a single source of truth that can be used for JLL packages (Julia packages providing C libraries)
 that link against MPI, localizes the choice of MPI implementation to a project.
```
sloede, line 15:
```suggestion
The MPIPreferences.jl package allows the user to choose which MPI implementation to use in MPI.jl. It uses [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl) to
configure the MPI backend for each project separately. This provides
a single source of truth that can be used for JLL packages (Julia packages providing C libraries)
that link against MPI.
```
I tried to simplify the paragraph somewhat and also to avoid starting 4 consecutive sentences with "this". This is just a suggestion of course; feel free to modify or discard.

docs/src/configuration.md
```diff
@@ -108,31 +31,33 @@ The following MPI implementations should work out-of-the-box with MPI.jl:
 - [Fujitsu MPI](https://www.fujitsu.com/global/about/resources/publications/technicalreview/2020-03/article07.html#cap-03)
 - [HPE MPT/HMPT](https://support.hpe.com/hpesc/public/docDisplay?docLocale=en_US&docId=a00105727en_us)
 
-If the implementation is changed, you will need to use [`MPI.use_system_binary()`](@ref MPIPreferences.use_system_binary)
-or [`MPI.use_jll_binary()`](@ref MPIPreferences.use_jll_binary).
+### Configuration
```
sloede, line 34: While not strictly necessary, I propose to make it clear that users need to have MPIPreferences.jl installed separately.

~~~suggestion
Furthermore, you need to install the MPIPreferences.jl package, e.g., by executing
```shell
julia -e 'using Pkg; Pkg.add("MPIPreferences")'
```

### Configuration
~~~

docs/src/usage.md
```diff
@@ -28,6 +28,46 @@ Hello world, I am rank 1 of 3
 
 The [`mpiexec`](@ref) function is provided for launching MPI programs from Julia itself.
 
+
```
sloede, line 31: This is the wrong line to attach this comment to, but in line 7 it says "operaions" instead of "operations"

sloede (commented):
Thanks for the improvements here and also for taking a swing at #581!

docs/src/configuration.md
simonbyrne: Thanks

docs/src/configuration.md
simonbyrne: Thanks, I added this further up.

docs/src/configuration.md
giordano: Sorry, it was three spaces, not three, I pushed directly a fix, it was too complicated to make a suggestion here, but now it renders correctly for me locally.

simonbyrne: Thank you!

giordano:
If the idea is to suggest users to use `MPIPreferences.jl` directly to choose the binary, probably should remove https://github.com/JuliaParallel/MPI.jl/blob/d607dc95d9a2aaf668ff7a0405b49b0f61d5b308/src/MPI.jl#L38-L39 as suggested by https://github.com/JuliaParallel/MPI.jl/issues/581#issuecomment-1111830529?

giordano:
https://juliaparallel.org/MPI.jl/previews/PR594/knownissues/#Hints-to-ensure-CUDA-aware-MPI-to-be-functional still references `JULIA_MPI_BINARY`

giordano (approved):
Looking good to me!

PR #572, opened 2022-04-19
Add ROCm support (AMDGPU)

luraess:
Add AMDGPU.ROCArray support to the MPI machinery to allow for ROCm-aware MPI device pointer exchange, as suggested by @simonbyrne.

**:warning: Note that this PR should replace and close #547 (without merging the latter) as suggested in https://github.com/JuliaParallel/MPI.jl/pull/547#issuecomment-1101671327** 

Requires AMDGPU v0.3.5 to work smoothly. (Thanks to @utkinis for helping out as well).

Most MPI.jl tests pass exporting `JULIA_MPI_TEST_ARRAYTYPE=ROCArray` with exception of:
- `test_reduce.jl`
- `test_subarray.jl`
- `test_threads.jl`

Failing test are currently commented out and preceded by `# DEBUG` comment.

The `dev` doc should also include references to the latest changes.

### To-dos
- to pass the tests, #564 is currently needed (issue #561) and there seems to still be an issue running the MPI tests outside of the MPI.jl project. Would require merging (#564) before this PR.
- selecting system MPI is still not super stable (#570)
- currently missing the analogy of [`CUDA.precompile_runtime()`](https://github.com/luraess/MPI.jl/blob/ef153a261ea73b48f6757b47065df8a485890880/test/runtests.jl#L10) for AMDGPU. The feature would first need to be added to AMDGPU.jl (note that all works fine without for now)
- one may want to add the CUDA-aware MPI + multi-GPU MWE to `usage.md` in the doc as well?
- add the correct definition to `.buildkite/pipeline.yml`(I am waiting to hear back from the sysadmin that built the ROCm-aware OpenMPI on the test server regarding the config)

PR #595, opened 2022-05-16
Base `MPI.identify_implementation` on helper function in `MPIPreferences`

giordano:
We basically moved the detection of ABI/implementation to `MPIPreferences`, and
added only there the support for new ABIs/implementations, which means that
`MPI.identify_implementation` is now giving incorrect results for them, and also
the constants `MPI.MPI_LIBRARY` and `MPI.MPI_LIBRARY_VERSION` are now wrong.

By having a new helper function in `MPIPreferences` we can have all other
functions use it under the hood and keep everything in-sync.

---

***Note***: this is still incomplete, because now `MPI.identify_implementation` returns a string instead of an instance of the enum `MPI.MPIImpl` (which is incomplete because it's missing the new implementations Fujitsu MPI and HPE MPT).  What we should do about that?  Make `MPI.identify_implementation` return a string or create a mapping between the strings returned by `MPIPreferences.identify_implementation_version_abi` and the enum `MPI.MPIImpl`, move `MPI.MPIImpl` to `MPIPreferences`, or simply give up on the enum and use the strings everywhere?  It doesn't look like `MPIImpl` is [widely used](https://github.com/JuliaParallel/MPI.jl/search?q=MPIImpl&type=code).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,6 +18,6 @@ Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
 
 [compat]
 DocStringExtensions = "0.8"
-MPIPreferences = "0.1"
+MPIPreferences = "0.1.2"
 Requires = "~0.5, 1.0"
 julia = "1.6"
--- a/docs/src/reference/library.md
+++ b/docs/src/reference/library.md
@@ -9,12 +9,6 @@ MPI.MPI_LIBRARY_VERSION
 MPI.MPI_LIBRARY_VERSION_STRING
 ```
 
-## Enums
-
-```@docs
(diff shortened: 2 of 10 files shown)


simonbyrne:
Lets get rid of `MPIImpl`, I don't think there is much value to making this an enum.

giordano:
Ok, this should finally be good to go.  Failures are same as https://github.com/JuliaParallel/MPI.jl/pull/572#issuecomment-1126655170

giordano:
Bump 🙂 it'd be great to cut a new release of MPIPreferences.jl after this is merged

PR #596, opened 2022-05-17
Use MPIWrapper 2.8.1 in CI

giordano:
Ref: https://github.com/JuliaParallel/MPI.jl/pull/572#issuecomment-1128869416


giordano:
Ok, I'm going to merge this PR so that we can have working CI at least for GitHub Actions everywhere else.

PR #602, opened 2022-05-27
add version errors

simonbyrne:
Extracted from #597

cc @giordano @eschnett

### DIFF
--- a/src/error.jl
+++ b/src/error.jl
@@ -2,7 +2,30 @@ struct MPIError <: Exception
     code::Cint
 end
 
-macro mpichk(expr)
+"""
+    FeatureLevelError
+
+This error is thrown if a feature is not implemented in the current MPI backend.
+"""
+struct FeatureLevelError <: Exception
+    function_name::String
+    min_version::VersionNumber # minimal MPI version required for this feature to be available
+end
+
+function Base.show(io::IO, err::FeatureLevelError)
+    print(io, "FeatureLevelError($(err.function_name)): Minimum MPI version is $(err.min_version)")
+end
+
+macro mpichk(expr, min_version=nothing)
+    if !isnothing(min_version) && expr.args[2].head == :tuple
+        fn = expr.args[2].args[1].value
+        if dlsym_e(Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL), fn) == C_NULL
+            return quote
+                throw(FeatureLevelError($(string(fn)), $min_version))
+            end
+        end
+    end
+
     expr = macroexpand(@__MODULE__, :(@mpicall($expr)))
     # MPI_SUCCESS is defined to be 0
     :((errcode = $(esc(expr))) == 0 || throw(MPIError(errcode)))
--- a/src/error.jl
+++ b/src/error.jl
@@ -1,36 +1,11 @@
+"""
+    MPIError
+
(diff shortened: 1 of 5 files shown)


src/error.jl
```diff
@@ -2,7 +2,30 @@ struct MPIError <: Exception
...
+end
+
+function Base.show(io::IO, err::FeatureLevelError)
+    print(io, "FeatureLevelError($(err.function_name)): Minimum MPI version is $(err.min_version)")
+end
+
+macro mpichk(expr, min_version=nothing)
+    if !isnothing(min_version) && expr.args[2].head == :tuple
+        fn = expr.args[2].args[1].value
+        if dlsym_e(Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL), fn) == C_NULL
```
giordano, line 22: Note that [`dlsym_e(...)`](https://docs.julialang.org/en/v1/stdlib/Libdl/#Base.Libc.Libdl.dlsym_e) is deprecated in favour of [`dlsym(...; throw_error=false)`](https://docs.julialang.org/en/v1/stdlib/Libdl/#Base.Libc.Libdl.dlsym)

src/error.jl
```diff
@@ -2,7 +2,30 @@ struct MPIError <: Exception
...
+    min_version::VersionNumber # minimal MPI version required for this feature to be available
+end
+
+function Base.show(io::IO, err::FeatureLevelError)
+    print(io, "FeatureLevelError($(err.function_name)): Minimum MPI version is $(err.min_version)")
+end
+
+macro mpichk(expr, min_version=nothing)
+    if !isnothing(min_version) && expr.args[2].head == :tuple
+        fn = expr.args[2].args[1].value
```
giordano, line 21: To be clear, I think this body is executed only the first time you call the macro, right?

simonbyrne: It's executed when the macro expands

src/error.jl
```diff
@@ -1,14 +1,43 @@
...
+    min_version::VersionNumber # minimal MPI version required for this feature to be available
+end
+function Base.show(io::IO, err::FeatureLevelError)
+    print(io, "FeatureLevelError($(err.function_name)): Minimum MPI version is $(err.min_version)")
+end
+
+macro mpichk(expr, min_version=nothing)
+    if !isnothing(min_version) && expr.args[2].head == :tuple
+        fn = expr.args[2].args[1].value
+        if dlsym(Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL), fn; throw_error=false) == C_NULL
```
giordano, line 29:
```suggestion
        if isnothing(dlsym(Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL), fn; throw_error=false))
```

giordano: Maybe we should keep the libmpi handle around, instead of running `dlopen` every time?  We already have it for JLL packages.

giordano:
After this PR I'm getting
```
julia> using MPI
[ Info: Precompiling MPI [da04e1cc-30fd-572f-bb4f-1f8673147195]
WARNING: could not import System.libmpi_handle into MPI
```

simonbyrne:
You need to use the master MPITrampoline

simonbyrne:
I mean MPIPreferences

giordano:
I guess this needs to be reflected in the compat section then :slightly_smiling_face:

Issue #613, opened 2022-06-20
cray-mpich toolchain not recognized by MPIPreferences


williamfgc:
@luraess and I have been testing MPI.jl on OLCF's Crusher which uses cray-mpich. 
Runs execute as independent MPI processes, we think the issue is that MPIPreferences can't recognize the cray-mpich system MPI. To reproduce: 

```
module load rocm cray-mpich

julia> using MPIPreferences

julia> MPIPreferences.use_system_binary()
ERROR: MPI library could not be found
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] use_system_binary(; library_names::Vector{String}, mpiexec::String, abi::Nothing, export_prefs::Bool, force::Bool)
   @ MPIPreferences ~/.julia/packages/MPIPreferences/MXVwb/src/MPIPreferences.jl:130
 [3] use_system_binary()
   @ MPIPreferences ~/.julia/packages/MPIPreferences/MXVwb/src/MPIPreferences.jl:117
 [4] top-level scope
   @ REPL[2]:1
```

cray-mpich doesn't have a `libmpi.so` but `libmpi_cray.so` and uses `srun` as the executable:

```
       libmpi_cray.so.12 (libc6,x86-64) => /opt/cray/pe/lib64/libmpi_cray.so.12
	libmpi_cray.so (libc6,x86-64) => /opt/cray/pe/lib64/libmpi_cray.so
	libmpi_amd.so.12 (libc6,x86-64) => /opt/cray/pe/lib64/libmpi_amd.so.12
	libmpi_amd.so (libc6,x86-64) => /opt/cray/pe/lib64/libmpi_amd.so
```

Just looking for guidance or a potential feature support. We'll be happy to help make this functional and test on Crusher.

giordano:
You can use `MPIPreferences.use_system_binary(; library_names=["libmpi_cray.so"], mpiexec="srun")`: https://juliaparallel.org/MPI.jl/dev/reference/mpipreferences/#MPIPreferences.use_system_binary

williamfgc:
> You can use

MPIPreferences.use_system_binary(; library_names=["libmpi_cray"], mpiexec="srun")

@giordano thanks, this made `hello world` work and recognize the 8 GPUs on one node. There are errors in the miniapp, complaining about requesting more devices and a seg fault on a different test. We are currently debugging, but this helped already to move forward. We'll keep you posted.

luraess:
[...]

PR #614, opened 2022-06-20
[MPIPreferences] Add `libmpi_cray` to default names and improve error message

giordano:
Ref: #613.  Suggestions about what we can do with `mpiexec` welcome.  One thing I just thought is that we could do something like
```julia
mpiexec = ["mpiexec", "mpirun", "srun", "arun"]
first(filter(x -> !isnothing(Sys.which(x)), mpiexec))
```

### DIFF
--- a/lib/MPIPreferences/src/MPIPreferences.jl
+++ b/lib/MPIPreferences/src/MPIPreferences.jl
@@ -86,7 +86,7 @@ end
 
 """
     MPIPreferences.use_system_binary(;
-        library_names = ["libmpi", "libmpi_ibm", "msmpi", "libmpich", "libmpitrampoline"],
+        library_names = ["libmpi", "libmpi_ibm", "msmpi", "libmpich", "libmpi_cray", "libmpitrampoline"],
         mpiexec = "mpiexec",
         abi = nothing,
         export_prefs = false,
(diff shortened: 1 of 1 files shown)


giordano:
@williamfgc @luraess what do you get on Crusher with
```julia
julia> mpiexec = ["mpiexec", "mpirun", "srun", "arun"]
4-element Vector{String}:
 "mpiexec"
 "mpirun"
 "srun"
 "arun"

julia> filter(x -> !isnothing(Sys.which(x)), mpiexec)
2-element Vector{String}:
 "mpiexec"
 "mpirun"
```
?

sloede:
> Suggestions about what we can do with `mpiexec` welcome.

I like the idea of helping the user here to make the right choice, but shouldn't the list be ordered the other way around then? That is, if there is, e.g., an `srun` command available, it is imho very likely that this is the command that should be used and not plain `mpiexec`. OTOH, this might be confusing for non-HPC users 🤷‍♂️

Either way, I think `aprun` should also be on the list of possible executables.

giordano:
> Either way, I think aprun should also be on the list of possible executables.

That was a typo: `arun` above should have been `aprun`.

simonbyrne:
Cluster launchers are so finicky that I think we should try to do as little magic as possible. For example, on our Slurm cluster, MPICH jobs have to be launched using `srun`, but Open MPI via `mpiexec`.

simonbyrne:
> Do you have different modules for them? Are the different executables all in PATH the same time?

Yes, there are different modules: MPICH doesn't have any executables (it uses `srun` which is always in the PATH), but Open MPI does (it adds its own `mpiexec` to PATH)

giordano:
Ugh, yeah, that's annoying.  Anyway this PR doesn't touch the launcher, so could go on if it looks ok.

williamfgc:
@giordano I get `srun` only. Thanks everyone for chiming in

```
julia> filter(x -> !isnothing(Sys.which(x)), mpiexec)
1-element Vector{String}:
 "srun"
```

PR #626, opened 2022-09-11
document how to pass MPI.Comm objects to C

stevengj:
This seems to be missing from the docs ([discourse](https://discourse.julialang.org/t/how-can-i-pass-the-pointer-of-mpi-communicator-to-c-function/87070)).

### DIFF
--- a/docs/src/reference/comm.md
+++ b/docs/src/reference/comm.md
@@ -12,7 +12,7 @@ each process a unique *rank* (see [`MPI.Comm_rank`](@ref)) taking an integer val
 MPI.Comm
 ```
 
-If you need to pass a Julia `MPI.Comm` object to an external C/C++ library ([via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/)) that expects an `MPI_Comm` argument, you should declare the corresponding `ccall` argument as a `Ptr{MPI.MPI_Comm}` and the correct conversion will be performed.
+If you need to pass a Julia `MPI.Comm` object to an external C/C++ library ([via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/)) that expects an `MPI_Comm` argument, you should declare the corresponding `ccall` argument as a `MPI.MPI_Comm` and the correct conversion will be performed.
 
 ## Constants
 
--- a/docs/src/reference/comm.md
+++ b/docs/src/reference/comm.md
@@ -12,6 +12,8 @@ each process a unique *rank* (see [`MPI.Comm_rank`](@ref)) taking an integer val
 MPI.Comm
 ```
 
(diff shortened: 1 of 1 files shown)


stevengj:
CI failure `LoadError: UndefVarError: get_default_queue not defined` seems unrelated.

giordano:
Yes.  For the record, that's in AMDGPU tests CC @luraess

docs/src/reference/comm.md
```diff
@@ -12,6 +12,8 @@ each process a unique *rank* (see [`MPI.Comm_rank`](@ref)) taking an integer val
 MPI.Comm
 ```
 
+If you need to pass a Julia `MPI.Comm` object to an external C/C++ library ([via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/)) that expects an `MPI_Comm` argument, you should declare the corresponding `ccall` argument as a `Ptr{MPI.MPI_Comm}` and the correct conversion will be performed.
```
simonbyrne, line 15: `Ptr` is used if it's a pointer to the handle, not the handle itself.

```suggestion
If you need to pass a Julia `MPI.Comm` object to an external C/C++ library ([via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/)) that expects an `MPI_Comm` argument, you should declare the corresponding `ccall` argument as a `MPI.MPI_Comm` and the correct conversion will be performed.
```

luraess:
> Yes. For the record, that's in AMDGPU tests CC @luraess

Yes, there is an issue related to renaming of `AMDGPU.get_default_queue()`. I'll fix it.

stevengj:
Committed suggestion from @simonbyrne.

docs/src/reference/comm.md
```diff
@@ -12,6 +12,8 @@ each process a unique *rank* (see [`MPI.Comm_rank`](@ref)) taking an integer val
 MPI.Comm
 ```
 
+If you need to pass a Julia `MPI.Comm` object to an external C/C++ library ([via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/)) that expects an `MPI_Comm` argument, you should declare the corresponding `ccall` argument as a `MPI.MPI_Comm` and the correct conversion will be performed.
```
simonbyrne, line 15:
```suggestion
If you need to pass a Julia `MPI.Comm` object to an external C/C++ library ([via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/)) that expects an `MPI_Comm` argument, you should declare the corresponding `ccall` argument type `MPI.API.MPI_Comm` and the correct conversion will be performed.
```

simonbyrne:
Thanks!

PR #621, opened 2022-08-24
Add CUDA-aware MPI all-to-all tests and fix typo

luraess:
- Add all-to-all CUDA-aware scripts analogous by analogy with ROCm-aware
- Remove white space in OpenMPI

PR #628, opened 2022-09-13
Fix AMDGPU `synchronize` in tests & update doc

luraess:
Update ROC queue initialisation syntax to latest to fix AMDGPU related tests. Follow-up from https://github.com/JuliaParallel/MPI.jl/pull/626#issuecomment-1244468308

Also includes doc updates previously contained in #621


luraess:
it seems that the GPU CI failure is due to the ROCr version used on the CI machine, according to @jpsamaroo (and is not something directly related to the `synchronize` function implementation).

This PR thus addresses following:
- update the doc regarding CUDA-aware MPI on multi-GPUs adding a MWE similar to what exists for ROCm-aware MPI 
- update the queue initialisation in the `synchronize` function to be up to date with latest AMDGPU (v0.4.2) release

Could @simonbyrne or @vchuravy have look at this and potentially merge it?

simonbyrne:
Thanks!

PR #629, opened 2022-09-15
Add `TOML` to test dependencies

giordano:
According to @vchuravy this should fix the failure in CI with Julia v1.8.


giordano:
It worked :partying_face:

PR #630, opened 2022-09-16
Wrap `MPI_Get_processor_name` function

### LINKED ISSUE #624: Wrap `MPI_Get_processor_name`
Is there any interest in adding a wrapper for `MPI_Get_processor_name`?  Today I needed to use it.  Something like
```julia
function Get_processor_name()
    proc_name = zeros(Cchar, MPI.Consts.MPI_MAX_PROCESSOR_NAME)
    name_len = Ref{Cint}(0)
    @ccall MPI.libmpi.MPI_Get_processor_name(proc_name::Ptr{Cchar}, name_len::Ptr{Cint})::Cint
    proc_name[end] = 0 # ensure null-termination
    GC.@preserve proc_name unsafe_string(pointer(proc_name))
end
```
should do it, I opened the issue to check if it's worth adding this function or not.

giordano:
I spent most of the time deciding where to put the test and the docstring.  I'm not sure I found sensible places, suggestions are welcome.

Fix #624.

### DIFF
--- a/test/test_test.jl
+++ b/test/test_test.jl
@@ -2,6 +2,8 @@ include("common.jl")
 
 MPI.Init()
 
+@test MPI.Get_processor_name() == gethostname()
+
 comm = MPI.COMM_WORLD
 size = MPI.Comm_size(comm)
 rank = MPI.Comm_rank(comm)
--- a/docs/src/reference/environment.md
+++ b/docs/src/reference/environment.md
@@ -18,6 +18,7 @@ MPI.ThreadLevel
 ```@docs
 MPI.Abort
 MPI.Init
+MPI.Get_processor_name
 MPI.Query_thread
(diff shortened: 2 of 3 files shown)


test/test_test.jl
```diff
@@ -2,6 +2,8 @@ include("common.jl")
 
 MPI.Init()
 
+@test MPI.Get_processor_name() == gethostname()
```
giordano, line 5: Note that documentation of [`MPI_Get_processor_name`](https://www.open-mpi.org/doc/v3.1/man3/MPI_Get_library_version.3.php) specifically says the function doesn't necessarily return the same as `gethostname`, but honestly I wouldn't know what to test otherwise, apart from `MPI.Get_processor_name() isa String`?

PR #632, opened 2022-09-17
Add wrapper for MPI_Get_processor_name

plessl:
This PR add a wrapper for MPI_Get_processor_name

The MPI_Get_processor_name function can be handy for testing MPI programs but was not yet available in MPI.jl. Also, some introductory MPI examples for C and Fortran rely on this function, hence the lack of a wrapper in MPI.jl prevents porting these examples to Julia.

Since the result of this function is by definition dependent on the machine that executes the code, I wasn't able to come up with a meaningful test.

### DIFF
--- a/src/environment.jl
+++ b/src/environment.jl
@@ -307,3 +307,21 @@ function has_cuda()
         return parse(Bool, flag)
     end
 end
+
+"""
+    Get_processor_name()
+
+Returns the name of the processor
+
+# External links
+$(_doc_external("MPI_Get_processor_name"))
+"""
+function Get_processor_name()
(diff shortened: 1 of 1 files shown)


giordano:
Duplicate of #630 :slightly_smiling_face:

plessl:
I agree, as it is a duplicate of #630

plessl:
What a coincidence! The lack of this function was bothering me for weeks and I finally got around to add it tonight. 

Regarding the test: I found this tricky as well. I think I have seen HPC clusters that return a string that is not equivalent to the hostname alone but a combination of hostname and core. But I need to doublecheck this.

PR #634, opened 2022-09-18
Move `Get_processor_name` to new Misc section, and fix test

giordano:
Following up from https://github.com/JuliaParallel/MPI.jl/pull/632#issuecomment-1250138261, @plessl found a system where `MPI.Get_processor_name()` doesn't indeed match `gethostname()`.  I guess we don't want tests to fail for this trivial issue, so I simplified the test to just check that the function returns a string, as suggested in https://github.com/JuliaParallel/MPI.jl/pull/630#discussion_r973322003.  Not super informative, but at least we still make sure the function runs and doesn't crash.

Also, following [mpi4py docs](https://mpi4py.readthedocs.io/en/stable/mpi4py.MPI.html#functions), I moved `MPI.Get_processor_name` to a new "Misc" section (also in source code, and tests).  I think that better classifies this function, and it was also my main doubt in https://github.com/JuliaParallel/MPI.jl/pull/630#issue-1376374138.  We may add more misc functions in the future.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -63,6 +63,7 @@ makedocs(
             "reference/mpipreferences.md",
             "reference/library.md",
             "reference/environment.md",
+            "reference/misc.md",
             "reference/comm.md",
             "reference/buffers.md",
             "reference/pointtopoint.md",
--- a/docs/src/reference/environment.md
+++ b/docs/src/reference/environment.md
@@ -18,7 +18,6 @@ MPI.ThreadLevel
 ```@docs
 MPI.Abort
 MPI.Init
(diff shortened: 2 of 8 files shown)


plessl:
I tried to reproduce the mismatch of the results of MPI.Get_processor_name and gethostname(). But I was unable to reproduce the issue.

But I also suggest to simply skip this test because the standard leaves the format of the result open to the implementation (see Section 9.1.2 in MPI 4.0 standard https://www.mpi-forum.org/docs/mpi-4.0/mpi40-report.pdf) and only defines that the string must be suitable to uniquely identify the processor in the system.

PR #636, opened 2022-09-21
no need for mutable struct with container fields

t-bltg:
There is no need for the `struct` to be mutable when the `struct` field are containers.


giordano (approved):
Looks good to me.  However this would be made redundant by #645, but not sure where that PR is heading to.

simonbyrne:
#645 is on hold

PR #637, opened 2022-09-21
[WIP] Use `FunctionWrappers` for custom reduction operators

giordano:
My attempt at addressing #404.  This is using [`FunctionWrappers.jl`](https://github.com/yuyichao/FunctionWrappers.jl) instead of `@cfunction`, as suggested in a recent JuliaHPC call.

Note: it doesn't work for me even on x86_64 when using more than 1 rank.  Simple reproducer:
```julia
using MPI
MPI.Init()
let
    T = Int
    dims = 1
    op = (x,y) -> 2x+y-x
    send_arr = Array(zeros(T, Tuple(3 for i in 1:dims)))
    send_arr[:] .= 1:length(send_arr)
    recv_arr = Array{T}(undef, size(send_arr))
    MPI.Reduce!(send_arr, recv_arr, op, MPI.COMM_WORLD; root=0)
end
```
Run the script with
```
mpiexecjl -np 2 julia /path/to/script
```
The program should die with a segmentation fault (but the generated pointer should be non-null, as far as I can tell)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,7 @@ version = "0.20-dev"
 [deps]
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+FunctionWrappers = "069b7b12-0de2-55c6-9aab-29f3d0a68a2e"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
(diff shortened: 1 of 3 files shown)


giordano:
I thought that there was a segfault in the call to `MPI_Reduce` at https://github.com/JuliaParallel/MPI.jl/blob/4cd71180218a5bb2e88a7c153ef8538dbb7ea74e/src/collective.jl#L598-L600 but putting some debug printing shows that the ccall should work fine.  I'm wondering if something is something should be `GC.@preserve`'d.

PR #638, opened 2022-09-22
lower level forms

t-bltg:
In order to ease one-to-one mapping of Fortran codes, I would like to be able to access lower forms of `Irecv!`, `Isend!`, ... for point to point communications.

This is a nfc regarding tests.


src/pointtopoint.jl
```diff
@@ -204,13 +212,19 @@ $(_doc_external("MPI_Irecv"))
 """
 Irecv!(recvbuf, comm::Comm; source::Integer=Consts.MPI_ANY_SOURCE[], tag::Integer=Consts.MPI_ANY_TAG[]) =
     Irecv!(recvbuf, source, tag, comm)
-function Irecv!(buf::Buffer, source::Integer, tag::Integer, comm::Comm)
-    req = Request()
+function Irecv!(data::Union{MPIPtr, MPIBuffertype}, count::Integer, datatype::Union{MPI_Datatype, Datatype}, source::Integer, tag::Integer, comm::Union{MPI_Comm, Comm}, req::Union{MPI_Request, Request})
```
simonbyrne, line 215: It doesn't make sense to include `MPI_Request` here in the signature, as it is an immutable object so can't be modified (you could use `Ptr{MPI_Request}`?)

t-bltg:
Closing in favor of MPI_Type_vector.

PR #641, opened 2022-09-27
add warning if JULIA_MPI_BINARY is inconsistent with MPIPreferences

### LINKED ISSUE #640: Add warning if JULIA_MPI_BINARY is set, and differs from MPIPreferences.binary
In order to make it easier to upgrade, it was suggested to provide a mechanism some mechanism that make use of previous configurations.

Some potential options:
1. At `MPI.__init__()`, check if `JULIA_MPI_BINARY` is set and if it is different to `MPIPreferences.binary`: if so, print a warning
2. In `Pkg.build()`, if `JULIA_MPI_BINARY` is set and different from `MPIPreferences.binary`, then call the appropriate `MPIPreferences` function, copying the preferences to `~/.julia/environments/v1.X`
    - it was decided on the JuliaHPC call that this was too invasive

cc: @sloede @vchuravy

simonbyrne:
Fixes #640

### DIFF
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -97,6 +97,23 @@ include("mpiexec_wrapper.jl")
 include("deprecated.jl")
 
 function __init__()
+
+    # an empty string was used to indicate "default"
+    # https://github.com/JuliaParallel/MPI.jl/blob/v0.19.2/deps/build.jl#L142
+    mpi_env_binary = get(ENV, "JULIA_MPI_BINARY", "")
+    if mpi_env_binary != "" && mpi_env_binary != MPIPreferences.binary
+        @info """
+        The JULIA_MPI_BINARY environment variable is no longer used to configure the MPI binary.
+        Please use the MPIPreferences.jl package instead:
+
+            MPIPreferences.use_system_binary()  # use the system binary
+            MPIPreferences.use_jll_binary()     # use JLL binary
+
+        See https://juliaparallel.org/MPI.jl/stable/configuration/ for more details
+
+        """ ENV["JULIA_MPI_BINARY"]=mpi_env_binary MPIPreferences.binary
+    end
+
     @static if Sys.isunix()
         # dlopen the MPI library before any ccall:
         # - RTLD_GLOBAL is required for Open MPI


src/MPI.jl
```diff
@@ -97,6 +97,23 @@ include("mpiexec_wrapper.jl")
...
+        @info """
+        The JULIA_MPI_BINARY environment variable is no longer used to configure the MPI binary.
+        Please use the MPIPreferences.jl package instead:
+
+            MPIPreferences.use_system_binary()  # use the system binary
+            MPIPreferences.use_jll_binary()     # use JLL binary
+
+        See https://juliaparallel.org/MPI.jl/stable/configuration/ for more details
+
+        """ ENV["JULIA_MPI_BINARY"]=mpi_env_binary MPIPreferences.binary
```
sloede, line 114: What's the motivation for resetting the environment variable here? For MPI.jl, it should have no effect anymore, so why bother at all?

src/MPI.jl
```diff
@@ -97,6 +97,23 @@ include("mpiexec_wrapper.jl")
...
 
 function __init__()
+
+    # an empty string was used to indicate "default"
+    # https://github.com/JuliaParallel/MPI.jl/blob/v0.19.2/deps/build.jl#L142
+    mpi_env_binary = get(ENV, "JULIA_MPI_BINARY", "")
+    if mpi_env_binary != "" && mpi_env_binary != MPIPreferences.binary
+        @info """
+        The JULIA_MPI_BINARY environment variable is no longer used to configure the MPI binary.
+        Please use the MPIPreferences.jl package instead:
```
sloede, line 107: What about playing back to the user what lead to the warning such that even unsuspecting users (e.g., on module-managed systems with automatic environment variables) know what to set their MPIPreferences to? For example, something like this:
```suggestion
        The JULIA_MPI_BINARY environment variable is no longer used to configure the MPI binary.
        This results in a mismatch between your environment variable setting ("$mpi_env_binary") and
        what MPI.jl actually uses ("$MPIPreferences.binary").
        
        Please use the MPIPreferences.jl package for configuration instead:
```

sloede (commented):
Thanks for getting started with a solution for this! I haven't tested it (have you?), but overall it looks good to me. And ❤️ for using the https://juliaparallel.org URL to refer to the MPI.jl docs 😊

src/MPI.jl
simonbyrne: This doesn't reset it, it is specifying a key to the logging macro (if you specify an expression as a key, it is converted to string). It prints as:
```
┌ Info: The JULIA_MPI_BINARY environment variable is no longer used to configure the MPI binary.
│ Please use the MPIPreferences.jl package instead:
│ 
│     MPIPreferences.use_system_binary()        # use the system binary
│     MPIPreferences.use_jll_binary()     # use JLL binary
│ 
│ See https://juliaparallel.org/MPI.jl/stable/configuration/ for more details
│ 
│   ENV["JULIA_MPI_BINARY"] = "system"
└   MPIPreferences.binary = "MPICH_jll"
```

src/MPI.jl
simonbyrne: that is actually what the keys below do: see my comment above.

src/MPI.jl
sloede: Oh, great, that is very nice!

src/MPI.jl
sloede: Great, obviously you were at least one step ahead of me 👍

sloede (approved):
LGTM!

vchuravy (approved):

PR #635, opened 2022-09-21
add `create_hvector` (support `MPI_Type_create_hvector`)

t-bltg:
https://www.mpich.org/static/docs/v3.3/www3/MPI_Type_create_hvector.html.

https://github.com/eschnett/MPItrampoline/blob/1b759d1f4255b507d2dbbd41a822a27558f5ba2d/mpiabi/mpi_functions.py#L355.

PR #644, opened 2022-09-28
Generate `MPI` `API` bindings using `Clang.jl`

t-bltg:
Automatically generate `MPI` bindings from `MPICH` using `Clang.jl`.

A few functions were skipped because of missing types, but they can easily be added by adding new types in the `Consts` module and re-running the generator.

Proof of concept after https://github.com/JuliaParallel/MPI.jl/pull/639#issuecomment-1261213644.

Needs https://github.com/JuliaInterop/Clang.jl/pull/385 (EDIT: released in `0.16.2`).

Supersedes https://github.com/JuliaParallel/MPI.jl/pull/635, https://github.com/JuliaParallel/MPI.jl/pull/638, https://github.com/JuliaParallel/MPI.jl/pull/639 (I want to have low level access to MPI routines, mostly to avoid allocations).


simonbyrne:
Thanks.
- In general, you shouldn't need to define the constants and types (these are already defined in https://github.com/JuliaParallel/MPI.jl/tree/master/src/consts)
- You can filter out to only include functions which start with `MPI_`
- these signatures should then be the same for all implementations.

t-bltg:
> In general, you shouldn't need to define the constants and types (these are already defined in https://github.com/JuliaParallel/MPI.jl/tree/master/src/consts)

Yes, but I need the types for the signatures !

> You can filter out to only include functions which start with MPI_

Done, I've configured the generator to export only filtered functions.

> these signatures should then be the same for all implementations.

`using MPI` now works without throwing.

src/MPI.jl
```diff
@@ -58,6 +58,23 @@ end
...
+    names_Consts = filter(n -> startswith(string(n), "MPI_"), names(Consts; all = true)) |> collect
+    for n in names_Consts
+        @eval $n = Consts.$n  # signatures need types and constants
+    end
+
+    include("../gen/out/api.jl")
+    for name in names(@__MODULE__; all=true)
+        name in names_Consts && continue  # only export symbols from api
+        startswith(string(name), "MPI_") && @eval export $name
+    end
```
simonbyrne, line 74: Do we need to export them?

src/MPI.jl
```diff
@@ -58,6 +58,23 @@ end
...
 using .Consts
 
+module API
+    using ..MPIPreferences
+    using ..Consts
+
+    names_Consts = filter(n -> startswith(string(n), "MPI_"), names(Consts; all = true)) |> collect
+    for n in names_Consts
+        @eval $n = Consts.$n  # signatures need types and constants
+    end
```
simonbyrne, line 68: what if we just moved all of `Consts` into `API` (and then just make `const Consts = API` for compatibility)

src/MPI.jl
t-bltg: No indeed, we said they should stay in a dedicated module, will fix.

src/MPI.jl
t-bltg: If you wish.

simonbyrne:
Wow, that worked way better than I had expected!

src/MPI.jl
t-bltg: Mmmh it's less easier than thought.

t-bltg:
> Wow, that worked way better than I had expected!

We sometimes forget that `julia` is amazing !
`Clang.jl` is also easy to use: first time and no really difficulty.

The tests are passing.

@simonbyrne, I only used the new API for `collective.jl/Barrier`: before I go on modifying everything, I would like to know if this is what we decide.

Maybe other reviewers can have a look and comment ?

gen/Project.toml
```diff
@@ -0,0 +1,6 @@
+name = "MPIgen"
+uuid = "01aa870b-fa19-4f9c-adda-70d3c70d3e08"
+
+[deps]
+Clang = "40e3b903-d033-50b4-a0cc-940c62c95e31"
```
giordano, line 5: Can you please add the compat for Clang?

gen/src/MPIgen.jl
```diff
@@ -0,0 +1,42 @@
+module MPIgen
+    using Clang.Generators
+    using MPICH_jll
+
+    generate_signatures(dir = MPICH_jll.artifact_dir) = begin
+        @info "Generate MPI bindings for MPICH in $dir"
+
+        out = joinpath(@__DIR__, "..", "out")
```
giordano, line 8: I wonder if this should be saved into the package src directory instead

gen/Project.toml
```diff
@@ -0,0 +1,6 @@
...
+MPICH_jll = "7cb0a576-ebde-5e09-9194-50597f1243b4"
```
giordano, line 6: Would also be useful to add MPIPreferences (with compat) and the local preference file, to ensure this works correctly?

src/MPI.jl
```diff
@@ -58,6 +58,23 @@ end
...
+    import ..libmpi
+
+    for name in filter(n -> startswith(string(n), "MPI_"), names(Consts; all = true))
+        @eval $name = Consts.$name  # signatures need types
+    end
+
+    struct MPIError <: Exception
+        code::Cint
+    end
+    macro mpichk(expr)
```
giordano, line 72: I guess `@mpichk` in the top-module should be a const alias of this one?

gen/out/api.jl
```diff
@@ -0,0 +1,2227 @@
...
+    @mpichk ccall((:MPI_Unpublish_name, libmpi), Cint, (Ptr{Cchar}, MPI_Info, Ptr{Cchar}), service_name, info, port_name)
+end
+
+function MPI_Wtick()
+    @mpichk ccall((:MPI_Wtick, libmpi), Cdouble, ())
+end
+
+function MPI_Wtime()
+    @mpichk ccall((:MPI_Wtime, libmpi), Cdouble, ())
+end
```
simonbyrne, line 1299: These two return doubles, not error codes, so don't need the `@mpichk`

simonbyrne: We have a macro `@mpicall` which calls the global symbols on unix platforms (to allow for MPI profilers)
https://github.com/JuliaParallel/MPI.jl/blob/f8e4749c1ce8141458d0a88664044ec64da357ba/src/implementations.jl#L8-L27

src/MPI.jl
simonbyrne: We shouldn't need it in the top-level now, just move it in here

t-bltg: Yes, solved.

gen/Project.toml
t-bltg: Yes, added.

gen/Project.toml
t-bltg: Yes, but please note that this will that we need https://github.com/JuliaInterop/Clang.jl/pull/385, so the current compat entry for `Clang` is a lie.

gen/src/MPIgen.jl
t-bltg: @simonbyrne, what is your preferred directory structure ?

gen/out/api.jl
t-bltg: I've reworked the `MPIgenerator` module to add either the `@mpichk` or the `@mpicall` macro prefix.

gen/src/MPIgen.jl
t-bltg: I've decided to move the auto-generated file to `src/auto_generated_api.jl` and added a warning at the beginning of the file advising against editing it.

t-bltg:
I've replaced the majority of `@mpichk ccall(...)` with calls to `API.MPI_XXX` methods.

I couldn't replace the calls using the `min_version` argument of the `@mpichk` macro (in `collective.jl` and `topology.jl`), because of the structural changes.
What can be do about it ?

gen/Project.toml
t-bltg: Fixed by https://github.com/JuliaRegistries/General/pull/69201.

t-bltg:
I've started to test the new API in my codes, and the low level allocations free version passes all my tests.

I can now write things very close to the `Fortran` version such as:
```julia
# somewhere, in my constructor ...
s_req = resize!(MPI.MPI_Request[], req_max)
r_req = resize!(MPI.MPI_Request[], req_max)
s_stat = resize!(MPI.MPI_Status[], req_max)
r_stat = resize!(MPI.MPI_Status[], req_max)

# somewhere, in a synchronization routine
# [...]
MPI.API.MPI_Irecv(Ref(mem, I), 1, a.rtype[n], ..., pointer(a.topo.r_req, r_irq))
# [...]
MPI.API.MPI_Isend(Ref(mem, I), 1, a.stype[n], ..., pointer(a.topo.s_req, s_irq))
# [...]
MPI.API.MPI_Waitall(s_irq, pointer(a.topo.s_req), pointer(a.topo.s_stat))
MPI.API.MPI_Waitall(r_irq, pointer(a.topo.r_req), pointer(a.topo.r_stat))
```

t-bltg:
The signatures generator can now work on either `MPICH_jll` or `OpenMPI_jll`.

simonbyrne:
> The signatures generator can now work on either `MPICH_jll` or `OpenMPI_jll`.

Do they generate the same code?

t-bltg:
Similar (all `MPI` tests pass for both on `linux`) but the lines are reordered, so I cannot produce a significant diff.

Here are the number of lines however:
```bash
$ wc -l
`MPICH_jll` 2229
`OpenMPI_jll` 1541
```

It is dependent of the include flags in the generator, but parsing `mpi_proto.h` for `MPICH` and `mpi.h` for `OpenMPI` seems sufficient for `MPI.jl` at this state.

simonbyrne:
> I couldn't replace the calls using the `min_version` argument of the `@mpichk` macro (in `collective.jl` and `topology.jl`), because of the structural changes. What can be do about this ?

In the generator, would it be possible to have say a list of `pattern => min_version` pairs, where we then check the function name against `pattern` (and leaves it off if no match)? (don't worry if the list isn't complete, we can fill this in as we go).

t-bltg:
Done, `MPI.jl/src/auto_generated_api.jl` now has `min_version` entries.

I've finalized the `@mpichk` API replacements for `collective.jl` and `topology.jl`.

simonbyrne (approved):
Overall, this looks great, and will make it much, much easier to add more functionality (or allow users to work around missing functionality). Thanks!

Let me know once you're done.

PR #645, opened 2022-09-28
make Request objects immutable

### LINKED ISSUE #643: Request objects and Waitall! allocations
Our `MPI.Request` objects are currently mutable structs containing an `MPI_Request` handle. The problem is that the non-blocking group test/wait functions (such as `MPI_Waitall`) require a vector of `MPI_Request` handles: if we give them a vector of `MPI.Request` objects, we need to convert this somehow.

On the current release version (v0.19.2), `MPI.Waitall!` just allocates a vector of `MPI_Request` objects:
https://github.com/JuliaParallel/MPI.jl/blob/v0.19.2/src/pointtopoint.jl#L451

In the current master branch, I defined a `RequestSet` object, which includes a vector of both: calling `setindex!` will update both accordingly
https://github.com/JuliaParallel/MPI.jl/blob/d34be21b2f139d1e6c4aa7b8d86f81e155417c18/src/nonblocking.jl#L204

`MPI.Waitall!` has an extra hook after completion which will go back and update the `Request` objects:
https://github.com/JuliaParallel/MPI.jl/blob/d34be21b2f139d1e6c4aa7b8d86f81e155417c18/src/nonblocking.jl#L267

An alternative approach would be to simply get rid of the mutable struct altogether (so `MPI.Request = MPI_Request`). This would avoid the copying and updates, reducing overhead, but at the cost of not having request objects being mutable by `MPI.Test` and friends: this would mean

 - `Request` objects are no longer `free`d by the GC (which, to be honest, I don't think is a big loss: I don't think much code actually allows dangling requests).
 - It would break any code that relies on checking the status of the request after `MPI.Wait` or `MPI.Test` (since we have no way to then mutate those objects).
 - `Request` objects would not be able to hold a reference to the buffer, so users would need to manually keep a reference to prevent it being gc-ed before the request has completed.

This was prompted by @OsKnoth who brought this to my attention, and is related to #639 which proposed a lower-level API

simonbyrne:
Fixes #643

### DIFF
--- a/docs/src/reference/pointtopoint.md
+++ b/docs/src/reference/pointtopoint.md
@@ -4,7 +4,6 @@
 
 ```@docs
 MPI.Request
-MPI.RequestSet
 MPI.Status
 ```
 
--- a/src/collective.jl
+++ b/src/collective.jl
@@ -29,10 +29,10 @@ If `comm` is an intercommunicator, then it blocks until all members of the other
 $(_doc_external("MPI_Ibarrier"))
 """
 function Ibarrier(comm::Comm)
-    req = Request()
+    rreq = Ref{MPI_Request}()
     # int MPI_Ibarrier(MPI_Comm comm, MPI_Req req)
-    @mpichk ccall((:MPI_Ibarrier, libmpi), Cint, (MPI_Comm, Ptr{MPI_Request}), comm, req)
-    return req
+    @mpichk ccall((:MPI_Ibarrier, libmpi), Cint, (MPI_Comm, Ptr{MPI_Request}), comm, rreq)
+    return Request(rreq[])
 end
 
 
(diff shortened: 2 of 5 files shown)


vchuravy:
The buffer rooting was kinda nice, since it prevented accidental GC errors

simonbyrne:
Yeah, that's my main reluctance to change this. Not sure what the best option is.

simonbyrne:
I think the biggest problem is going to be where we allocate the actual send buffer inside the sender, e.g.:
https://github.com/JuliaParallel/MPI.jl/blob/d34be21b2f139d1e6c4aa7b8d86f81e155417c18/src/pointtopoint.jl#L75-L76
and
https://github.com/JuliaParallel/MPI.jl/blob/d34be21b2f139d1e6c4aa7b8d86f81e155417c18/src/pointtopoint.jl#L88-L91

I'm not sure there is a good solution to this.

vchuravy:
Yeah we could have an `AbstractRequest` and then two subtypes, one of it rooting the other one doesn't.

But generally speaking I prefer the safety of the current approach and I am less worried about the cost of memory allocations for these small buffers.

simonbyrne:
Yeah, I don't see a great way around it, unfortunately.

It's not the cost of the small objects I'm worried about, it's the cost of the allocations inside the `Waitall!` (but `RequestSet` should hopefully help with that).

simonbyrne:
Closed in favor of #677.

PR #647, opened 2022-09-29
Update buildkite to use latest Julia version

simonbyrne:


simonbyrne:
@vchuravy @jpsamaroo our AMDGPU tests are failing. Any ideas?
https://buildkite.com/julialang/mpi-dot-jl/builds/854#01838a7a-34ed-4448-af7c-4aeb8ada0efa

giordano:
I pushed a commit to fix building of the documentation.

luraess:
I had a look at that AMDGPU CI issue wrt `synchronize` a week ago. `ROCQueue` should have same behaviour as `default_queue`. According to @jpsamaroo it could be a known issue on CI related to updated kernel driver not handling well very old ROCT which has no idea how to handle the new data inputed.

simonbyrne:
What's weird is that one run actually passed this point:
https://buildkite.com/julialang/mpi-dot-jl/builds/898#018391b5-af5d-4b4a-ae3d-25f93fa5a7a9

PR #648, opened 2022-09-29
add _format key to MPIPreferences

### LINKED ISSUE #642: versioning MPIPreferences metadata
We should probably add some mechanism to version the MPIPreferences metadata. The obvious way is to add a version and/or format key (similar to those in the Manifest.toml).

@giordano @vchuravy any suggestions on how best to do this?

simonbyrne:
Fixes #642.

### DIFF
--- a/docs/src/reference/mpipreferences.md
+++ b/docs/src/reference/mpipreferences.md
@@ -1,6 +1,7 @@
 # MPIPreferences.jl
 
-MPIPreferences.jl is a small package used for selecting MPI implementations.
+MPIPreferences.jl is a small package based on [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl/) for selecting MPI implementations.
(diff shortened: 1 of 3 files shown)


docs/src/reference/mpipreferences.md
```diff
@@ -21,3 +22,17 @@ MPIPreferences.use_jll_binary
 ```@docs
 MPIPreferences.identify_abi
 ```
+
+## Preferences schema
+
+MPIPreferences utilizes the following keys to store information in the Preferences key-value store.
+
+- `_format`: the version number of the schema. Currently only `"1.0"` is supported.
+- `binary`: the choice of binary. This shoudl be one of the strings listed in [`MPIPreferences.binary`](@ref).
```
giordano, line 31:
```suggestion
- `binary`: the choice of binary. This should be one of the strings listed in [`MPIPreferences.binary`](@ref).
```

giordano (approved):

PR #649, opened 2022-09-29
Import constants into `API` instead of creating non-const variables

giordano:
On `master`:
```julia
julia> isconst(MPI.API, :MPI_File)
false
```
with this PR:
```julia
julia> isconst(MPI.API, :MPI_File)
true
```

CC: @t-bltg.

### DIFF
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -79,10 +79,10 @@ include("error.jl")
 module API
     import ..libmpi, ..libmpi_handle, ..MPIPtr
     import ..use_stdcall, ..MPIError, ..@mpicall, ..@mpichk
-    using ..Consts
+    import ..Consts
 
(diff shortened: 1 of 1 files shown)


simonbyrne:
I was actually thinking we should just merge them?

giordano:
Merge what?  The `Consts` and `API` modules?

simonbyrne:
> Merge what? The `Consts` and `API` modules?

Yes. See #650.

t-bltg:
> Merge what? The Consts and API modules?

Yes, @simonbyrne mentioned this in #644, but I couldn't find a way to do this (https://github.com/JuliaParallel/MPI.jl/pull/644#discussion_r982914914).

giordano:
Closing in favour of #650.

Issue #651, opened 2022-09-30
Add automatic docstrings to functions in `API` module


giordano:
Now that we have full low-level API, I think it'd be nice to have it in the documentation.  But this would require attaching a docstring (with link to MPICH docs?) to the automatically generated methods, which at the moment don't have a type signature.  No idea how feasible is that with `Clang.jl`.

CC: @t-bltg

t-bltg:
Fair enough, I'll see what I can do (we systematically think about the docs later :laughing:).

PR #652, opened 2022-09-30
add auto-generated doc links

t-bltg:
Fix https://github.com/JuliaParallel/MPI.jl/issues/651.

```julia
julia> using MPI
help?> MPI.API.MPI_Wait
  MPI_Wait(request, status)

    •  MPI_Wait man page https://www.mpich.org/static/docs/latest/www3/MPI_Wait.html
```

Needs https://github.com/JuliaInterop/Clang.jl/pull/386.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -73,8 +73,7 @@ makedocs(
             "reference/advanced.md",
         ],
         "refindex.md",
-    ],
-    strict = true,
+    ]
 )
 
 deploydocs(
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -73,7 +73,8 @@ makedocs(
             "reference/advanced.md",
         ],
         "refindex.md",
-    ]
+    ],
+    strict = true,
 )
 
 deploydocs(
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -71,6 +71,7 @@ makedocs(
             "reference/topology.md",
             "reference/io.md",
             "reference/advanced.md",
+            "reference/api.md",
         ],
         "refindex.md",
     ],
--- a/gen/src/MPIgenerator.jl
+++ b/gen/src/MPIgenerator.jl
@@ -19,6 +19,13 @@ module MPIgenerator
 
         options = load_options(joinpath(@__DIR__, "generator.toml"))  # wrapper generator options
 
+        doc_url = if MPIPreferences.binary == "MPICH_jll"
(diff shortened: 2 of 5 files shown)


docs/make.jl
```diff
@@ -73,8 +73,7 @@ makedocs(
             "reference/advanced.md",
         ],
         "refindex.md",
-    ],
-    strict = true,
```
giordano, line 77: I'm not happy at all about this: #569.

giordano (changes requested):

docs/make.jl
t-bltg: Should we add an `API` page in the docs listing all the low level methods ?
How can I detect them automatically ?
It is not conceivable to do it by hand.

giordano: Ask Documenter to collect the functions in a given module: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/7721c30465a397cba49ee75a7647c4fbac403d96/docs/src/index.md#functions

gen/src/MPIgenerator.jl
```diff
@@ -19,6 +19,13 @@ module MPIgenerator
 
         options = load_options(joinpath(@__DIR__, "generator.toml"))  # wrapper generator options
 
+        doc_url = if MPIPreferences.binary == "MPICH_jll"
+            fname -> "- `$fname` man page `https://www.mpich.org/static/docs/latest/www3/$fname.html`"
+        else
+            fname -> "- `$fname` man page `https://www.open-mpi.org/doc/current/man3/$fname.3.php`"
+        end
+        options["general"]["callback_documentation"] = node -> String[doc_url(node.id)]
```
simonbyrne, line 27: why not link to both?

simonbyrne: Also, you can use  https://github.com/JuliaParallel/MPI.jl/blob/4e8dcd03fb428ee37abb963052cd6009a9aed45b/src/MPI.jl#L30-L34

docs/make.jl
t-bltg: Fixed, by adding `api.md`.

gen/src/MPIgenerator.jl
t-bltg: Done, both `OpenMPI` and `MPICH`.

simonbyrne (approved):

PR #655, opened 2022-10-01
add PROC_NULL, ANY_SOURCE, ANY_TAG as globals

### LINKED ISSUE #611: Handle `PROC_NULL` in a consistent way
Handle `MPI.Consts.MPI_PROC_NULL[]` in a similar fashion to `MPI_COMM_TYPE_SHARED` (related / by analogy to #584). 

In short, avoid exposing the `MPI.Const.XYZ[]` to the user.

_Note: feel free to edit this issue as the terminology may not be most accurate._

simonbyrne:
Fixes #611.

cc @luraess

### DIFF
--- a/docs/src/reference/pointtopoint.md
+++ b/docs/src/reference/pointtopoint.md
@@ -22,6 +22,14 @@ MPI.Status.error::Cint
 MPI.Get_count
 ```
 
+### Constants
+
+```@docs
+MPI.PROC_NULL
+MPI.ANY_SOURCE
+MPI.ANY_TAG
+```
+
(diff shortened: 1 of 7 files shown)


luraess:
> Fixes #611.
> 
> cc @luraess

Super! Are other consts now also accessible in a more concise way than `MPI.Const.XYZ[]`?

simonbyrne:
Which other constants do you need?

luraess:
> Which other constants do you need?

Right now nothing more, was just wondering if there was a goal design decision about syntax harmonisation.

PR #656, opened 2022-10-01
Improve low-level `API` documentation

t-bltg:
Minor https://juliaparallel.org/MPI.jl/dev/reference/api/ improvements:
- trim off trailing newline (appearing in the docstring, not in `html` files);
- ~~remove `_c` suffixed functions, as proposed by @giordano~~;
- remove singleton `MPIR_` prefixed function;
- use `MPI._doc_external`;
- simplify regex.


gen/src/MPIgenerator.jl
```diff
@@ -19,11 +19,7 @@ module MPIgenerator
 
         options = load_options(joinpath(@__DIR__, "generator.toml"))  # wrapper generator options
         options["general"]["callback_documentation"] = node -> String[
-            """
-            `$(node.id)` man page:
-            - [OpenMPI](https://www.open-mpi.org/doc/current/man3/$(node.id).3.php)
-            - [MPICH](https://www.mpich.org/static/docs/latest/www3/$(node.id).html)
-            """
+            "- man page: [OpenMPI](https://www.open-mpi.org/doc/current/man3/$(node.id).3.php), [MPICH](https://www.mpich.org/static/docs/latest/www3/$(node.id).html)"
```
giordano, line 22: Why not using `_doc_external` as [already suggested](https://github.com/JuliaParallel/MPI.jl/pull/652#discussion_r984781481)?

gen/src/generator.toml
```diff
@@ -2,4 +2,4 @@
 library_name = "libmpi"
 output_common_file_path = "./out/common.jl"
 output_api_file_path = "./out/api.jl"
-output_ignorelist = ["^PMPI_", "^QMPI_", "^MPIX_", "^PMPIX_", "^QMPIX_", "^MPI_T_", "^MPI_Session_", "MPI_Status_f082c", "MPI_Status_f082f", "MPI_Status_f2f08", "MPI_Status_c2f08", "MPI_Group_from_session_pset"]
+output_ignorelist = ["^PMPI_", "^QMPI_", "^MPIX_", "^MPIR_", "^PMPIX_", "^QMPIX_", "^MPI_T_", "^MPI_Session_", "MPI_Status_f082c", "MPI_Status_f082f", "MPI_Status_f2f08", "MPI_Status_c2f08", "MPI_Group_from_session_pset"]
```
giordano, line 5: I noticed that there are a lot of functions called `*_c` which look redundant and aren't available in OpenMPI. From MPICH manpage I can't tell the difference with the non-`*_c` ones. You may want to skip also those

gen/src/MPIgenerator.jl
giordano: I'm not sure what you. You just need to print the string `_doc_external`.

gen/src/generator.toml
t-bltg: You're right. I've also improved the regex for other filters.

gen/src/MPIgenerator.jl
t-bltg: Nice, but I can't get rid of the backslash in the docstring ... any hint ?

t-bltg: Fixed.

gen/src/generator.toml
simonbyrne: The _c ones are the "big count" ones added in MPI 4. They allow 64-bit length buffers

t-bltg: Oh, then I guess we shall restore them ?

giordano: Oh, wow, MPICH manpages are highly non descriptive: see for example https://www.mpich.org/static/docs/latest/www3/MPI_Accumulate_c.html vs https://www.mpich.org/static/docs/latest/www3/MPI_Accumulate.html.  The difference is the `_c` in the function name and the arguments types repeated twice, you can't tell anything else from that.

t-bltg: > arguments types repeated twice

They could be more descriptive, but `int` is replaced by `MPI_Count`.

Issue #599, opened 2022-05-24
Add transition guide for sysadmins to docs before v0.20 is tagged


sloede:
Right now there is only a [(very) short section](https://juliaparallel.org/MPI.jl/dev/configuration/#Migration-from-MPI.jl-version-0.19-or-earlier) on transitioning from v0.19 to v0.20 in the docs. It would be great if this could be expanded a little from a sysadmin's point of view to include information on how you have to set up your system such that you can support users who want to use the system MPI library out of the box. Since we have to assume that for a certain time both users with pre-v0.20 and post-v0.20 MPI installations will need to be supported, it would be very helpful to write down exactly what kind of environment variables and preferences file one would have to provide to allow for a smooth transition.

simonbyrne:
@sloede any interest in writing something?

sloede:
Interest yes, but no capacity at the moment 😞 For me to write this section, I'd first have to find the time to set up and try out how to do this myself, which is currently out of scope. I should be able to test and review someone else's efforts, though, if that's any help.

simonbyrne:
@sloede what are the key points you would want to address?

sloede:
> @sloede what are the key points you would want to address?

I gave it a shot in #662.

simonbyrne:
Closed by #662.

PR #658, opened 2022-10-02
bump versions

simonbyrne:
Bump versions MPIPreferences.jl and MPI.jl

Also remove some unlinked lines from the doc.


vchuravy (approved):

sloede:
I really think we should have something on #599 before the next release. It's easy to say, I know, especially since I don't think I have time to write and test a guide myself (I can help with a review if useful). However, without proper documentation, I fear we will annoy a load of people...

PR #662, opened 2022-10-03
Transition guide from v0.19 and earlier to v0.20+

### LINKED ISSUE #661: Switching to default JLL-provided binary
As far as I can tell, [`MPIPreferences.use_jll_binary`](https://juliaparallel.org/MPI.jl/dev/reference/mpipreferences/#MPIPreferences.use_jll_binary) requires to manually specify the desired JLL binary to use. IIRC, setting `JULIA_MPI_BINARY=""` (to the empty string) automatically selected the default JLL binary. This was convenient if you just wanted to restore MPI.jl to "factory defaults". Would it be possible to bring this back, i.e., a `MPIPreferences.use_jll_binary` method without the `binary` argument that then selects the default depending on the OS?

sloede:
Closes #661 and #599.

Since it was almost as difficult to write down what I had in mind for a transition guide as writing the guide itself, I tried to draft a first version of the guide. Please let me know what you think.

### DIFF
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -18,6 +18,11 @@ that link against MPI. It can be installed by
 julia -e 'using Pkg; Pkg.add("MPIPreferences")'
 ```
 
+!!! note
+    The way how MPI.jl is configured has changed with MPI.jl v0.20.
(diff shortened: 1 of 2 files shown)


simonbyrne:
Thank you. This is fantastic!

PR #664, opened 2022-10-03
check if MPI tests select correct implementation

simonbyrne:
Check that #561 is now fixed in Julia 1.8


.github/workflows/UnitTests.yml
```diff
@@ -275,15 +269,13 @@ jobs:
         using Pkg
         Pkg.develop(path="lib/MPIPreferences")
 
-    - name: add MPIPreferences, use system
+    - name: use system MPI
+      shell: julia --color=yes --project=. {0}
```
giordano, line 273: This can't use julia as shell

Issue #667, opened 2022-10-04
Intended way to generate MPI.jl compatible C bindings 


termi-official:
I am currently trying to generate MPI.jl compatible bindings of the C library HYPRE with the help Clang.jl . The current generator (found here https://github.com/termi-official/HYPRE.jl/blob/master/gen/generator.jl) creates files that have incompatible data types (see e.g. https://github.com/termi-official/HYPRE.jl/blob/master/lib/LibHYPRE.jl#L6), so I was wondering what the intended way to grab the correct MPI headers and data type definitions is?

cc @fredrikekre @Gnimuc

fredrikekre:
(Slightly tangential, I am curious why there is a mismatch between https://github.com/JuliaParallel/MPI.jl/blob/3f21294863a55fbbefa8376bd489bceec84dc478/src/api/mpich.jl#L35 and https://github.com/fredrikekre/HYPRE.jl/blob/50c06161e9b654c5095dffd6a6b6e0978f1ae851/lib/LibHYPRE.jl#L6 since they are both (apparently) generated based on MPICH headers)

termi-official:
Thanks for the quick answer and your suggestion. I already tried to use this one as a template.

```
diff --git a/gen/generator.jl b/gen/generator.jl
index d6880cb..3df8f19 100644
--- a/gen/generator.jl
+++ b/gen/generator.jl
@@ -1,10 +1,18 @@
 using Clang.Generators
-using HYPRE_jll, MPICH_jll
+using HYPRE_jll, MPIPreferences
 
 cd(@__DIR__)
 
+if MPIPreferences.binary == "MPICH_jll"
+    import MPICH_jll: artifact_dir
+elseif MPIPreferences.binary == "OpenMPI_jll"
+    import OpenMPI_jll: artifact_dir
+else
+    error("Unknown MPI binary: $(MPIPreferences.binary)")
+end
+
 hypre_include_dir = normpath(HYPRE_jll.artifact_dir, "include")
-mpi_include_dir = normpath(MPICH_jll.artifact_dir, "include")
+mpi_include_dir = normpath(artifact_dir, "include")
 
 options = load_options(joinpath(@__DIR__, "generator.toml"))
 ```

but the types are still not matching - and we generate quite a bit of redundant code which is readily available in MPI.jl.

giordano:
[...]

Issue #555, opened 2022-04-15
Spurios test failures on Windows for `get_position_shared`


vchuravy:
```
Test Failed at D:\a\MPI.jl\MPI.jl\test\test_io_shared.jl:70
  Expression: MPI.File.get_position_shared(fh) == sum(1:sz)
   Evaluated: 0 == 3
ERROR: LoadError: There was an error during testing
```
https://github.com/JuliaParallel/MPI.jl/runs/6022508469?check_suite_focus=true

We require atomicity on the file handle https://github.com/JuliaParallel/MPI.jl/blob/3b8963c7fe9a04c12aaf1725aca8cb46730d3c5c/test/test_io_shared.jl#L24, but we still seem to run into non-atomic updates...

PR #668, opened 2022-10-04
switch MPICH and MSMPI consts to signed ints

simonbyrne:
Discussed in #667

### DIFF
--- a/src/api/microsoftmpi.jl
+++ b/src/api/microsoftmpi.jl
@@ -28,16 +28,16 @@ struct MPI_Status
 end
 
 # Opaque handles:
-const MPI_Comm = UInt32
-const MPI_Datatype = UInt32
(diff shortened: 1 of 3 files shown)


src/api/microsoftmpi.jl
```diff
@@ -28,16 +28,16 @@ struct MPI_Status
 end
 
 # Opaque handles:
-const MPI_Comm = UInt32
-const MPI_Datatype = UInt32
-const MPI_Errhandler = UInt32
-const MPI_File = UInt
```
giordano, line 34: I wonder if this is related to #555

simonbyrne: maybe?

simonbyrne: not sure how though...

PR #670, opened 2022-10-05
[MPIPreferences] add functions to check (and error) if the MPI implementation has changed

simonbyrne:
Currently call `@warn` if the MPI implementation has changed: in general I don't like using `@warn` unless something is bad.

Instead it will now use `@info` to print the information, but throws an error if MPI is already loaded, or is loaded subsequently.

It also modifies the message from #641 to be a warning (since that presumably _is_ bad).

### DIFF
--- a/lib/MPIPreferences/src/MPIPreferences.jl
+++ b/lib/MPIPreferences/src/MPIPreferences.jl
@@ -197,10 +197,10 @@ end
 """
     MPIPreferences.check_unchanged()
 
-This will throw an error if the MPIPreferences have been modified in the current
-Julia session, or if they are modified after this function is called.
+Throws an error if the preferences have been modified in the current Julia
+session, or if they are modified after this function is called.
 
-This is typically called from the `__init__()` function of any package which
+This is should be called from the `__init__()` function of any package which
 relies on the values of MPIPreferences.
 """
 function check_unchanged()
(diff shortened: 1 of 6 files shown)


lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -75,7 +82,8 @@ function use_jll_binary(binary = Sys.iswindows() ? "MicrosoftMPI_jll" : "MPICH_j
         force=force
     )
 
-    @warn "The underlying MPI implementation has changed. You will need to restart Julia for this change to take effect" binary
+    PREFS_CHANGED[] = true
+    @info "MPIPreferences changed" binary
```
giordano, line 86: Should we check whether the preferences actually changed, in case one just maintained the same as before?

giordano (approved):
Agreed this should be an info, not a warning. I only have a comment about whether to check if the preferences actually changed

lib/MPIPreferences/src/MPIPreferences.jl
simonbyrne: Yeah, I think that is worthwhile.

simonbyrne:
Okay, if we're on board with this, the question is does this need a 0.2 bump of MPIPreferences? if so, do i need to rebuild all the binaries?

simonbyrne:
I guess not? We will need to bump the version requirement in MPI.jl though.

PR #673, opened 2022-10-21
add support for persistent requests

simonbyrne:
These are helpful to avoid allocations in creating request objects (#643)


src/pointtopoint.jl
```diff
@@ -255,3 +255,60 @@ function Sendrecv!(sendbuf, dest::Integer, sendtag::Integer, recvbuf, source::In
...
+    Startall(reqs::AbstractVector{Request})
+
+Start a group of persistent communication requests.
+"""
+Startall(reqs::AbstractVector{Request}) = Startall(RequestSet(reqs))
+function Startall(reqs::RequestSet)
+    API.MPI_Startall(length(reqs), reqs.vals)
+    update!(reqs)
+    return nothing
+end
```
giordano, line 314:
```suggestion
end

```

PR #675, opened 2022-11-03
cache the created Datatypes

s-fuerst:
I did run into out of memory problems, first when I used MPI.Get (which I fixed by reusing always the same Buffer) and later also in Alltoall calls. This happens when derived datatypes are used in those calls, and the MPI.Datatype constructor is called implicitly, like in this example:

```
using MPI

struct Derived
    val::Int64
end    

MPI.Init()

comm = MPI.COMM_WORLD
size = MPI.Comm_size(comm)
rank = MPI.Comm_rank(comm)

send_counts = Vector{Cint}(1:size)
recv_counts = fill(Cint(rank+1), size)

send_vals = collect(Iterators.flatten([1:i for i = 1:size])) 
send_vals = map(Derived, send_vals)

A = Array{Derived}(send_vals)
C = Array{Derived}(undef, sum(recv_counts))

while true
    MPI.Alltoallv!(VBuffer(A,send_counts), VBuffer(C,recv_counts), comm)
end

MPI.Finalize()
```

This PR fixes this problem, by caching the created Datatypes, to avoid that in each repeated Alltoallv! call the Datatype is constructed (incl. a API.MPI_Type_create_struct call) and commited again.

### DIFF
--- a/src/datatypes.jl
+++ b/src/datatypes.jl
@@ -129,21 +129,28 @@ for (mpiname, T) in [
 end
 _defined_datatype_methods = nothing
 
+# Cache the created datatypes. The datatype constructor is often
+# called for the same type, e.g. when the Buffer object is implicitly
+# constructed in MPI.Get. Without the cache, each Get would commit the
+# same datatype over and over again.
+const created_datatypes = Dict{Type, Datatype}()
 
 function Datatype(::Type{T}) where {T}
-    datatype = Datatype()
-    # lazily initialize so that it can be safely precompiled
-    function init()
-        Types.create!(datatype, T)
-        Types.commit!(datatype)
-        set_attr!(datatype, JULIA_TYPE_PTR_ATTR[], pointer_from_objref(T))
+    global created_datatypes
+    get!(created_datatypes, T) do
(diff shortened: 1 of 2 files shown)


src/datatypes.jl
```diff
@@ -129,22 +129,29 @@ for (mpiname, T) in [
 end
 _defined_datatype_methods = nothing
 
+# Cache the created datatypes. The datatype constructor is often
+# called for the same type, e.g. when the Buffer object is implicitly
+# constructed in MPI.Get. Without the cache, each Get would commit the
+# same datatype over and over again.
+const created_datatypes = Dict{Type, Datatype}()
```
vchuravy, line 136:
```suggestion
const created_datatypes = IdDict{Type, Datatype}()
```

simonbyrne:
Would you be able to add a test?

s-fuerst:
I'm not sure what I should test. The changed Datatype constructor is already part of many tests. Should I write something, that checks that this fix really solves the memory leak?

simonbyrne:
Thanks!

PR #676, opened 2022-11-05
increase timeout

simonbyrne:


simonbyrne:
I think we need to call MPI.Abort from our atexit hook

src/environment.jl
```diff
@@ -98,12 +98,16 @@ function Init(;threadlevel=:serialized, finalize_atexit=true, errors_return=true
...
+                # MPI_Finalize is a collective and can act like a barrier (this
+                # may be implementation specific). If we are terminating due to
+                # a Julia exception, we should call MPI_Abort instead (which
+                # will attempt to abort other processes).
+                if !Finalized()
+                    if ccall(:jl_current_exception, Any, ()) === nothing
+                        Finalize()
+                    else
+                        Abort(MPI.COMM_WORLD, 1)
+                    end
```
simonbyrne, line 110: @vchuravy does this seem okay to you?

simonbyrne: Actually, failures seem to be only on nightly: did `jl_current_exception` change?

simonbyrne:
Yup, it's an issue on nightly: https://github.com/JuliaLang/julia/issues/47472

simonbyrne:
Fixed now in #680

PR #678, opened 2022-11-09
check for length 0 in Win_shared_query

s-fuerst:
The Intel MPI (2018.6.288) implementation returns in Win_shared_query invalid pointers in the case that the array on the owner_rank has length 0. In this case Win_shared_query should create an empty Array of type $T, but unsafe_wrap checks that the alignment of the return pointer is correct, which is now not always the case as the pointer is just uninitialized memory.  

So this pull request checks the returned length, and calls unsafe_wrap only when the length is greater then zero.

### DIFF
--- a/src/onesided.jl
+++ b/src/onesided.jl
@@ -130,8 +130,8 @@ end
 function Win_shared_query(::Type{Array{T}}, win::Win, owner_rank::Integer) where T
     len, sizeT, ptr = Win_shared_query(Ptr{T}, win, owner_rank)
     sizeT == sizeof(T) || error("type sizes don't match")
-    # the ptr can be invalid for empty arrays but unsafe_wrap always
-    # check the alignment of ptr, even for lenght 0
+    # the ptr may be invalid for empty arrays, which will cause an error as unsafe_wrap
+    # checks the alignment of ptr, even for length 0
     if len > 0
         return unsafe_wrap(Array, ptr, div(len, sizeT))
     else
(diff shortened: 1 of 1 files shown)


giordano:
Any test that can be added?

s-fuerst:
The current Win_shared_query test already has empty arrays on all but one rank, so both branches are covered.

s-fuerst:
For whatever reason this only happens when I use multiple nodes, maybe this explains why the problem wasn't detected by the tests so far? E.g. I run the test suite after installation, but only on a single node. 

I have not checked other implementations.

giordano:
Ugh, ok, that's unfortunate.

src/onesided.jl
```diff
@@ -130,7 +130,13 @@ end
 function Win_shared_query(::Type{Array{T}}, win::Win, owner_rank::Integer) where T
     len, sizeT, ptr = Win_shared_query(Ptr{T}, win, owner_rank)
     sizeT == sizeof(T) || error("type sizes don't match")
-    return unsafe_wrap(Array, ptr, div(len, sizeT))
+    # the ptr can be invalid for empty arrays but unsafe_wrap always
+    # check the alignment of ptr, even for lenght 0
```
simonbyrne, line 134:
```suggestion
    # the ptr may be invalid for empty arrays, which will cause an error as unsafe_wrap
    # checks the alignment of ptr, even for length 0
```

simonbyrne:
Thanks!

Issue #688, opened 2022-12-11
`Char` datatype isn't handled correctly


giordano:
I'm not sure the `Char` datatype is handled correctly throughout `MPI.jl`.  For example, with `MPICH_jll` I get this on my machine:
```julia
julia> using MPI

julia> MPI.Init()
MPI.ThreadLevel(2)

julia> MPI.Datatype(Char)
MPI.Datatype(Char): 

julia> MPI.get_name(MPI.Datatype(Char))
""

julia> MPI.get_name.(MPI.Datatype.(Base.uniontypes(MPI.MPIDatatype)))
13-element Vector{String}:
 "MPI_C_COMPLEX"
 "MPI_C_DOUBLE_COMPLEX"
 ""
 "MPI_FLOAT"
 "MPI_DOUBLE"
 "MPI_INT16_T"
 "MPI_INT32_T"
 "MPI_INT64_T"
 "MPI_INT8_T"
 "MPI_UINT16_T"
 "MPI_UINT32_T"
 "MPI_UINT64_T"
 "MPI_UINT8_T"
```
`Char` is the only element of `MPI.MPIDatatype` (BTW, this constant is only ever used in the tests, so I'm not sure why it's defined in `src/`)
https://github.com/JuliaParallel/MPI.jl/blob/a179cf876a52876d297b7e635881dbc7a2f3a81c/src/buffers.jl#L5-L8
for which [`get_name`](https://github.com/JuliaParallel/MPI.jl/blob/a179cf876a52876d297b7e635881dbc7a2f3a81c/src/datatypes.jl#L59-L65) returns an empty string, and my understanding is that this is kinda an indication MPI doesn't really know about this type.  @eschnett [suggested](https://github.com/JuliaParallel/MPI.jl/issues/684#issuecomment-1341860002) `Char` should behave closer to `UInt32`, however I'm missing where/how this should be fixed.

`MPI.jl` behaving weirdly with `Char` is probably the cause of most issues in #684 but not all of them, so I opened this issue to track specifically the problems with `Char`.

giordano:
From https://github.com/JuliaParallel/MPI.jl/pull/689#issue-1492688679

> An easy fix to #688.

I don't think #689 was a real fix for this issue, we're just skipping the `Char` tests, but the `Char` datatype is still used incorrectly and https://github.com/JuliaParallel/MPI.jl/issues/684#issuecomment-1341698748 still segfaults for me with
```console
[...]

PR #689, opened 2022-12-12
remove Char from MPIDatatypes

simonbyrne:
An easy fix to #688.

cc @eschnett @giordano

### DIFF
--- a/src/buffers.jl
+++ b/src/buffers.jl
@@ -2,9 +2,8 @@ const MPIInteger = Union{Int8, UInt8, Int16, UInt16, Int32, UInt32, Int64, UInt6
 const MPIFloatingPoint = Union{Float32, Float64}
 const MPIComplex = Union{ComplexF32, ComplexF64}
 
-const MPIDatatype = Union{Char,
-                            Int8, UInt8, Int16, UInt16, Int32, UInt32, Int64,
-                            UInt64,
+const MPIDatatype = Union{Int8, UInt8, Int16, UInt16,
+                          Int32, UInt32, Int64, UInt64,
                             Float32, Float64, ComplexF32, ComplexF64}
 MPIBuffertype{T} = Union{Ptr{T}, Array{T}, SubArray{T}, Ref{T}}
 


simonbyrne:
We now have:
```
julia> using MPI

julia> MPI.Init()
MPI.ThreadLevel(2)

julia> MPI.Datatype(UInt32)
MPI.Datatype(UInt32): MPI_UINT32_T

julia> MPI.Datatype(Char)
MPI.Datatype(Char): Dup MPI_UINT32_T
```

giordano:
Ok, let me try this PR on Ookami again, but I think I was still getting some segfaults when I last tried.

giordano:
Ok, this does solve the crash in `test_allgather`, but then I still run into other segfaults in `test_bcast` and `test_datatypes` (and haven't gone further with this PR), which I had already mentioned in https://github.com/JuliaParallel/MPI.jl/issues/684#issuecomment-1341937162 and don't seem to be fixed by removing `Char` from `MPIDatatypes`, so there may be something else, I'd need to track it down.

This PR is probably OK anyway, but perhaps we can add a test like
```julia
all(.!(isempty.(MPI.get_name.(MPI.Datatype.(Base.uniontypes(MPI.MPIDatatype))))))
```
?

Side note, I'm still missing where `MPIDatatype` is used in `MPI.jl/src` :thinking:

simonbyrne:
> Side note, I'm still missing where `MPIDatatype` is used in `MPI.jl/src` 🤔

It isn't. I think it's just a vestige of some earlier code with how we handled datatypes and buffers.

giordano:
But this change does affect `MPI.Datatype(Char)`: https://github.com/JuliaParallel/MPI.jl/pull/689#issuecomment-1347253030

simonbyrne:
> But this change does affect `MPI.Datatype(Char)`: [#689 (comment)](https://github.com/JuliaParallel/MPI.jl/pull/689#issuecomment-1347253030)

Actually it doesn't: I get the same on `master`

giordano:
What MPI are you using?

simonbyrne:
Ah, it was Open MPI.

PR #691, opened 2022-12-14
[docs] Expand known issues

giordano:
Mention new problem with HCOLL, and make it clear that custom operators don't work on ARM architecture either.

### DIFF
--- a/docs/src/knownissues.md
+++ b/docs/src/knownissues.md
@@ -36,9 +36,32 @@ This can be worked around be either:
 
 ## Open MPI
 
-### Segmentation fault
+### Segmentation fault when loading the library
 
-When attempting to use a system-provided Open MPI implementation, you may encounter a segmentation fault. This can be fixed by setting the environment variable `ZES_ENABLE_SYSMAN=1`. See [Open MPI issue #10142](https://github.com/open-mpi/ompi/issues/10142).
+When attempting to use a system-provided Open MPI implementation, you may encounter a segmentation fault upon loading the library, or whenever the value of an environment variable is requested.
+This can be fixed by setting the environment variable `ZES_ENABLE_SYSMAN=1`.
+See [Open MPI issue #10142](https://github.com/open-mpi/ompi/issues/10142) for more details.
+
+### Segmentation fault in HCOLL
+
+If Open MPI was built with support for HCOLL, you may encounter a segmentation fault in certain operations involving custom datatypes.
+The stacktrace may look something like
+
+```
+hcoll_create_mpi_type at /opt/mellanox/hcoll/lib/libhcoll.so.1 (unknown line)
+ompi_dtype_2_hcoll_dtype at /lustre/software/openmpi/llvm14/4.1.4/lib/openmpi/mca_coll_hcoll.so (unknown line)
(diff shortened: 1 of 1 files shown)


docs/src/knownissues.md
```diff
@@ -36,9 +36,32 @@ This can be worked around be either:
 
 ## Open MPI
 
-### Segmentation fault
+### Segmentation fault when loading the library
 
-When attempting to use a system-provided Open MPI implementation, you may encounter a segmentation fault. This can be fixed by setting the environment variable `ZES_ENABLE_SYSMAN=1`. See [Open MPI issue #10142](https://github.com/open-mpi/ompi/issues/10142).
+When attempting to use a system-provided Open MPI implementation, you may encounter a segmentation fault upon loading the library, or whenever the value of an environment variable is requested.
+This can be fixed by setting the environment variable `ZES_ENABLE_SYSMAN=1`.
+See [Open MPI issue #10142](https://github.com/open-mpi/ompi/issues/10142) for more details.
```
giordano, line 43: Is this still a problem in practice?  We try to work around it at https://github.com/JuliaParallel/MPI.jl/blob/096574032bfe6bde8ced4282f0ca458515c17dd2/lib/MPIPreferences/src/MPIPreferences.jl#L149 Is this not enough?

simonbyrne: That's only used when detecting the library. I think it may still be required when actually using MPI.jl

docs/src/knownissues.md
```diff
@@ -129,8 +152,11 @@ Make sure to:
 
 After that, [this script](https://gist.github.com/luraess/c228ec08629737888a18c6a1e397643c) can be used to verify if ROCm-aware MPI is functional (modified after the CUDA-aware version from [here](https://discourse.julialang.org/t/cuda-aware-mpi-works-on-system-but-not-for-julia/75060/11)). It may be preferred to run the Julia ROCm-aware MPI script launching it from a shell script (as suggested [here](https://discourse.julialang.org/t/cuda-aware-mpi-works-on-system-but-not-for-julia/75060/4)).
 
-## Microsoft MPI
+## Custom reduction operators
 
-### Custom operators on 32-bit Windows
+It is not possible to use custom reduction operators [with 32-bit Microsoft MPI](https://github.com/JuliaParallel/MPI.jl/issues/246) on Windows and on [ARM CPUs](https://github.com/JuliaParallel/MPI.jl/issues/404) with any operating system.
```
simonbyrne, line 157:
```suggestion
It is not possible to use custom reduction operators [with 32-bit Microsoft MPI](https://github.com/JuliaParallel/MPI.jl/issues/246) on Windows and on [ARM or PowerPC CPUs](https://github.com/JuliaParallel/MPI.jl/issues/404) with any operating system.
```

giordano: My understanding is that this now works on PowerPC.  @vchuravy?

giordano: See https://github.com/JuliaLang/julia/issues/32154, now closed as resolved.

simonbyrne: Ah, good to know!

PR #692, opened 2022-12-16
Use `@ccall` for autogenerated API

giordano:
Ref: https://github.com/JuliaParallel/MPI.jl/issues/688#issuecomment-1354010721.  This isn't complete yet, I haven't fixed the `MPIPtr` replacements, but I'm saving the work so far before going to sleep (fixing `@mpicall` was already a fun ride, I hope I got it right) :slightly_smiling_face: 

***Edit***: I'm not really sure how to specify the calling convention in `@ccall`, I haven't touched forcing `stdcall` with msmpi, I asked for help in Slack.

***Edit 2***: https://docs.julialang.org/en/v1.10-dev/manual/calling-c-and-fortran-code/#ccall-interface

> This interface is slightly less convenient but it does allow one to specify a [calling convention](https://docs.julialang.org/en/v1.10-dev/manual/calling-c-and-fortran-code/#calling-convention).

Sigh

### DIFF
--- a/src/api/api.jl
+++ b/src/api/api.jl
@@ -81,14 +81,25 @@ end
 
 const use_stdcall = startswith(basename(libmpi), "msmpi")
 
-macro mpicall(expr)
-    @assert expr isa Expr && expr.head == :call && expr.args[1] == :ccall
+macro mpicall(expr::Expr)
+    is_ccall = expr.head == :call && expr.args[1] == :ccall
+    is_at_ccall = expr.head == :macrocall && expr.args[1] == Symbol("@ccall")
+    if !is_ccall && !is_at_ccall
+        error("@mpicall can only be applied to `ccall`/`@ccall` expressions!")
+    end
 
     # On unix systems we call the global symbols to allow for LD_PRELOAD interception
     # It can be emulated in Windows (via Libdl.dllist), but this is not fast.
-    if Sys.isunix() && expr.args[2].head == :tuple &&
-            (VERSION ≥ v"1.5-" || expr.args[2].args[1] ≠ :(:MPI_Get_library_version))
-        expr.args[2] = expr.args[2].args[1]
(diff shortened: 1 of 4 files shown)


src/api/api.jl
```diff
@@ -81,14 +81,25 @@ end
...
+    is_ccall = expr.head == :call && expr.args[1] == :ccall
+    is_at_ccall = expr.head == :macrocall && expr.args[1] == Symbol("@ccall")
+    if !is_ccall && !is_at_ccall
+        error("@mpicall can only be applied to `ccall`/`@ccall` expressions!")
+    end
 
     # On unix systems we call the global symbols to allow for LD_PRELOAD interception
     # It can be emulated in Windows (via Libdl.dllist), but this is not fast.
-    if Sys.isunix() && expr.args[2].head == :tuple &&
-            (VERSION ≥ v"1.5-" || expr.args[2].args[1] ≠ :(:MPI_Get_library_version))
```
giordano, line 90: Am I correct to say that this line is always true since we support Julia v1.6+ now?

Gnimuc:
I can add a config to Clang.jl that supports generating custom "ccall" macros. For example, 

```toml
[codegen]
use_ccall_macro = true
ccall_macro_name = "stdcall" # defaults to "ccall"
```

Then, we can define the `stdcall` macro in MPI.jl:

```julia
macro stdcall(expr)
    return Base.ccall_macro_lower(:stdcall, ccall_macro_parse(expr)...)
end
```

giordano:
As much as I personally like the `@ccall` macro, I'm not sure there is much motivation for switching to it at the moment given that handling the `stdcall` calling convention would add more complexity than we really need, but it's great to know about `Base.ccall_macro_lower(:stdcall, Base.ccall_macro_parse(expr)...)`, that'd be very useful in case we want to eventually switch to `@ccall`!

PR #693, opened 2022-12-16
rework c function pointers

t-bltg:
Fix https://github.com/JuliaParallel/MPI.jl/issues/688.

~~Not yet ready for review ;)~~


gen/src/MPIgenerator.jl
```diff
@@ -10,6 +11,39 @@ module MPIgenerator
...
+            if startswith(signature, "MPI_Type_create_keyval") || true
+                # @show signature node.cursor 
+                for a in get_function_args(node.cursor)
+                    # @show a, typeof(a)
+                    if (ct = Clang.getCursorType(a)) isa CLPointer
+                        # @show translate(tojulia(ct))
+                        # if Clang.Generators.is_jl_funcptr(tojulia(ct))
+                        if (pt = Clang.getPointeeType(ct)) isa CLTypedef
+                            dumpobj(a)
+                            # FIXME: how do we replace node function arguments ?
```
t-bltg, line 27: @Gnimuc, could you advise how to replace some node arguments ?

gen/src/MPIgenerator.jl
```diff
@@ -10,6 +11,39 @@ module MPIgenerator
         error("Unknown MPI binary: $(MPIPreferences.binary)")
     end
 
+    rewrite!(dag::ExprDAG) = begin
+        replace!(get_nodes(dag)) do node
+            signature = Clang.name(node.cursor)
+
+            if startswith(signature, "MPI_Type_create_keyval") || true
```
giordano, line 18: This isn't the only one wrong, all pointers which aren't buffers should be changed

t-bltg: Yes, I know, this is a draft.
I'm only opening this PR now to discuss Clang internals with @Gnimuc.

Gnimuc:
Is it still a problem after merging https://github.com/JuliaInterop/Clang.jl/pull/406 ?

t-bltg:
https://github.com/JuliaInterop/Clang.jl/pull/406 is only for fixing debugging and printing stuff.

The question I want to ask here is: "is it possible to modify node function arguments" ?

Or do we have to modify the underlying node exprs generated (this seems unfeasible, since information is lost on the way ...) ?

Gnimuc:
> (this seems unfeasible, since information is lost on the way ...)

The only elegant way is to overload [`_get_func_arg `](https://github.com/JuliaInterop/Clang.jl/blob/master/src/generator/codegen.jl#L23).

t-bltg:
This was fun actually.
Monkey patching `_get_func_arg` is horrible, but it works ...

Gnimuc:
I'm trying to understand the root cause of #688. To me, it seems like Clang.jl doesn't handle `typedef function prototype` well for MPI's special usage.

t-bltg:
By the way, is the `output_ignorelist` option from `generator.toml` broken ?

It doesn't filter the function definitions anymore ...

Gnimuc:
Only on the master branch? This could be related: https://github.com/JuliaInterop/Clang.jl/pull/403

giordano:
Worked for me in #692

t-bltg:
> Only on the master branch?

Ha ok, we have to adjust the generator regexes.

t-bltg:
> I'm trying to understand the root cause of https://github.com/JuliaParallel/MPI.jl/issues/688. To me, it seems like Clang.jl doesn't handle typedef function prototype well for MPI's special usage.

Right, I think we'd better fix this:
The declaration in `mpi.h` is:
```c
typedef int (MPI_Type_copy_attr_function)(MPI_Datatype, int, void *, void *, 
					  void *, int *);
```

t-bltg:
All-right, so with the patch fixing function pointer (https://github.com/JuliaInterop/Clang.jl/pull/408), the local MPI test are green on linux.

t-bltg:
So, with the `Clang.jl` function pointer patch in https://github.com/JuliaInterop/Clang.jl/pull/408, I think I got this right now.

I also [got confused with c function pointers](https://github.com/JuliaParallel/MPI.jl/issues/688#issuecomment-1353919247) ...

gen/src/MPIgenerator.jl
```diff
@@ -1,4 +1,5 @@
 module MPIgenerator
+    __precompile__(false)
```
giordano, line 2: We only need to update `Clang.jl`, with no other changes to the script?

t-bltg: Yeah, I can remove this now, since this was used at early stage when monkey patching.

t-bltg:
I also need to add a test for https://github.com/JuliaParallel/MPI.jl/issues/688.

giordano:
> I also need to add a test for https://github.com/JuliaParallel/MPI.jl/issues/688.

Revert #689 :stuck_out_tongue:

giordano:
> @giordano, what is in your test.jl from https://github.com/JuliaParallel/MPI.jl/issues/688 ?

https://github.com/JuliaParallel/MPI.jl/issues/684#issuecomment-1341698748, which is simply extracted from https://github.com/JuliaParallel/MPI.jl/blob/7da078cf179a4d39e555713acaa74c84dbbf6729/test/test_allgather.jl#L3-L13 and before #689 that was running on `Char` as well.  Also, `MPIDatatype` should be moved to the tests.

test/test_issues.jl
```diff
@@ -0,0 +1,15 @@
+include("common.jl")
+
+MPI.Init(; threadlevel=:single)
+
+@testset "#688" begin
```
giordano, line 5: This is redundant if you restore `Char` in `MPIDatatype` (which I this point I believe is the right thing to do), the same code is executed in `test_allgather.jl` already.

t-bltg: I prefer to be explicit about it, but your call. Removed.

giordano:
I can confirm the test script works with this branch of MPI.jl on Ookami with MVAPICH:
```console
[mosgiordano@fj004 mvapich]$ srun -n 2 julia --project test.jl 
(rank, C) = (0, ['\x01', '\x02'])
(rank, C) = (1, ['\x01', '\x02'])
[mosgiordano@fj004 mvapich]$ srun -n 4 julia --project test.jl 
(rank, C) = (1, ['\x01', '\x02', '\x03', '\x04'])
(rank, C) = (2, ['\x01', '\x02', '\x03', '\x04'])
(rank, C) = (3, ['\x01', '\x02', '\x03', '\x04'])
(rank, C) = (0, ['\x01', '\x02', '\x03', '\x04'])
[mosgiordano@fj004 mvapich]$ srun -n 8 julia --project test.jl 
(rank, C) = (3, ['\x01', '\x02', '\x03', '\x04', '\x05', '\x06', '\a', '\b'])
(rank, C) = (2, ['\x01', '\x02', '\x03', '\x04', '\x05', '\x06', '\a', '\b'])
(rank, C) = (5, ['\x01', '\x02', '\x03', '\x04', '\x05', '\x06', '\a', '\b'])
(rank, C) = (6, ['\x01', '\x02', '\x03', '\x04', '\x05', '\x06', '\a', '\b'])
(rank, C) = (7, ['\x01', '\x02', '\x03', '\x04', '\x05', '\x06', '\a', '\b'])
(rank, C) = (1, ['\x01', '\x02', '\x03', '\x04', '\x05', '\x06', '\a', '\b'])
(rank, C) = (0, ['\x01', '\x02', '\x03', '\x04', '\x05', '\x06', '\a', '\b'])
(rank, C) = (4, ['\x01', '\x02', '\x03', '\x04', '\x05', '\x06', '\a', '\b'])
```

:rocket:

t-bltg:
> I can confirm the test script works with this branch of MPI.jl on Ookami with MVAPICH:

So
```julia
julia> MPI.get_name(MPI.Datatype(Char))
""
```
remains expected ?

test/common.jl
```diff
@@ -19,3 +19,8 @@ else
     ArrayType = Array
     synchronize() = nothing
 end
+
+const MPIUnionTypes = Union{Char,
+                            Int8, Int16, Int32, Int64,
+                            UInt8, UInt16, UInt32, UInt64,
+                            Float32, Float64, ComplexF32, ComplexF64} |> Base.uniontypes
```
giordano, line 26: Missing newline
```suggestion
                            Float32, Float64, ComplexF32, ComplexF64} |> Base.uniontypes

```
Maybe also add a comment to explain these are the types we want to test with MPI and warn to not remove any of them? :smile:

giordano:
> So
> ```julia
> julia> MPI.get_name(MPI.Datatype(Char))
> ""
> ```
> remains expected ?

Sadly yes:

```
julia> using MPI

julia> MPI.Init(;threadlevel=:single)
MPI.ThreadLevel(0)

julia> MPI.get_name(MPI.Datatype(Char))
""
```

I get the same with `MPICH_jll` (and MVAPICH has the MPICH ABI, so this shouldn't be surprising).

I don't know if there is a way to query the "MPI equivalent type" for `Char`.

test/common.jl
t-bltg: Done. Also, I don't think we need `Base.uniontypes` right ?
IIUC, listing the types in a vector should be sufficient.

t-bltg:
Let's reserve that fix for another PR.

test/common.jl
giordano: Yeah, can also just be a vector.  In #689 suggested this is just a leftover from the past, so we can change it for whatever we need in the tests, which is always an iterable from what I can tell.

simonbyrne: It might be worth keeping some type alias for native MPI types (which would exclude Char)

t-bltg: Sorry I don't get this could you clarify ?

giordano:
I spoke too soon, I hadn't closed the julia session:
```
julia> using MPI

julia> MPI.Init(;threadlevel=:single)
MPI.ThreadLevel(0)

julia> MPI.Datatype(Char)
MPI.Datatype(Char): 

julia> MPI.Finalize()

[813476] signal (11.1): Segmentation fault
in expression starting at REPL[4]:1
MPIR_Call_attr_delete at /lustre/software/mvapich2/gcc11/2.3.6/lib/libmpi.so (unknown line)
MPIR_Attr_delete_list at /lustre/software/mvapich2/gcc11/2.3.6/lib/libmpi.so (unknown line)
Allocations: 644473 (Pool: 643990; Big: 483); GC: 1
Segmentation fault (core dumped)
[mosgiordano@fj-debug2 mvapich]$
```
As I mentioned in https://github.com/JuliaParallel/MPI.jl/issues/688#issuecomment-1353846363, lots of segfaults happen at `MPI_Finalize`-time.

Also, this PR does ***not*** solve #684, there's a load of other segmentation faults I get there, with both MVAPICH2 and Open MPI, this is just the first one I got with MVAPICH.

t-bltg:
> Also, this PR does not solve https://github.com/JuliaParallel/MPI.jl/issues/684, there's a load of other segmentation faults I get there

Ok, I will remove the `fix ...`.

t-bltg:
By the way, buildkite should have a some timeouts, there are jobs running for a week now:
https://buildkite.com/julialang/mpi-dot-jl/builds/999
https://buildkite.com/julialang/mpi-dot-jl/builds/1001

src/buffers.jl
```diff
@@ -2,9 +2,6 @@ const MPIInteger = Union{Int8, UInt8, Int16, UInt16, Int32, UInt32, Int64, UInt6
 const MPIFloatingPoint = Union{Float32, Float64}
 const MPIComplex = Union{ComplexF32, ComplexF64}
 
-const MPIDatatype = Union{Int8, UInt8, Int16, UInt16,
-                          Int32, UInt32, Int64, UInt64,
-                            Float32, Float64, ComplexF32, ComplexF64}
```
simonbyrne, line 7: Can we keep this for now? These are the "native" MPI types (i.e. the ones that exist out of the box)

simonbyrne:
> By the way, buildkite should have some timeouts, there are jobs running for a week now: https://buildkite.com/julialang/mpi-dot-jl/builds/999 https://buildkite.com/julialang/mpi-dot-jl/builds/1001

Not currently possible, unfortunately:
https://forum.buildkite.community/t/timeout-waiting-for-agent/188/4

src/buffers.jl
t-bltg: Sure.

simonbyrne (approved):

simonbyrne:
Ah, sorry i merged too early.

@giordano doesn't this include the change from https://github.com/JuliaParallel/MPI.jl/issues/688#issuecomment-1353923564?

giordano:
I think it does: https://github.com/JuliaParallel/MPI.jl/pull/693/files#diff-b01acb82d6f2ed53e2d5907924a42adb7162f0e91fbfe6beb70aaaae0d0e2c36R162 ?

simonbyrne:
Great!

Issue #681, opened 2022-11-30
Non-blocking version of recv (irecv)


ClarkGuilty:
Hello.

I remember there being a non-blocking version of serialized Recv (irecv). Was this deprecated, if so, why? Is there an already established alternative?

Thank you.

simonbyrne:
It still exists, but has been deprecated:
https://github.com/JuliaParallel/MPI.jl/blob/2e94c5b54054e6343133f22a8b9a164e341d4189/src/deprecated.jl#L239-L246

It was undocumented, and honestly it doesn't really make sense how its defined: it's not really analogous to `Irecv` since it didn't return a `Request` object (instead you have to keep calling it). Also in a multithreaded context, there is no guarantee that the `Recv` will match the same operation as the `Iprobe`. To be done properly:
- It should use `MPI_Improbe` and `MPI_Mrecv`
- It should be implemented via the MPI "Generalized Request" interface.

simonbyrne:
If you do have need for it, we could add it back, but give it a more reasonable name (`test_recv`?)

ClarkGuilty:
Thank you very much! I do need the function and have been trying to implement it (for now, only in the scope of my project but with the hope of eventually making a PR), but I am finding it hard to interact with MPI.API (this is the first time I use C bindings in Julia). I cannot use the deprecated version because the syntax has changed since it was deprecated (I believe) and the signature no longer works. Thus, I have a question regarding MPI.API.Improbe

Reading the Docs, the way to use MPI.API.Improbe is very similar to MPI.API.Iprobe, the only difference being the message pointer, However I keep getting:
```Julia
MethodError: no method matching unsafe_convert(::MethodError: no method matching unsafe_convert(::Type{Ptr{Ptr{Nothing}}}, ::Bas
e.RefValue{Int32})Type{Ptr{Ptr{Nothing}}}, ::Base.RefValue{Int32})
Closest candidates are:
  unsafe_convert(::Type{Ptr{T}}, !Matched::Union{NNlib.BatchedAdjoint{T, S}, NNlib.BatchedTranspose{T, S}} where S) where T at ~/.julia/packages/NNlib/c0XLe/src/bat
ched/batchedadjtrans.jl:90
  unsafe_convert(::Type{Ptr{T}}, !Matched::Base.Threads.Atomic{T}) where T at atomics.jl:328
[...]

PR #699, opened 2022-12-31
add matched recv functions

simonbyrne:
Add matched probe/receive operations. Discussed in #681

### DIFF
--- a/src/pointtopoint.jl
+++ b/src/pointtopoint.jl
@@ -321,7 +321,13 @@ function Startall(reqs::Union{AbstractMultiRequest, RequestSet})
 end
 
 
+"""
+    MPI.Message
 
+An MPI message handle object, used by matched receive operations. These are
+returned by [`MPI.Mprobe`](@ref) and [`MPI.Improbe`](@ref) operations, and must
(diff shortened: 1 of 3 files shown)


src/pointtopoint.jl
```diff
@@ -319,3 +319,136 @@ function Startall(reqs::Union{AbstractMultiRequest, RequestSet})
...
+    MPI.Message
+
+An MPI message handle object, used by matched receive operations. These are
+returned by [`MPI.Mprobe`](@ref) and [`MPI.Improbe`](@ref) operations, and must
+be received by either [`MPI.Mrecv!`](@ref) or [`MPI.Imrecv!`](@ref).
+"""
+mutable struct Message
+    val::MPI_Message
+end
+Base.cconvert(::Type{MPI_Message}, msg::Message) = msg
```
vchuravy, line 334: I think this definition is uncessary?

vchuravy (approved):

src/pointtopoint.jl
simonbyrne: I think you're right

PR #704, opened 2023-01-12
[CI] Use newer versions of some GitHub actions

giordano:


giordano:
This PR clears most warnings when running GitHub Actions CI, compare the annotations in

* [the last full CI run on `master`](https://github.com/JuliaParallel/MPI.jl/actions/runs/3833885124) (156 warnings)
* [this PR](https://github.com/JuliaParallel/MPI.jl/actions/runs/3901349361) (6 warnings).

As far as I can tell, all remaining warnings are from the use of the deprecated `set-output` in https://github.com/vsoch/spack-package-action/blob/2f0a9359c56906d4f56df568d2c991f8968cfca2/envpackage/scripts/build.sh#L58-L74, plus a couple of notices that `macOS-latest` will become `macOS-12` soon, but that's not worrying.

PR #713, opened 2023-01-29
Add section on external libraries

### LINKED ISSUE #702: Document how to use in other packages
From Slack, we need to document how to use with other packages, especially when working with binaries that use MPI.

Common issues:
- how to pass MPI objects to C (partially described in #626)
- object finalizers (which call MPI functions) being invoked after `atexit`
  - either gate via `MPI.Finalized()`, or invoke when `MPI.Finalize` is called via `MPI.add_finalize_hook!` (e.g. https://github.com/fredrikekre/HYPRE.jl/pull/8)

simonbyrne:
Fixes #702

### DIFF
--- a/docs/src/external.md
+++ b/docs/src/external.md
@@ -27,7 +27,7 @@ External libraries may maintain allocate their own MPI handles (e.g.  create or 
 
 There are two typical solutions to this problem:
 
-1. Gate the clean up functions behind an [`MPI.Finalized](@ref) call, e.g.
+1. Gate the clean up functions behind an [`MPI.Finalized`](@ref) call, e.g.
 
    ```julia
    finalizer(obj) do obj
--- a/docs/src/external.md
+++ b/docs/src/external.md
@@ -8,7 +8,7 @@ You need to ensure that external libraries are built correctly. In particular, i
 
 ## Passing MPI handles via `ccall`
 
-When passing MPI.jl handle objects ([`MPI.Comm`](@ref), [`MPI.Info`](@ref), etc) to C/C++ functions [via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/), you should pass the object directly as an argument, and specify the argument type as either the underlying handle type (`MPI.API.MPI_Comm`, `MPI.API.MPI_Info`, etc.), or a pointer (`Ptr{MPI.API.MPI_Comm}`, `Ptr{MPI.API.MPI_Info}`, etc.). This will internally handle the unwrapping, but ensure that a reference is kept to avoid premature garbage collection.
+When passing MPI.jl handle objects ([`MPI.Comm`](@ref), [`MPI.Info`](@ref), etc) to C/C++ functions [via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/), you should pass the object directly as an argument, and specify the argument type as either the underlying handle type (`MPI.MPI_Comm`, `MPI.API.MPI_Info`, etc.), or a pointer (`Ptr{MPI.MPI_Comm}`, `Ptr{MPI.MPI_Info}`, etc.). This will internally handle the unwrapping, but ensure that a reference is kept to avoid premature garbage collection.
 
 For example the C function signatures
 ```C
@@ -17,8 +17,8 @@ int cfunc2(MPI_Comm * comm);
 ```
 would be called as
 ```julia
-ccall((:cfunc1, lib), Cint, (MPI.API.MPI_Comm,), comm)
-ccall((:cfunc2, lib), Cint, (Ptr{MPI.API.MPI_Comm},), comm)
+ccall((:cfunc1, lib), Cint, (MPI.MPI_Comm,), comm)
+ccall((:cfunc2, lib), Cint, (Ptr{MPI.MPI_Comm},), comm)
 ```
 
 ## Object finalizers and `MPI.Finalize`
--- a/docs/src/external.md
+++ b/docs/src/external.md
@@ -8,7 +8,7 @@ You need to ensure that external libraries are built correctly. In particular, i
 
 ## Passing MPI handles via `ccall`
 
-When passing MPI.jl handle objects ([`MPI.Comm`](@ref), [`MPI.Info`](@ref), etc) to C/C++ functions [via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/), you should pass the object directly as an argument, and specify the argument type as either the underlying handle type (`MPI.MPI_Comm`, `MPI.API.MPI_Info`, etc.), or a pointer (`Ptr{MPI.MPI_Comm}`, `Ptr{MPI.MPI_Info}`, etc.). This will internally handle the unwrapping, but ensure that a reference is kept to avoid premature garbage collection.
+When passing MPI.jl handle objects ([`MPI.Comm`](@ref), [`MPI.Info`](@ref), etc) to C/C++ functions [via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/), you should pass the object directly as an argument, and specify the argument type as either the underlying handle type (`MPI.MPI_Comm`, `MPI.MPI_Info`, etc.), or a pointer (`Ptr{MPI.MPI_Comm}`, `Ptr{MPI.MPI_Info}`, etc.). This will internally handle the unwrapping, but ensure that a reference is kept to avoid premature garbage collection.
 
 For example the C function signatures
 ```C
--- a/docs/src/external.md
+++ b/docs/src/external.md
@@ -45,3 +45,18 @@ There are two typical solutions to this problem:
    end
    MPI.add_finalize_hook!(() -> finalize(obj))
    ```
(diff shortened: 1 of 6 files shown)


docs/src/reference/environment.md
```diff
@@ -23,6 +23,8 @@ MPI.Is_thread_main
 MPI.Initialized
 MPI.Finalize
 MPI.Finalized
+MPI.add_init_hook!
+MPI.add_finalizer_hook!
```
giordano, line 27:
```suggestion
MPI.add_finalize_hook!
```

docs/src/external.md
```diff
@@ -0,0 +1,47 @@
...
+ccall((:cfunc2, lib), Cint, (Ptr{MPI.API.MPI_Comm},), comm)
+```
+
+## Object finalizers and `MPI.Finalize`
+
+External libraries may maintain allocate their own MPI handles (e.g.  create or duplicate MPI communictors), which need to be cleaned up. If these are attached to [object finalizers](https://docs.julialang.org/en/v1/base/base/#Base.finalizer), they may not be guaranteed to be called before `MPI.Finalize`, which can result in an error upon program exit. (By default, MPI.jl will install an [`atexit`](https://docs.julialang.org/en/v1/base/base/#Base.atexit) hook that calls `MPI.Finalize` if it hasn't already been invoked.)
+
+There are two typical solutions to this problem:
+
+1. Gate the clean up functions behind an [`MPI.Finalized](@ref) call, e.g.
```
giordano, line 30:
```suggestion
1. Gate the clean up functions behind an [`MPI.Finalized`](@ref) call, e.g.
```

giordano:
https://github.com/JuliaParallel/MPI.jl/actions/runs/4039207722/jobs/6943733193 (macOS + julia v1.6 + OpenMPI_jll):
```
A system call failed during shared memory initialization that should
not have.  It is likely that your MPI job will now either abort or
experience performance degradation.

  Local host:  Mac-1675033451153.local
  System call: unlink(2) /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T//ompi.Mac-1675033451153.501/pid.4601/1/vader_segment.Mac-1675033451153.501.76970001.2
  Error:       No such file or directory (errno 2)
```
Is this a known problem?

vchuravy:
Haven't seen this before

docs/src/external.md
```diff
@@ -0,0 +1,47 @@
+# External libraries and packages
+
+Other libraries and packages may also make use of MPI. There are several concerns to ensure things are set up correctly.
+
+## Binary requirements
+
+You will need to ensure that the binaries are built correctly. In particular, if you are using [using a system-provided MPI backend](@ref using_system_mpi), you will also need to use the system provided binary for the package in question.
```
sloede, line 7:
```suggestion
You need to ensure that external libraries are built correctly. In particular, if you are [using a system-provided MPI backend](@ref using_system_mpi) in Julia, you also need to use the *same* system-provided binary for all packages and external libraries you use.
```
Clarify the "single ABI" idea?

docs/src/external.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Passing handles via `ccall`
```
sloede, line 9:
```suggestion
## Passing MPI handles via `ccall`
```
For clarification?

docs/src/external.md
```diff
@@ -0,0 +1,47 @@
...
+
+When passing MPI.jl handle objects ([`MPI.Comm`](@ref), [`MPI.Info`](@ref), etc) to C/C++ functions [via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/), you should pass the object directly as an argument, and specify the argument type as either the underlying handle type (`MPI.API.MPI_Comm`, `MPI.API.MPI_Info`, etc.), or a pointer (`Ptr{MPI.API.MPI_Comm}`, `Ptr{MPI.API.MPI_Info}`, etc.). This will internally handle the unwrapping, but ensure that a reference is kept to avoid premature garbage collection.
```
sloede, line 11:
```suggestion
When passing MPI.jl handle objects ([`MPI.Comm`](@ref), [`MPI.Info`](@ref), etc) to C/C++ functions [via `ccall`](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/), you should pass the object directly as an argument, and specify the argument type as either the underlying handle type (`MPI.MPI_Comm`, `MPI.API.MPI_Info`, etc.), or a pointer (`Ptr{MPI.MPI_Comm}`, `Ptr{MPI.MPI_Info}`, etc.). This will internally handle the unwrapping, but ensure that a reference is kept to avoid premature garbage collection.
```
In my experience it works well without the `API` part. If we require the `API` part for sanity/future compatibility, I think we should state it somewhere prominently which parts of MPI.jl are expected to be called via `API` and which are just `MPI.xxx`. If there's no hard requirement yet, I'd prefer to not use the `API` part to avoid confusion. If it's just that some commonly used types are used via `MPI.xxx` and other, less commonly used or low-level entities are available only via `MPI.API.yyy` it should be documented as well.

docs/src/external.md
```diff
@@ -0,0 +1,47 @@
...
+
+For example the C function signatures
+```C
+int cfunc1(MPI_Comm comm);
+int cfunc2(MPI_Comm * comm);
+```
+would be called as
+```julia
+ccall((:cfunc1, lib), Cint, (MPI.API.MPI_Comm,), comm)
+ccall((:cfunc2, lib), Cint, (Ptr{MPI.API.MPI_Comm},), comm)
```
sloede, line 21:
```suggestion
ccall((:cfunc1, lib), Cint, (MPI.MPI_Comm,), comm)
ccall((:cfunc2, lib), Cint, (Ptr{MPI.MPI_Comm},), comm)
```

docs/src/external.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+## Object finalizers and `MPI.Finalize`
+
+External libraries may maintain allocate their own MPI handles (e.g.  create or duplicate MPI communictors), which need to be cleaned up. If these are attached to [object finalizers](https://docs.julialang.org/en/v1/base/base/#Base.finalizer), they may not be guaranteed to be called before `MPI.Finalize`, which can result in an error upon program exit. (By default, MPI.jl will install an [`atexit`](https://docs.julialang.org/en/v1/base/base/#Base.atexit) hook that calls `MPI.Finalize` if it hasn't already been invoked.)
```
sloede, line 26:
```suggestion
External libraries may allocate their own MPI handles (e.g., create or duplicate MPI communictors), which need to be cleaned up before MPI is finalized. If these are attached to [object finalizers](https://docs.julialang.org/en/v1/base/base/#Base.finalizer), they may not be guaranteed to be called before `MPI.Finalize`, which can result in an error upon program exit. (By default, MPI.jl will install an [`atexit`](https://docs.julialang.org/en/v1/base/base/#Base.atexit) hook that calls `MPI.Finalize` if it hasn't already been invoked.)
```
Sentence structure; some suggestion for clarification

sloede (commented):
This is a great improvement to the docs, thanks a lot! I left a few notes and suggestions, I hope that's ok.

docs/src/external.md
simonbyrne: @vchuravy @giordano what do you think?

giordano: Probably I agree it's nice to keep `MPI_Comm`  part of the "high-level" API (i.e. within `MPI`, not `MPI.API`).

giordano:
CC @fredrikekre (probably interested because of this https://github.com/fredrikekre/HYPRE.jl/pull/8).

docs/src/external.md
```diff
@@ -0,0 +1,62 @@
...
+External libraries may allocate their own MPI handles (e.g., create or duplicate MPI communictors), which need to be cleaned up before MPI is finalized. If these are attached to [object finalizers](https://docs.julialang.org/en/v1/base/base/#Base.finalizer), they may not be guaranteed to be called before `MPI.Finalize`, which can result in an error upon program exit. (By default, MPI.jl will install an [`atexit`](https://docs.julialang.org/en/v1/base/base/#Base.atexit) hook that calls `MPI.Finalize` if it hasn't already been invoked.)
+
+There are two typical solutions to this problem:
+
+1. Gate the clean up functions behind an [`MPI.Finalized`](@ref) call, e.g.
+
+   ```julia
+   finalizer(obj) do obj
+       if !MPI.Finalized
+           # call clean up funciton
```
giordano, line 35:
```suggestion
           # call clean up function
```

docs/src/external.md
```diff
@@ -0,0 +1,62 @@
...
+       end
+   end
+   ```
+
+2. Keep track of all such objects, clean them up via [`MPI.add_finalize_hook!`](@ref), e.g.
+
+   ```julia
+   finalizer(obj) do obj
+       # call clean up funciton
```
giordano, line 44:
```suggestion
       # call clean up function
```

docs/src/external.md
```diff
@@ -0,0 +1,62 @@
...
+   const REFS = WeakKeyDict{ObjType, Nothing}()
+   MPI.add_finalize_hook!() do
+       for obj in keys(REFS)
+           finalize(obj)
+       end
+   end
+
+   # for each object `obj`
+   finalizer(obj) do obj
+       # call clean up funciton
```
giordano, line 59:
```suggestion
       # call clean up function
```

PR #716, opened 2023-02-26
Add GTL

JBlaschke:
Yay!!! GPU-Aware MPI breaks more ABIs. Here's an example what happens without loading GTL before a `libmpi` that needs it:
```
MPICH ERROR [Rank 0] [job id 5680267.11] [Wed Feb 22 23:01:58 2023] [nid002845] - Abort(-1) (rank 0 in comm 0): MPIDI_CRAY_init: GPU_SUPPORT_ENABLED is requested, but GTL library is not linked
 (Other MPI error)

aborting job:
MPIDI_CRAY_init: GPU_SUPPORT_ENABLED is requested, but GTL library is not linked
```
(sorta makes sense I guess :raised_eyebrow: ... vendors don't want to have to compile two different libmpis ... just insert a libgtl whenever GPUs are around ... yes, that's wayyyy better :stuck_out_tongue: )

In the case of some system MPI libraries, GPU-aware MPI is implemented as another library -- bearing the fancy name of GPU Transport Layer (GTL).  For example, on Perlmutter it's called `libmpi_gtl_cuda.so`. Often it's important that this library is loaded _before_ libmpi. These changes do:

0. Don't change the default behavior 
1. `MPIPreferences` has an option: `gtl_names`, which -- if not  `nothing` -- is a list of possible names for the GTL library
2. `MPI` will dlopen `libgtl` before `libmpi` (if not `nothing`).

I have tested this on Perlmutter. Will test on Crusher next. Also I don't know if I accidentally broke MPITrampoline, which I will do asap.

This PR represents a tradeoff. Clearly there is no standard way that GTL is defined. So I avoided creating a default search strategy. One could be tempted to look for Cray systems and then "just load GTL". This would cause problems on our CPU nodes, which have the GTL libraries installed (we want to have a single SW image for all nodes), but don't support it (what GPUs? this is a CPU node!). 

This PR allows us (the helpful sysadmins) to provide two different `Preferences.toml` files for each type of node. It does come at the cost of forcing users to potentially having to manage different `LocalPreferences.toml` files if they have MPI in the `LocalPreferences`.


sloede:
> It does come at the cost of forcing users to potentially having to manage different `LocalPreferences.toml` files if they have MPI in the `LocalPreferences`.

What would they have to do right now to make it work? Or asked differently: Is this changing things from "inconvenient" to "even more inconvenient" for users on systems without sysadmin support, or from "impossible" to "doable but inconvenient"? If it is the latter, I think it will be an improvement nonetheless, wouldn't it?

simonbyrne:
This is quite a cumbersome patch just to deal with Cray MPI, I wish there was a better way. What does mpi4py do?

simonbyrne:
Since it's only required at runtime, we could just do it based on environment.


> This would cause problems on our CPU nodes, which have the GTL libraries installed (we want to have a single SW image for all nodes), but don't support it (what GPUs? this is a CPU node!).

How is this logic handled for C programs?

JBlaschke:
@sloede 
> What would they have to do right now to make it work? Or asked differently: Is this changing things from "inconvenient" to "even more inconvenient" for users on systems without sysadmin support, or from "impossible" to "doable but inconvenient"? If it is the latter, I think it will be an improvement nonetheless, wouldn't it?

Right now, they would either have to use:
```
LD_PRELOAD=${CRAY_MPICH_ROOTDIR}/gtl/lib/libmpi_gtl_cuda.so
```
or add 
```
Libc.Libdl.dlopen("libmpi_gtl_cuda.so", Libc.Libdl.RTLD_GLOBAL)
```
**before** the first `using MPI` line. 

Re advice for users who can't ask the sysadmin, I would document an example `use_system_binary` call -- they would have to do that anyway. So something like:
```
MPIPreferences.use_system_binary(; export_prefs=true, library_names=["libmpi_cray"], gtl_names=["libmpi_gtl_cuda", "libmpi_gtl_hsa"], mpiexec="srun")
```
would cover both Perlmutter and Frontier.

JBlaschke:
> This is quite a cumbersome patch just to deal with Cray MPI, I wish there was a better way. What does mpi4py do?

They do the same thing that Cray always tells us to do: "use the compiler wrappers to build mpi4py".

JBlaschke:
@simonbyrne 
> Since it's only required at runtime, we could just do it based on environment variables?

Urgh ... If it where up to me alone, then sure! Let's put in an env variable. But I kinda like the idea of having preferences managed by ... well ... Preferences (with a capital "P"). Anyway, GTL is part of using the system binary, so I think keeping this alongside libmpi makes sense.

> How is this logic handled for C programs?

They are compiled using the Cray compiler wrappers -- I don't know how they work (please don't make me open that pandora's box again :vomiting_face: ). I suspect they futz around with the linker to make sure that GTL is linked before MPI. Note: when you build a program with GTL enabled, it can't run on CPU nodes. So the compiler wrappers do insert something....

simonbyrne:
> Note: when you build a program with GTL enabled, it can't run on CPU nodes.

Ah, that's disappointing. I was hoping there was some magic environment variable set on your GPU nodes that we could rely upon. Ah well.

> This PR allows us (the helpful sysadmins) to provide two different Preferences.toml files for each type of node.

Why do you need two Preferences.toml files for each type of node?

simonbyrne:
I need to go to bed, but I can get on board with this if we make it a little less Cray-specific: what if we just called it `preload`, and it could be a list of libraries that get dlopen-ed before `libmpi`?

JBlaschke:
> Ah, that's disappointing. I was hoping there was some magic environment variable set on your GPU nodes that we could rely upon. Ah well.

That's what I was hoping for also...

> Why do you need two Preferences.toml files for each type of node?

One with the preloads and one without.

JBlaschke:
@simonbyrne 

> I need to go to bed, but I can get on board with this if we make it a little less Cray-specific: what if we just called it `preload`, and it could be a list of libraries that get dlopen-ed before `libmpi`?

I like this! It would be a bit more effort, but would cover a broader set of use cases. If we can define preloads that depend on an env var (`GPU_SUPPORT_ENABLED`) then we might be able to get away with just one set of preferences. (That env var is turned off on the CPU nodes)

vchuravy:
> we can define preloads that depend on an env var `MPICH_GPU_SUPPORT_ENABLED`

So I am wondering if we should do something:
1. Add a vendor flag to MPIPreferences and try to autodetect `cray`
2. When vendor is `cray` check the environment variable `MPICH_GPU_SUPPORT_ENABLED` 
    and attempt a `Libc.Libdl.dlopen("libmpi_gtl_cuda.so", Libc.Libdl.RTLD_GLOBAL)`?
    
Now the big question for me is how to deal with rocm vs cuda... and do we need something like a `LD_LIBRARY_PATH` (I would expect the module file to set that correctly)?

JBlaschke:
Vendor flags might make a lot of sense for a different reason @vchuravy : as it stands right now, if a user loads say `PrgEnv-nvidia` then they have to load `libmpi_nvidia`, not `libmpi_cray` and so on. Sometimes these are compatible (e.g. `libmpi_cray` seems to work with `PrgEnv-gnu`), but not always.

I build different Julia modules for each PE, but if a user rolls their own Julia environment then it might rely on a specific PE. Having smarter logic in MPI.jl would fix that.

> Now the big question for me is how to deal with rocm vs cuda

The rocm version of gtl is called `libmpi_gtl_hsa`.  My first attempt would be to try to detect either using `find_library`. I noticed that Perlmutter has both `libmpi_gtl_cuda.so` and `libmpi_gtl_hsa.so`, but `find_library` only detects the cuda version. Maybe on an AMD machine it's the other way around.

giordano:
How is libgtl related to libmpi? The latter requires something (symbols?) from the former? If so, does libmpi dynamically link to libgtl (i.e. what's the output of `readelf -d /path/to/libmpi`)? If the answer to this last question is no, then something looks broken to me in how this is packaged up, but perhaps I'm missing something.

vchuravy:
@giordano my assumption is that they use dlsym to see if the library is preloaded/linked into the binary.

Right now I just want to shout at cray and have everyone use OpenMPI.

JBlaschke:
Yea, there are no symbols the libmpi needs from `libmpi_gtl_*` that I can see -- otherwise we'd be getting linker errors, or Cray's own error.

What @vchuravy says makes sense. I can't find any documentation on this (other than: if you see this error, recompile with the cray compiler wrappers). 

Following this up at NERSC, to see if Cray would be willing to change the behavior of Cray MPICH. We should still  work on vendor flags in the meantime.

simonbyrne:
> > Why do you need two Preferences.toml files for each type of node?
> 
> 
> 
> One with the preloads and one without.

I don't understand: wouldn't you always want the preloads for the GPU nodes, and no preloads on non-GPU?

JBlaschke:
> > > Why do you need two Preferences.toml files for each type of node?
> 
> I don't understand: wouldn't you always want the preloads for the GPU nodes, and no preloads on non-GPU?

You mean CPU? Mainly for sanity: I don't know what GTL will do on a system without GPUs ....

simonbyrne:
Are you able to join the JuliaHPC meeting on Tuesday? It might be easier to discuss there.

JBlaschke:
> Right now I just want to shout at cray and have everyone use OpenMPI.

@vchuravy in 20 years we'll have compatible ABIs https://www.mpich.org/abi/

JBlaschke:
> Are you able to join the JuliaHPC meeting on Tuesday? It might be easier to discuss there.

Sadly no. I can do an impromptu call at 9am PT tomorrow (Monday)

simonbyrne:
What if we were to load the GTL if `MPICH_GPU_SUPPORT_ENABLED=1` is set? Can we simply assume that it is in the same directory as the libmpi?

vchuravy:
What is `nm -D` on these libraries?

JBlaschke:
> What is `nm -D` on these libraries?

@vchuravy Here you go:
* [nm_libmpi_gtl_cuda.txt](https://github.com/JuliaParallel/MPI.jl/files/10844684/nm_libmpi_gtl_cuda.txt)
* [nm_libmpi_cray.txt](https://github.com/JuliaParallel/MPI.jl/files/10844685/nm_libmpi_cray.txt)
* [nm_libmpi_gnu_91.txt](https://github.com/JuliaParallel/MPI.jl/files/10844687/nm_libmpi_gnu_91.txt)
* [nm_libmpi_nivida.txt](https://github.com/JuliaParallel/MPI.jl/files/10844688/nm_libmpi_nivida.txt)

vchuravy:
So now the question is what is exported on Frontier/Crusher. My worry is that the symbols overlap and it would only be legal to preload one of them

JBlaschke:
I'm currently traveling, so will send you these for Crusher (sans Nvidia) tomorrow. I have a feeling you're right. Funnily enough, even thought `libmpi_gtl_hsa.so` is in my `LD_LIBRARY_PATH`, `find_library("libmpi_gtl_hsa")` returns an empty string on Perlmutter.

luraess:
> So now the question is what is exported on Frontier/Crusher

If interested to see how things are on LUMI, let me know what should be queried (and tested) there.

vchuravy:
I am after `nm -D libmpi_gtl_hsa.so`

luraess:
> nm -D libmpi_gtl_hsa.so

On LUMI compute node with modules loaded:
```
lurass@nid005378:/scratch/project> nm -D libmpi_gtl_hsa.so
nm: 'libmpi_gtl_hsa.so': No such file
```

giordano:
`libmpi_gtl_hsa.so` has to be the path to where the file is 🙂

luraess:
Thanks @giordano for the precision

Output of `nm -D libmpi_gtl_hsa.so` on LUMI: [out.txt](https://github.com/JuliaParallel/MPI.jl/files/10848876/out.txt)

giordano:
```
T mpix_gtl_check_peer_access
T mpix_gtl_close_mem_handle
T mpix_gtl_enable_peer_access
T mpix_gtl_finalize
T mpix_gtl_free
T mpix_gtl_free_host
T mpix_gtl_get_ctx
T mpix_gtl_get_memcpy_kind
T mpix_gtl_get_mem_handle
T mpix_gtl_get_my_numa_node
T mpix_gtl_gpu_arch_query
T mpix_gtl_gpu_mem_allow_access
T mpix_gtl_host_register
T mpix_gtl_host_unregister
T mpix_gtl_init
T mpix_gtl_malloc
T mpix_gtl_malloc_host
T mpix_gtl_map_gpu_buf
T mpix_gtl_memcpy_async
T mpix_gtl_memcpy_blk
T mpix_gtl_mem_host_get_device_pointer
T mpix_gtl_mem_host_register
T mpix_gtl_mem_host_unregister
T mpix_gtl_nt_memcpy
T mpix_gtl_open_mem_handle
T mpix_gtl_pointer_type
T mpix_gtl_query_request
T mpix_gtl_reduce_execute
T mpix_gtl_reduce_prepare
T mpix_gtl_reduce_trigger
T mpix_gtl_stream_synchronize
T mpix_gtl_stream_wait32
T mpix_gtl_stream_write32
T mpix_gtl_verify_driver_version
```
seem to be defined in both of them.

simonbyrne:
Following discussion on HPC call, I think we should do the following:

1. Add "implementation" (or "vendor"?) key to MPIPreferences.jl. This should be one of those in https://github.com/JuliaParallel/MPI.jl/blob/master/lib/MPIPreferences/src/MPIPreferences.jl#L218-L324

2. Add a implementation-specific key-value section to MPIPreferences.jl (not sure how it should be named). e.g.

i.e. we will have something like:
```toml
implementation = "CrayMPICH"
implementation_options.libgtl = "libmpi_gtl_cuda.so"
```

3. In the `MPIPreferences.System.__init__()` we change this line
https://github.com/JuliaParallel/MPI.jl/blob/817caa70bbe443f29da47632cc1b85694857299d/lib/MPIPreferences/src/system.jl#L12
to

```julia
if @load_preference("implementation") == "CrayMPICH"
   libgtl = @load_preference("implementation_options.libgtl", nothing)
   if !isnothing(libgtl) && get(ENV, "MPICH_GPU_SUPPORT_ENABLED", "0") == "1"
       Libdl.dlopen(libgtl)
   end
end

# now dlopen libmpi
Libdl.dlopen(libmpi, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
```

src/MPI.jl
```diff
@@ -103,6 +103,11 @@ function __init__()
     end
 
     @static if Sys.isunix()
+        # dlopen GTL library (if needed) before dlopen'ing the MPI library:
+        # https://github.com/JuliaParallel/MPI.jl/pull/716
+        if ! isnothing(libgtl)
+            Libdl.dlopen(libgtl, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
+        end
```
simonbyrne, line 110: Hmm: is this redundant with the `lib/MPIPreferences/src/system.jl`?

JBlaschke: I've changed `libgtl` -> a vector or libraries to preload.  `Libdl.dlopen` also here in case any of the other jll's start to fall into that pattern -- MPItrampoline comes to mind. If the MPItrampoline provider is built against a system MPI that needs preloads, then they wouldn't be loaded in `MPIPreferences.Sytem`. It's possible that some clever linking would prevent this -- but I haven't had time to figure that out. In the meantime, this is me being cautious.

JBlaschke:
@simonbyrne take a look now -- I've deviated from your recipe a bit, in the name of simplicity.

1. I am using a parser (following the urging of @vchuravy )  which interprets the `cc --cray-print-opts` to find the name of i) libmpi (yea... that changes based on the PrgEnv); ii) libgtl (depending on AMD vs NVIDIA); iii) any other libraries that the cray compiler wrappers link against (right now i've kept them -- `cclibs` for debugging purposes).

2. The preloads are put into a vector `preloads` which will be traversed only if the environment variable named in `preloads_env_switch` has the value `1`.

From the user's perspective, this new behavior can be triggered by setting `vendor="cray"` -- eg:
```
MPIPreferences.use_system_binary(; vendor="cray", mpiexec="srun")
```

JBlaschke:
@simonbyrne I'm not sure why some of these CI tests failed -- Is there some way to get more info?

lib/MPIPreferences/src/parse_cray_cc.jl
```diff
@@ -0,0 +1,94 @@
...
+
+function other_libs(libs)
+    x = libs |>
+        filter(x->!startswith(x, libmpi_prefix)) |> 
+        filter(x->!startswith(x, libgtl_prefix))
+    return x
+end
+
+function analyze_cray_cc()
+    cray_opts = communicate(Cmd(["cc", "--cray-print-opts=all"]))
```
simonbyrne, line 73: @JBlaschke Should be able to `read` directly?

lib/MPIPreferences/src/parse_cray_cc.jl
```diff
@@ -0,0 +1,94 @@
+module CrayParser
+
+import Base: filter, map, reduce
+filter(f::Function)::Function = x -> filter(f, x)
+map(f::Function)::Function    = x -> map(f, x)
+reduce(f::Function)::Function = x -> reduce(f, x)
```
giordano, line 6: I think it's better to use `Base.Fix1(filter)` and such where you need to use these functions instead of adding methods to Base functions.

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -160,23 +180,28 @@ function use_system_binary(;
...
+
     set_preferences!(MPIPreferences,
         "_format" => "1.0",
         "binary" => binary,
         "libmpi" => libmpi,
         "abi" => abi,
         "mpiexec" => mpiexec,
+        "preloads" => preloads,
+        "preloads_env_switch" => preloads_env_switch,
+        "cclibs" => cclibs;
```
simonbyrne, line 200: Can we document all these (what are valid values, etc)

lib/MPIPreferences/src/parse_cray_cc.jl
```diff
@@ -0,0 +1,94 @@
...
+    )
+end
+
+const libmpi_prefix = "mpi_"
+const libgtl_prefix = "mpi_gtl_"
+
+function cray_mpi(libs)
+    x = libs |>
+        filter(x-> startswith(x, libmpi_prefix)) |> 
+        filter(x->!startswith(x, libgtl_prefix))
```
giordano, line 52: For example this can be simplified to
```suggestion
        Base.Fix1(filter)(startswith(libmpi_prefix)) |> 
        Base.Fix1(filter)(!startswith(libgtl_prefix))
```

lib/MPIPreferences/src/parse_cray_cc.jl
```diff
@@ -0,0 +1,94 @@
...
+    @assert length(x) == 1
+    return x[1]
```
giordano, line 54:
```suggestion
    return only(x)
```

JBlaschke:
Note to self: test on Frontier or Crusher

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -160,23 +180,28 @@ function use_system_binary(;
...
     if isnothing(abi)
         abi = identify_abi(libmpi)
     end
+
     if mpiexec isa Cmd
         mpiexec = collect(mpiexec)
     end
+
     set_preferences!(MPIPreferences,
         "_format" => "1.0",
```
simonbyrne, line 193: I think what we should do is set this to `"1.1"` if any of the new variables are set, but keep it as 1.0 if it is backward compatible (and document it).

simonbyrne: will also need to update the check at the top of the file.

lib/MPIPreferences/src/parse_cray_cc.jl
simonbyrne: `readchomp(::Cmd)`

lib/MPIPreferences/src/parse_cray_cc.jl
JBlaschke: makes sense -- done

lib/MPIPreferences/src/parse_cray_cc.jl
JBlaschke: yes, this is neater -- done

lib/MPIPreferences/src/MPIPreferences.jl
JBlaschke: done

lib/MPIPreferences/src/parse_cray_cc.jl
JBlaschke: I've just defined `filter` etc at the top of the file -- without importing Base.filter, etc first -- I think it makes the code more readable

lib/MPIPreferences/src/parse_cray_cc.jl
JBlaschke: yes, that's nicer -- done

lib/MPIPreferences/src/MPIPreferences.jl
JBlaschke: done

JBlaschke:
After a hiatus involving far too many word documents, I can finally return to coding. I've made the changes you suggested @simonbyrne and @giordano  -- let me know what you think.

I am also watching CI, and I might need some help understanding the following errors:
1. Buildkite is showing curl errors for UCX -- do I ignore those? 
2. docs-build can't find the 'configure_jll_binary' reference in src/configuration.md. I checked and the reference still exists. But that's right after my new additions, so I probably messed up.  If someone can tell me how, that would be much appreciated.

JBlaschke:
Ah! looks like cleaning up formatting seems to have solved the docs-build problem

JBlaschke:
Can someone familiar with CI comment on what I should do with the failing tests. Right now I don't understand how and if my changes triggered these regressions

JBlaschke:
@simonbyrne any chance you can merge this?

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -77,11 +78,14 @@ function use_jll_binary(binary = Sys.iswindows() ? "MicrosoftMPI_jll" : "MPICH_j
               Accepted options are:
                   $(join(known_binaries, ", "))""")
     set_preferences!(MPIPreferences,
-        "_format" => "1.0",
+        "_format" => "1.1",
```
simonbyrne, line 81:
```suggestion
        "_format" => "1.0",
```
since it should be backward compatible with older MPIPreferences

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -160,23 +185,28 @@ function use_system_binary(;
...
         abi = identify_abi(libmpi)
     end
+
     if mpiexec isa Cmd
         mpiexec = collect(mpiexec)
     end
+
     set_preferences!(MPIPreferences,
-        "_format" => "1.0",
+        "_format" => "1.1",
```
simonbyrne, line 198:
```suggestion
        "_format" => isnothing(vendor) ? "1.0" : "1.1",
```

docs/src/configuration.md
```diff
@@ -107,6 +106,48 @@ Preferences are merged across the Julia load path, such that it is feasible to p
...
+ * `cclibs` is a list of libraries also linked by the compiler wrappers. This is
+   recorded mainly for debugging purposes, and the libraries listed here are not
+   explicitly loaded by `MPI.jl`.
+
+An example of running `MPIPreferences.use_system_library(vendor="cray")` in
+`PrgEnv-gnu` is:
+
+```toml
+[MPIPreferences]
+_format = "1.0"
```
simonbyrne, line 137:
```suggestion
_format = "1.1"
```

docs/src/configuration.md
```diff
@@ -107,6 +106,48 @@ Preferences are merged across the Julia load path, such that it is feasible to p
...
+
+The function of these settings are as follows:
+ * `preloads` specifies a list of libraries that are to be loaded (in order)
+   before `libmpi`.
+ * `preloads_env_switch` specifies the name of an environment variable that, if
+   set to `0`, can disable the `preloads`
+ * `cclibs` is a list of libraries also linked by the compiler wrappers. This is
+   recorded mainly for debugging purposes, and the libraries listed here are not
+   explicitly loaded by `MPI.jl`.
+
```
simonbyrne, line 131:
```suggestion

If these are set, the `_format` key will be set to `"1.1"`.

```

simonbyrne (changes requested):
To try to keep backward compatibility, we should only update the `_format` key if it uses the new features. Otherwise, we can keep it at `"1.0"`.

lib/MPIPreferences/src/MPIPreferences.jl
JBlaschke: Oh, so later versions can read earlier formats? Nice!

simonbyrne (approved):

src/MPI.jl
```diff
@@ -113,6 +113,20 @@ function __init__()
...
+        if isnothing(preloads_env_switch)
+            preload_enabled = true
+        elseif get(ENV, preloads_env_switch, "0") == "1"
+            preload_enabled = true
+        end
+        if preload_enabled
+            for preload in preloads
+                Libdl.dlopen(preload, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
+            end
+        end
```
simonbyrne, line 128: this is where MPItrampoline is erroring: it ends up in a state `preload_enabled == true`, but `preloads == nothing`

simonbyrne: How about:
```suggestion
        if !isnothing(preloads)
            if isnothing(preloads_env_switch) || get(ENV, preloads_env_switch, "0") == "1"
                for preload in preloads
                    Libdl.dlopen(preload, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
                end
            end
        end
```

simonbyrne: Actually, this should be redundant altogether, since it should have already been triggered by `MPIPreferences.System.__init__()`

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -114,6 +118,7 @@ end
         mpiexec = "mpiexec",
         abi = nothing,
         export_prefs = false,
+        gtl_names=nothing,
```
vchuravy, line 121: ?vendor

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -145,10 +157,23 @@ function use_system_binary(;
...
+    preloads = []
+    preloads_env_switch = nothing
+    cclibs = []
+    if vendor == "cray"
+        cray_pe = CrayParser.analyze_cray_cc()
+        library_names = [cray_pe.libmpi]
+        preloads = [cray_pe.libgtl]
+        preloads_env_switch = cray_pe.gtl_env_switch
+        cclibs = cray_pe.cclibs
+    end
```
vchuravy, line 175:
```suggestion
    if vendor === nothing
    elseif vendor == "cray"
        cray_pe = CrayParser.analyze_cray_cc()
        library_names = [cray_pe.libmpi]
        preloads = [cray_pe.libgtl]
        preloads_env_switch = cray_pe.gtl_env_switch
        cclibs = cray_pe.cclibs
    else
        error("Unknown vendorr $vendor")
    end
```

vchuravy (approved):

vchuravy:
LGTM!

lib/MPIPreferences/src/MPIPreferences.jl
```diff
@@ -4,7 +4,7 @@ export use_jll_binary, use_system_binary
 
 using Preferences, Libdl
 
-if !(VersionNumber(@load_preference("_format", "1.0")) <= v"1.0")
```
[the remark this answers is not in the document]
JBlaschke: I think the current solution works now

JBlaschke:
Ok, so I've cleaned things up a bit. I moved all the preload logic to `MPIPreferences`. This way, if the vendor logic changes, then we only need to bump `MPIPreferences`. It also avoids code duplication. I tested on Perlmutter. Things are working nicely.

@simonbyrne @vchuravy feel free to commit.

simonbyrne:
LGTM

Just need to add the docstring to the docs:
https://github.com/JuliaParallel/MPI.jl/actions/runs/5616415928/job/15218697196?pr=716#step:4:21

JBlaschke:
@simonbyrne Docstring added

src/MPI.jl
```diff
@@ -112,6 +112,10 @@ function __init__()
         """ ENV["JULIA_MPI_BINARY"]=mpi_env_binary MPIPreferences.binary
     end
 
+    # preload any dependencies of libmpi (if needed, eg. GTL on cray) before
+    # dlopen'ing the MPI library: https://github.com/JuliaParallel/MPI.jl/pull/716
+    MPIPreferences.dlopen_preloads()
```
simonbyrne, line 117:
```suggestion
    if isdefined(MPIPreferences, :dlopen_preloads)
        # preload any dependencies of libmpi (if needed, eg. GTL on cray) before
        # dlopen'ing the MPI library: https://github.com/JuliaParallel/MPI.jl/pull/716
        MPIPreferences.dlopen_preloads()
    end
```

simonbyrne:
Can you bump the patch version of MPIPreferences?

JBlaschke:
This looks good -- @simonbyrne do you also want to bump the MPI.jl patch version?

PR #723, opened 2023-03-10
Test GC in multi-thread code

alexandrebouchard:


alexandrebouchard:
@miguelbiron FYI: trying to reproduce the bug using a slight modif of MPI.jl's test cases

PR #724, opened 2023-03-10
Test GC in multi-thread code

alexandrebouchard:
Branched right off Tag 0.20.8 in which all tests were passing.


alexandrebouchard:
@miguelbiron FYI: create a new one, this time branching off the latest point where all tests were passing

test/test_threads.jl
```diff
@@ -24,6 +24,10 @@ if provided == MPI.THREAD_MULTIPLE
     Threads.@threads for i = 1:N
         reqs[N+i] = MPI.Irecv!(@view(recv_arr[i:i]), comm; source=src, tag=i)
         reqs[i] = MPI.Isend(@view(send_arr[i:i]), comm; dest=dst, tag=i)
+        if i == 1 
+            GC.gc()
```
simonbyrne, line 28:
```suggestion
            # test that multithreaded GC doesn't trigger segfaults
            GC.gc()
```

simonbyrne:
We should probably merge this

PR #728, opened 2023-04-26
improve loadtime hooks

ranocha:
I reduced the amount of compilation burden of the loadtime hooks. With Julia v1.90-rc1 and the current `master` branch of the MPI.jl repository, I get

```
julia --project=. -e 'import Pkg; Pkg.precompile(); @time @eval using MPI'
  0.276089 seconds (196.80 k allocations: 13.368 MiB, 21.34% gc time, 84.87% compilation time)
```

With this PR, I get

```
julia --project=. -e 'import Pkg; Pkg.precompile(); @time @eval using MPI'
  0.089453 seconds (125.92 k allocations: 8.464 MiB, 53.86% compilation time)
```

This is roughly a factor of three improvement. Any ideas for further improvements are welcome!

There is still a lot of type inference going on when loading MPI.jl. This is caused by the type-unstable access of `const _mpi_load_time_hooks = Any[]`. `__init__()` seems to be precompiled, but this does not include dynamic dispatch from type-unstable code inside it (in accordance with discussions on Slack). Thus, it appears to make sense to compile this already at precompile time. However, doing so naively by calling `__init__()` or `run_load_time_hooks()` while precompiling violates `@assert !_finished_loading[]` in `run_load_time_hooks()` when called later from `__init__()` when loading MPI.jl. I am not sure there is a good workaround, but this PR yields already a nice first improvement for me.

Edit: I added some PrecompileTools setup in 9187fbb to reduce the loading time a bit more.


ranocha:
Does somebody have time to review this?

src/MPI.jl
```diff
@@ -145,4 +154,12 @@ function __init__()
     end
 end
 
+
+using PrecompileTools: @compile_workload
+@static if VERSION >= v"1.8"
+    @compile_workload begin
+        foreach(hook -> hook(), _mpi_load_time_hooks)
```
sloede, line 161:
```suggestion
        # Running the load time hooks here shaves off a significant amount of loading time,
        # see also https://github.com/JuliaParallel/MPI.jl/pull/728
        foreach(hook -> hook(), _mpi_load_time_hooks)
```

sloede (commented):
To me, this looks good and brings a nice load time improvement 👍 However, someone else with more insights into the initialization procedures should probably review it as well.

sloede (approved):

lcw:
I just tried the changes locally and everything worked great.  I saw similar speedups.  I suggest removing commits f91831faf59b23dab7319be345fdeddc2c3eaefc and 582e4f75d9680d997687fe3a7fae27dded12d14b.  I am not totally positive, but the test failures seem unrelated.

ranocha:
> I just tried the changes locally and everything worked great. I saw similar speedups. I suggest removing commits [f91831f](https://github.com/JuliaParallel/MPI.jl/commit/f91831faf59b23dab7319be345fdeddc2c3eaefc) and [582e4f7](https://github.com/JuliaParallel/MPI.jl/commit/582e4f75d9680d997687fe3a7fae27dded12d14b). I am not totally positive, but the test failures seem unrelated.

Thanks! I removed the commits as suggested

PR #735, opened 2023-05-29
some explicit precompile statements and precompile-friendly loadtime hooks

ranocha:
On my system, I get the following results.

Current `master`:
```
$ julia --project=. -e 'import Pkg; Pkg.precompile(); @time @eval using MPI'
Precompiling project...
  1 dependency successfully precompiled in 3 seconds. 13 already precompiled.
  0.068064 seconds (60.26 k allocations: 4.630 MiB, 51.31% compilation time)
```

This PR:
```
$ julia --project=. -e 'import Pkg; Pkg.precompile(); @time @eval using MPI'
Precompiling project...
  1 dependency successfully precompiled in 3 seconds. 13 already precompiled.
  0.038507 seconds (57.11 k allocations: 4.375 MiB, 6.25% compilation time)
```

### DIFF
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -167,9 +167,9 @@ using PrecompileTools: @compile_workload
     # are likely to be called from anyone using MPI in Julia. Thus, it is reasonable
     # to compile them ahead of time instead of compiling them in parallel for each
     # run.
-    Base.precompile(Tuple{typeof(Init)})
-    Base.precompile(Tuple{typeof(Comm_size), Comm})
-    Base.precompile(Tuple{typeof(Comm_size), Comm})
+    Base.precompile(Init, ())
+    Base.precompile(Comm_size, (Comm,))
+    Base.precompile(Comm_size, (Comm,))
 end
 
 end
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -53,7 +53,7 @@ end
 function run_load_time_hooks()
     @assert !_finished_loading[]
     _finished_loading[] = true
-    foreach(hook -> hook(), _mpi_load_time_hooks)
+    foreach(call, _mpi_load_time_hooks)
     empty!(_mpi_load_time_hooks)
     nothing
 end
@@ -66,6 +66,7 @@ function (hook::LoadTimeHookSetVal)()
     hook.dst.val = hook.src[]
     return nothing
 end
+call(hook) = hook()
 
 include("implementations.jl")
 include("error.jl")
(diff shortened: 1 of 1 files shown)


src/MPI.jl
```diff
@@ -160,8 +161,15 @@ using PrecompileTools: @compile_workload
...
         # Running the load time hooks here shaves off a significant amount of loading time,
         # see also https://github.com/JuliaParallel/MPI.jl/pull/728
-        foreach(hook -> hook(), _mpi_load_time_hooks)
+        foreach(call, _mpi_load_time_hooks)
     end
+    # We insert some explicit precompile statements here. The corresponding methods
+    # are likely to be called from anyone using MPI in Julia. Thus, it is reasonable
+    # to compile them ahead of time instead of compiling them in parallel for each
+    # run.
+    Base.precompile(Tuple{typeof(Init)})
```
giordano, line 170: Are you sure the argument is correct? Shouldn't it be 
```suggestion
    Base.precompile(Init, ())
```
and similarly below.

ranocha: That's the output I get from SnoopCompile.jl, e.g., `Base.precompile(Tuple{typeof(main)})` in https://timholy.github.io/SnoopCompile.jl/stable/snoopi_deep_parcel/#SnoopCompile.write

giordano: Well, that's not what documentation says: https://docs.julialang.org/en/v1/base/base/#Base.precompile

src/MPI.jl
```diff
@@ -160,8 +161,15 @@ using PrecompileTools: @compile_workload
...
+    Base.precompile(Tuple{typeof(Comm_size), Comm})
+    Base.precompile(Tuple{typeof(Comm_size), Comm})
```
ranocha, line 172:
```suggestion
    Base.precompile(Init, ())
    Base.precompile(Comm_size, (Comm,))
    Base.precompile(Comm_size, (Comm,))
```
Change as suggested by @giordano

src/MPI.jl
ranocha: Updated in c5393d51eeaaae507b8a30cc62b8b67340b73be6

sloede: Maybe the docs in SnoopCompile and/or the generated code are out of date? The method referenced by @giordano internally calls the variant used by @ranocha, afaict:
https://github.com/JuliaLang/julia/blob/8e630552924eac54c809aa7bc30871c7df1582d3/base/loading.jl#L2871

ranocha: That's something to be discussed at the SnoopCompile.jl repo. The current version of this PR should be fine here, isn't it?

ranocha:
Bump

ranocha:
Thanks for the review and for merging this, @giordano. It would be nice to get a new release of MPI.jl :slightly_smiling_face:

PR #738, opened 2023-06-13
add MPI.versioninfo function

simonbyrne:
This should make it easier to debug and help users.


giordano (approved):

Project.toml
```diff
@@ -12,6 +12,7 @@ MPIPreferences = "3da0fdf6-3ccc-4f1b-acd9-58baa6c99267"
 MPItrampoline_jll = "f1f71cc9-e9ae-5b93-9b94-4fe0e1ad3748"
 MicrosoftMPI_jll = "9237b28f-5490-5468-be7b-bb81f5f5e6cf"
 OpenMPI_jll = "fe0851c0-eecd-5654-98d4-656369965a5c"
+PkgVersion = "eebad327-c553-4316-9ea0-9fa01ccd7688"
```
vchuravy, line 15: Annoying that we need a new dependency for this

vchuravy (approved):

Project.toml
simonbyrne: Yeah probably should be part of Pkg. CUDA.jl implements its own logic, but honestly this seems easier.

PR #741, opened 2023-06-24
[CI] Run Intel MPI jobs on `ubuntu-latest`

giordano:
`ubuntu-18.04` isn't supported anymore in GitHub Actions, we aren't running Intel MPI CI jobs at all at the moment.

Opening as draft to see whether this works, because the comments say "20.04 not supported".

PR #742, opened 2023-06-24
Avoid segmentation faults in threaded code when using Intel MPI

### LINKED ISSUE #725: GC in a multithreaded MPI context causing segfaults beyond UCX
Thanks again for your help with https://github.com/JuliaParallel/MPI.jl/issues/720 - this one is unrelated (except that issue #720 lead us to create more comprehensive unit test revealing this new, probably unrelated segfault). 

**Summary of this problem:** a segfault occurs when GC is triggered in a multithreaded+MPI context. 

**How to reproduce:** I have create a draft PR adding a GC.gc() call in one of MPI.jl's existing multithreaded test: see PR Request https://github.com/JuliaParallel/MPI.jl/pull/724

The draft PR is based off the most recent commit where all tests passed (Tag 0.20.8). In the output of "test-intel-linux", the salient output is

```
signal (11): Segmentation fault
in expression starting at /home/runner/work/MPI.jl/MPI.jl/test/test_threads.jl:18
ijl_gc_enable at /cache/build/default-amdci4-2/julialang/julia-release-1-dot-8/src/gc.c:2955
```

The change we made is in the file test/test_threads.jl, where we added the following if clause:

```
    Threads.@threads for i = 1:N
        reqs[N+i] = MPI.Irecv!(@view(recv_arr[i:i]), comm; source=src, tag=i)
        reqs[i] = MPI.Isend(@view(send_arr[i:i]), comm; dest=dst, tag=i)
        if i == 1 
            GC.gc()
        end

    end 
```

We experience similar problems with MPICH 4.0 in our package (https://github.com/Julia-Tempering/Pigeons.jl), but not with MPICH 4.1. 

**Related discussions**

- https://juliaparallel.org/MPI.jl/stable/knownissues/#Multi-threading-and-signal-handling

This describes a similar issue in the context of UCX. However this problem does not seem limited to UCX from our investigations so far. 

- https://github.com/JuliaParallel/MPI.jl/issues/337

This describes a similar issue in the context of OpenMPI. However it seems that certain versions of MPICH and intel MPI (which is MPICH-derived) might suffer from a similar issue? 

In light of these two sources, perhaps other environment variables in the style of https://github.com/JuliaParallel/MPI.jl/blob/6d513bb37da76182f2c31baff44f1f53a103295d/src/MPI.jl#L133 could be set to address this issue? I was wondering if anyone might have some suggestion on whether that's a reasonable hypothesis? Having limited MPI experience I am not sure what these environment variables might be.

Thank you so much for your time.

giordano:
This is supposed to fix #725.  I haven't tested this locally, but now that Intel MPI CI is actually working (#741) and can reproduce the segmentation fault reported, we can see how this goes.

### DIFF
--- a/docs/src/knownissues.md
+++ b/docs/src/knownissues.md
@@ -94,6 +94,14 @@ ENV["UCX_ERROR_SIGNALS"] = "SIGILL,SIGBUS,SIGFPE"
 ```
 at `__init__`. If set externally, it should be modified to exclude `SIGSEGV` from the list.
 
+Similarly, Intel MPI would error if the above signal is raised ([#725](https://github.com/JuliaParallel/MPI.jl/issues/725)), resulting in a segmentation fault error like
+```
+[6950] signal (11.-6): Segmentation fault
+in expression starting at /home/runner/work/MPI.jl/MPI.jl/test/test_threads.jl:18
+ijl_gc_enable at /cache/build/default-amdci4-6/julialang/julia-release-1-dot-9/src/gc.c:3222
+```
+MPI.jl works around this problem by setting the environment variable `IPATH_NO_BACKTRACE` to `1`, unless already defined.
+
 ## CUDA-aware MPI
 
 ### Memory pool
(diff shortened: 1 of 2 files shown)


giordano:
This is [still crashing](https://github.com/JuliaParallel/MPI.jl/actions/runs/5365210592/jobs/9733887656?pr=742#step:12:378)
```
signal (11): Segmentation fault
in expression starting at /home/runner/work/MPI.jl/MPI.jl/test/test_threads.jl:18
signal (11): Segmentation fault
```
:cry:

PR #743, opened 2023-06-29
Bump AMDGPU compat

luraess:
Will be required to support the merge of https://github.com/JuliaGPU/AMDGPU.jl/pull/423 which requires a minor version bump.


luraess:
Now that AMDGPU v0.5.0 is released, one could update compat and maybe bump MPI.jl

PR #747, opened 2023-07-16
run_init_hooks() if MPI already initialized

### LINKED ISSUE #746: set_attr! fails when using initialized system-provided MPI
Me and @sloede are working on [`libtrixi`](https://github.com/trixi-framework/libtrixi/), which is a C / Fortran interface to the Julia-based flow solver [`Trixi.jl`](https://github.com/trixi-framework/Trixi.jl), which internally uses `MPI.jl`. In our typical user stories, an external application will have MPI already initialized when invoking `libtrixi`. We therefore followed the instructions for [Using a system-provided MPI backend](https://juliaparallel.org/MPI.jl/stable/configuration/#using_system_mpi).

Now we ran into the following error:
```
fatal: error thrown and no exception handler available.
MPI.MPIError(code=16)
MPI_Type_set_attr at ~/libtrixi-julia/dev/MPI/src/api/generated_api.jl:198 [inlined]
set_attr! at ~/libtrixi-julia/dev/MPI/src/datatypes.jl:46 [inlined]
init at ~/libtrixi-julia/dev/MPI/src/datatypes.jl:163
#65 at ~/libtrixi-julia/dev/MPI/src/datatypes.jl:168
get! at ./iddict.jl:178 [inlined]
Datatype at ~/libtrixi-julia/dev/MPI/src/datatypes.jl:156 [inlined]
create! at ~/libtrixi-julia/dev/MPI/src/datatypes.jl:475
unknown function (ip: 0x7eff828f85ee)
unknown function (ip: 0x7eff828e7ecd)
unknown function (ip: 0x7eff828e7e9c)
init at ~/libtrixi-julia/dev/MPI/src/datatypes.jl:160
#65 at ~/libtrixi-julia/dev/MPI/src/datatypes.jl:168
get! at ./iddict.jl:178 [inlined]
Datatype at ~/libtrixi-julia/dev/MPI/src/datatypes.jl:156 [inlined]
RBuffer at ~/libtrixi-julia/dev/MPI/src/buffers.jl:344 [inlined]
Reduce! at ~/libtrixi-julia/dev/MPI/src/collective.jl:592
Reduce! at ~/libtrixi-julia/dev/MPI/src/collective.jl:598
```
`Reduce` is called by `Trixi.jl` here.

Going through the code we realized that 
```
JULIA_TYPE_PTR_ATTR[] = create_keyval(Datatype)
```
is not called when MPI had already been initialized, leading to the error in `MPI_Type_set_attr`. 

We can fix this by running:
```
using MPI
if MPI.Initialized() && MPI.JULIA_TYPE_PTR_ATTR[]==0
    MPI.run_init_hooks()
end
```
Is there any, more natural way to get this right?

simonbyrne:
Fixes #746

### DIFF
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -154,9 +154,9 @@ function __init__()
         @require CUDA="052768ef-5323-5732-b1bb-66c8b64840ba" include("../ext/CUDAExt.jl")
     end
 
-    if MPI.Initialized()
+    if Initialized()
          # if MPI initialized elsewhere, e.g. when building a library
-        MPI.run_init_hooks()
+        run_init_hooks()
     end
 end
 
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -153,6 +153,11 @@ function __init__()
         @require AMDGPU="21141c5a-9bdb-4563-92ae-f87d6854732e" include("../ext/AMDGPUExt.jl")
         @require CUDA="052768ef-5323-5732-b1bb-66c8b64840ba" include("../ext/CUDAExt.jl")
     end
+
+    if MPI.Initialized()
+         # if MPI initialized elsewhere, e.g. when building a library
+        MPI.run_init_hooks()
+    end
 end
 
 


sloede (commented):
Thanks for having a look at this so quickly! Two questions:

* `set_default_error_handler_return`: Should this also be invoked if MPI was initialized externally? Or would one reasonably assume that in that case the external application is responsible for managing the error handling and not Julia?
* This PR wouldn't work in the case where MPI is initialized externally _after_ `using MPI` has been called, right?

Re the second point: Maybe this could be handled by recommending in the docs to call `MPI.run_init_hooks()` whenever MPI is to be used after it was initialized externally? As far as I can tell from
https://github.com/JuliaParallel/MPI.jl/blob/8d2d4c8170cf2e8ca95c4822a181a81c7a929981/src/environment.jl#L53-L59
`run_init_hooks` is idempotent and can safely be called multiple times, thus there's no need to actually have a `is_initialized_internal` function like I suggested in #746.

simonbyrne:
> * `set_default_error_handler_return`: Should this also be invoked if MPI was initialized externally? Or would one reasonably assume that in that case the external application is responsible for managing the error handling and not Julia?

Good question, I'm not sure to be honest.

> * This PR wouldn't work in the case where MPI is initialized externally _after_ `using MPI` has been called, right?

Good point: if you build a system image, I think the `__init__()` hooks are triggered when you load the library (which would typically be before `MPI_Init()`).

Perhaps the solution for now is to make `set_default_error_handler_return` and `run_init_hooks` well-defined APIs.

> `run_init_hooks` is idempotent and can safely be called multiple times, thus there's no need to actually have a `is_initialized_internal` function like I suggested in #746.

Yes, that was my intention.

sloede:
> Perhaps the solution for now is to make `set_default_error_handler_return` and `run_init_hooks` well-defined APIs.

I think this is maybe the best course of action for now, until we (and possibly others) have gathered more experience with this use case.

What about the following approach:

1. Extend the [External libraries and packages](https://juliaparallel.org/MPI.jl/v0.20/external/) section of the MPI.jl docs with a new entry along the following lines:
   ```markdown
   ## Externally initialized MPI
   When working with non-Julia libraries or tools, `MPI_Init` may be invoked in another part
   of the execution flow and not via MPI.jl's [`MPI.Init`](@ref) function. This leaves some
   package-internal settings uninitialized. In this case, you need to call
   [`MPI.run_init_hooks`)(@ref) manually to fully initialize MPI.jl. You may also want to
   consider calling [`MPI.set_default_error_handler_return`](@ref).
   ```
2. Add a docstring to `run_init_hooks` with something like this:
   ```markdown
   """
       run_init_hooks()

   Execute all functions that have been registered using [`add_init_hook!`](@ref).

   This function is executed automatically by [`Init`](@ref) but may be invoked manually if MPI
   has been initialized externally by a direct call to `MPI_Init()`. It is safe to call this
   function multiple times (subsequent runs will be a no-op).
   """
   ```
3. Add a docstring to `set_default_error_handler_return`:
   ```markdown
   """
      set_default_error_handler_return()

   Set the error handler for `MPI_COMM_SELF` and `MPI_COMM_WORLD` to `MPI_ERRORS_RETURN`. This
   will cause certain MPI errors to appear as Julia exceptions.

   This function is executed automatically by [`Init`](@ref) but may be invoked manually if MPI
   has been initialized externally by a direct call to `MPI_Init()`. It is safe to call this
   function multiple times (subsequent runs will be a no-op).
   """
   ```
?

simonbyrne:
That sounds good: can you open a PR?

sloede:
> That sounds good: can you open a PR?

#748

simonbyrne:
I'll close this for now. If this does turn out to be useful, we can reconsider it.

PR #751, opened 2023-07-21
[CI] Use patched MVAPICH2 to avoid segfault

giordano:


giordano:
@simonbyrne this solves the segmentation fault in MVAPICH, but we need to add the Spack recipe and the patch to this repo.  Are you ok with that?  I can try to upstream the patch to Spack, but I don't know whether they want to be bug-compatible with upstream projects, if so they'd likely not accept this patch.

giordano:
The PR in Spack was accepted and this is already working in #716: https://github.com/JuliaParallel/MPI.jl/actions/runs/5645989211/job/15292961084?pr=716.  So I can close this PR, nothing to do here.

PR #752, opened 2023-07-21
Add compat for AMDGPU.jl in test environment

giordano:


giordano:
AMDGPU.jl 0.3 isn't compatible with Julia v1.6 at all.  This can't possibly work.

PR #753, opened 2023-07-22
Update AMDGPU test

luraess:
Modify sync function.

Other infos:

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MPI"
 uuid = "da04e1cc-30fd-572f-bb4f-1f8673147195"
 authors = []
-version = "0.20.11"
+version = "0.20.12"
 
 [deps]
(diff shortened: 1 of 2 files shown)


giordano:
As a general remark working with Git and GitHub, opening a PR from a branch with the same name as the target branch is an [anti-pattern](https://blog.jasonmeridth.com/posts/do-not-issue-pull-requests-from-your-master-branch/). You should always create a new branch. This is the reason why the history of this PR has completely unrelated commits, which make revie confusing.

luraess:
> As a general remark working with Git and GitHub, opening a PR from a branch with the same name as the target branch is an [anti-pattern](https://blog.jasonmeridth.com/posts/do-not-issue-pull-requests-from-your-master-branch/). You should always create a new branch. This is the reason why the history of this PR has completely unrelated commits, which make review confusing.

Closing this in favour of #754

PR #754, opened 2023-07-22
Update AMDGPU test to 0.5

luraess:
Modify sync function.

Other infos:


luraess:
@giordano don't know why the history stills shows up here. Sorry.

giordano:
Maybe because you just branched out from your already borked `master` branch, instead of `origin/master`.  You should never do local changes to your `master` branch for this reason, all branches will be messed up.

luraess:
Let me try fixing this then.

luraess:
Should be good now in #755

PR #755, opened 2023-07-22
Update AMDGPU test

luraess:
Modify sync function.

Else:


vchuravy:
Can you update the Prohject.toml and do we need to directly use `AMDGPU.HIP` can we use `AMDGPU.synchronize` directly?

PR #757, opened 2023-07-29
Print libmpi dlpath in versioninfo

simonbyrne:


src/implementations.jl
```diff
@@ -113,6 +113,7 @@ function versioninfo()
     println()
     println("Library information:")
     println("  libmpi:  ", API.libmpi)
+    println("  libmpi dlpath:  ", Libdl.dlpath(API.libmpi))
```
giordano, line 116: This is good, but should we also have a variable `libmpi_path` in the `API` submodule?  We already have that for JLLs, we should need to add it only for system libmpi.

PR #758, opened 2023-07-29
Implement automatically serializing `scatter`

lukas-weber:
Thanks for merging the `gather` pull request! I went ahead and implemented `scatter` as well!

The current version does a tiny amount of unnecessary work by serializing `objs[root + 1]`, which is never sent. I figured saving this might not be worth the additional complexity. Happy to implement it that way though.

### DIFF
--- a/src/collective.jl
+++ b/src/collective.jl
@@ -186,11 +186,18 @@ function scatter(objs::Union{AbstractVector, Nothing}, comm::Comm; root::Integer
             throw(ArgumentError("Length of argument objs ($(length(objs))) != number of ranks in comm ($(Comm_size(comm)))."))
         end
 
-        datasegments = MPI.serialize.(objs)
-        counts = length.(datasegments)
+        sendbuffer = IOBuffer()
+        counts = Vector{Int64}(undef, length(objs))
+
+        last_pos = 0
+        for (i, obj) in enumerate(objs)
+            Serialization.serialize(sendbuffer, i == root + 1 ? nothing : obj)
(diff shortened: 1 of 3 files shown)


src/collective.jl
```diff
@@ -167,6 +167,45 @@ Scatter(sendbuf, T, comm; root::Integer=Cint(0)) =
...
+    if isroot
+        if length(objs) != Comm_size(comm)
+            throw(ArgumentError("Length of argument objs ($(length(objs))) != number of ranks in comm ($(Comm_size(comm)))."))
+        end
+
+        datasegments = MPI.serialize.(objs)
+        counts = length.(datasegments)
+
+        count = Scatter(counts, Int64, comm; root = root)
+        sendbuf = VBuffer(vcat(datasegments...), counts)
```
simonbyrne, line 193: Splatting vectors is generally a bad idea. 

The optimal approach would be to incrementally serialize to a single buffer, but at the minimum we should call `reduce`:
```suggestion
        sendbuf = VBuffer(reduce(vcat, datasegments), counts)
```

lukas-weber:
Whoopsie, I was always wondering about the preferred way to do that. But here is a version that uses a single buffer. Don’t know if there is a better way to calculate the counts.

simonbyrne (approved):

simonbyrne:
thank you!

simonbyrne:
I've sent an invite to give you commit access.

lukas-weber:
Thanks!

PR #760, opened 2023-08-15
Add optional io argument to versioninfo()

mfherbst:
Make `versioninfo` behave similar to what is done in InteractiveUtils, CUDA, AMDGPU, ...

### DIFF
--- a/src/implementations.jl
+++ b/src/implementations.jl
@@ -82,41 +82,41 @@ const MPI_VERSION = Get_version()
 
 using PkgVersion
 """
-    MPI.versioninfo()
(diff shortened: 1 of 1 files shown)


src/implementations.jl
```diff
@@ -82,41 +82,41 @@ const MPI_VERSION = Get_version()
 
 using PkgVersion
 """
-    MPI.versioninfo()
+    MPI.versioninfo([io::IO=stdout])
```
lcw, line 85:
```suggestion
    MPI.versioninfo(io::IO=stdout)
```

PR #763, opened 2023-09-05
Fix ROCm-aware MPI extension

utkinis:
This PR fixes a bug introduced with a recent update of AMDGPU.jl

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 *.mem
 Manifest.toml
 LocalPreferences.toml
+
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@
 *.mem
 Manifest.toml
 LocalPreferences.toml
-
-.vscode/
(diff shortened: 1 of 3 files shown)


luraess:
We 

> Does this require a compat bounds update? How about a regression test?

Yes, as already done in #755. But let's do it here.

utkinis:
@giordano I've updated combat bounds for AMDGPU.jl. There is no other functionality to do the regression testing against, this change just fixes the conversion of ROCArray to a Julia pointer, which didn't exist in AMDGPU prior to 0.5.7

luraess:
Can we merge this and patch release?

.gitignore
```diff
@@ -2,3 +2,5 @@
 *.mem
 Manifest.toml
 LocalPreferences.toml
+
+.vscode/
```
simonbyrne, line 6: This should go in your own excludes file, see https://git-scm.com/docs/gitignore

simonbyrne, line 6:
```suggestion
```

luraess:
Thanks!

Issue #762, opened 2023-09-04
Implement cooperative test


luraess:
Seems it would be desired to have ` cooperative_test!` as part of MPI.jl (cc @vchuravy)

Where would it make sense to be added to within MPI.jl

Code would be close to:
```julia
function cooperative_test!(req)
    done = false
    while !done
        done, _ = MPI.Test(req, MPI.Status)
        yield()
    end
end
```

luraess:
@giordano any hint on where to place this one? 🙏

lcw:
I am not certain this function should be in MPI.jl.  I am curious what others think.

luraess:
> I am not certain this function should be in MPI.jl. I am curious what others think.

Unsure either - it was @vchuravy 's suggestion. I could happily have it hosted elsewhere as well.

simonbyrne:
What it sounds like you want (and what I also want) is an `MPI.Wait` that acts like `Base.wait` (i.e. will yield to other Julia tasks while blocking). Since MPI doesn't provide an interface for calling back into task-based runtimes, there are two basic approaches:
1. spin loop with `MPI.Test` and `yield()` (as is proposed here)
2. put the `MPI.Wait` on a separate thread (as I proposed in #452) (either a Julia thread or a libuv thread), and notify once it is done.

You can also use a hybrid of the two. e.g. here is what CUDA.jl does for stream/device synchronization:
https://github.com/JuliaGPU/CUDA.jl/blob/08f9c3da46fc5814406cd04c7d4701e641d1722c/lib/cudadrv/synchronization.jl
(basically it does a spinning synchronize:
   https://github.com/JuliaGPU/CUDA.jl/blob/08f9c3da46fc5814406cd04c7d4701e641d1722c/lib/cudadrv/synchronization.jl#L77-L97
  and if that fails, blocks on a libuv thread which notifies an event when it wakes up.

Some earlier discussion on discourse:
https://discourse.julialang.org/t/how-to-combine-mpi-non-blocking-isend-irecv-and-julia-tasks/52524/5?u=simonbyrne

luraess:
[...]

PR #766, opened 2023-09-06
Add non-blocking wait

luraess:
A first implementation of non-blocking wait as discussed in #762

### DIFF
--- a/docs/src/reference/pointtopoint.md
+++ b/docs/src/reference/pointtopoint.md
@@ -55,7 +55,7 @@ MPI.Testall
 MPI.Testany
 MPI.Testsome
 MPI.Wait
-Base.wait
+Base.wait(req::MPI.Request)
 MPI.Waitall
 MPI.Waitany
 MPI.Waitsome
--- a/docs/src/reference/pointtopoint.md
+++ b/docs/src/reference/pointtopoint.md
@@ -55,6 +55,7 @@ MPI.Testall
 MPI.Testany
 MPI.Testsome
 MPI.Wait
+Base.wait
 MPI.Waitall
 MPI.Waitany
 MPI.Waitsome
--- a/src/nonblocking.jl
+++ b/src/nonblocking.jl
@@ -778,10 +778,10 @@ end
 """
     Base.wait(req::MPI.Request)
 
-Non-blocking wait. This will yield to other (Julia) tasks while blocking in a spin loop.
+Wait for an MPI request to complete. This will yield to other (Julia) tasks resulting in a cooperative wait.
 """
 function Base.wait(req::MPI.Request)
     while !MPI.Test(req)
         yield()
     end
- end
+end
--- a/src/nonblocking.jl
+++ b/src/nonblocking.jl
@@ -778,7 +778,7 @@ end
 """
     Base.wait(req::MPI.Request)
 
-Wait for an MPI request to complete. This will yield to other (Julia) tasks resulting in a cooperative wait.
+Wait for an MPI request to complete. Unlike [`MPI.Wait`](@ref), it will yield to other Julia tasks resulting in a cooperative wait.
 """
 function Base.wait(req::MPI.Request)
     while !MPI.Test(req)
--- a/src/nonblocking.jl
+++ b/src/nonblocking.jl
@@ -774,3 +774,14 @@ function Cancel!(req::AbstractRequest)
     API.MPI_Cancel(req)
     nothing
 end
+
+"""
+    Base.wait(req::MPI.Request)
+
+Non-blocking wait. This will yield to other (Julia) tasks while blocking in a spin loop.
+"""
+function Base.wait(req::MPI.Request)
+    while !MPI.Test(req)
+        yield()
+    end
+ end
--- a/test/test_cooperative_wait.jl
+++ b/test/test_cooperative_wait.jl
@@ -16,14 +16,13 @@ recv_check = zeros(Int, nsends)
 
 @sync for i = 1:nsends
     Threads.@spawn begin
-        # send to self
         recv_req = MPI.Irecv!(recv_arr[i], MPI.COMM_WORLD; source=myrank, tag=i)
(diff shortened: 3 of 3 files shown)


vchuravy:
Should we also add a `MPI.wait(t::Task)`?

```
function wait(task::Task)
    while !Base.istaskdone(task)
        MPI.Iprobe(MPI.MPI_ANY_SOURCE, MPI.MPI_ANY_TAG, MPI.COMM_WORLD)
        yield()
    end
    wait(task)
end
```

luraess:
> Should we also add a `MPI.wait(t::Task)`?

If so, should one add it in similar location as the `Base.wait`?

simonbyrne:
I think leave that for a separate PR.

simonbyrne:
Can you add it to the docs, and add a test?

luraess:
> Can you add it to the docs, and add a test?

Done. Feel free to update the docstring @simonbyrne

src/nonblocking.jl
```diff
@@ -774,3 +774,14 @@ function Cancel!(req::AbstractRequest)
     API.MPI_Cancel(req)
     nothing
 end
+
+"""
+    Base.wait(req::MPI.Request)
+
+Wait for an MPI request to complete. This will yield to other (Julia) tasks resulting in a cooperative wait.
```
simonbyrne, line 781:
```suggestion
Wait for an MPI request to complete. Unlike [`MPI.Wait`](@ref), it will yield to other Julia tasks resulting in a cooperative wait.
```

test/test_cooperative_wait.jl
```diff
@@ -0,0 +1,38 @@
...
+        send_req = MPI.Isend(send_arr[i], MPI.COMM_WORLD; dest=myrank, tag=i)
+
+        wait(recv_req)
+        @test MPI.isnull(recv_req)
+        recv_check[i] += 1
+
+        wait(send_req)
+        @test MPI.isnull(send_req)
+        send_check[i] += 1
+    end
```
simonbyrne, line 30: I'm not sure that this is correct, in that it seems to assume that you can wait on the receive before the send is complete (this is typically true for short messages, but not guaranteed by the spec).

How about something like:
```suggestion
    Threads.@spawn begin
        recv_req = MPI.Irecv!(recv_arr[i], MPI.COMM_WORLD; source=myrank, tag=i)
        wait(recv_req)
        @test MPI.isnull(recv_req)
        recv_check[i] += 1
    end
    Threads.@spawn begin
        send_req = MPI.Isend(send_arr[i], MPI.COMM_WORLD; dest=myrank, tag=i)
        wait(send_req)
        @test MPI.isnull(send_req)
        send_check[i] += 1
    end
```

docs/src/reference/pointtopoint.md
```diff
@@ -55,6 +55,7 @@ MPI.Testall
 MPI.Testany
 MPI.Testsome
 MPI.Wait
+Base.wait
```
vchuravy, line 58:
```suggestion
Base.wait(req::MPI.Request)
```

I think this might only give us the docstring for the method added here?

test/test_cooperative_wait.jl
luraess: I am not sure I follow your point. Also, how would having 2 threads guarantee the following to be true.
> you can wait on the receive before the send is complete

Why would the semantic change from usual non-blocking MPI point to point comm which has the initially proposed workflow (as in [here](https://enccs.github.io/intermediate-mpi/non-blocking-communication-pt1/#performant-mpi-workflows-for-stencil-applications-non-blocking-messages)). Initially suggested, we have the non-blocking workflow done per MPI process in a sequential way, and have one thread per process.

luraess:
Could this be reviewed and finalised 🙏 ?

test/test_cooperative_wait.jl
simonbyrne: In your original proposal, the `Isend` and `Irecv!` are in one Julia task, and you `wait` on the receive first: this means that the receive has to complete before it can wait on the send. This is typically fine, but there is no guarantee that it actually will complete. From https://stackoverflow.com/a/67099566/392585

> MPI_Isend does not necessarily progress in the background but only when the MPI implementation is given the chance to progress it. MPI_Wait progresses the operation and guarantees its completion. Some MPI implementations can progress operations in the background using progression threads. Some cannot. It is implementation-dependent and one should never rely on one or another specific behaviour.

My proposal splits the send and receive into two Julia tasks: if the task is waiting on the receive, it can then switch to the send task and complete that.

luraess: Ok. Thanks for the details.

lcw: @simonbyrne  I don't quite understand the need for threads.  Page 75 of the MPI 4.0 standard (section 3.7 Nonblocking Communication) states:

> Progress A call to MPI_WAIT that completes a receive will eventually terminate and return
if a matching send has been started , unless the send is satisfied by another receive. In
particular, if the matching send is nonblocking, then the receive should complete even if no
call is executed by the sender to complete the send. Similarly, a call to MPI_WAIT that
completes a send will eventually return if a matching receive has been started, unless the
receive is satisfied by another send, and even if no call is executed to complete the receive.

Is there something that I am missing?  The above indicates to me that the wait on the receive should return even if the send has not been waited on.

p.s. This statement is also on page 45 of the MPI 1.3 standard.

simonbyrne: Ah, TIL! Thanks @lcw.

lcw: Your welcome.  I was worried I had been using an invalid assumption for 20 years.  So, I had to look it up.

luraess: > I don't quite understand the need for threads

The usage of threads here emulates a use case in which one would use a TLS approach to perform some async work on GPUs which requires to launch kernels on different streams from different Julia tasks/threads. This thus requires the "cooperative" wait. But indeed, I was also confused by the fact that non-blocking comm would stall in the single thread loop implementation.

luraess:
Shall one thus revert the test and spawn a single thread to do the non-blocking send and receive within as originally proposed?

PR #770, opened 2023-09-13
Init cooperative wait test for multiple threads

lcw:
In the cooperative wait test multiple threads are calling MPI functions in a non-serialized way. Thus, MPI needs to be initialized with `threadlevel=:multiple`.

### DIFF
--- a/test/test_cooperative_wait.jl
+++ b/test/test_cooperative_wait.jl
@@ -1,7 +1,7 @@
 # tests for the various kinds of waits
 include("common.jl")
 
-MPI.Init()
+MPI.Init(threadlevel=:multiple)
 
 myrank = MPI.Comm_rank(MPI.COMM_WORLD)
 commsize = MPI.Comm_rank(MPI.COMM_WORLD)


simonbyrne (approved):

test/test_cooperative_wait.jl
```diff
@@ -1,7 +1,7 @@
 # tests for the various kinds of waits
 include("common.jl")
 
-MPI.Init()
+MPI.Init(threadlevel=:multiple)
```
vchuravy, line 4: Isn't that our default?

lcw: Nope, the default is ``:serialized`, see https://github.com/JuliaParallel/MPI.jl/blob/a57528f7686431a47e009faf882d16ec23eacaeb/src/environment.jl#L114

luraess: Is this always the case here or only if one launches Julia with `threads > 1` ?

lcw: Good question.  I guess it would be only when julia is launched with more than one thread.  But Valentin would know better.

simonbyrne: We went with `:serialized` as that is the most common use case, and not all MPI libraries support `:multiple` by default (or they might incur additional overhead)

luraess: So ideally one should always have 
```
(Threads.nthreads() > 1) ? MPI.Init(threadlevel=:multiple) : MPI.Init()
```

vchuravy: This doesn't account for interactive threadpool usage, or foreign threads being dynamically added.

`maxthreadid` is probably okay, but that get's skewed by GC threads and the IO thread.

PR #772, opened 2023-09-19
Fix resize! for (Unsafe)MultiRequest

jipolanco:
This fixes `resize!` for `MultiRequest` and `UnsafeMultiRequest`.

Note that `resize!` was broken for these types since `MultiRequestItem` doesn't have `val` or `buffer` fields.


jipolanco:
Docs build should be fixed by #773.

simonbyrne (approved):

PR #774, opened 2023-09-26
Bump CUDA compat

simonbyrne:


simonbyrne:
Good catch!

lcw:
Is this ready to merge and a new release made?

simonbyrne:
We should probably make sure the CUDA tests pass...

lcw:
> We should probably make sure the CUDA tests pass...

Probably a good idea :-).  It looks like there are some scalar indexing issues at least.

luraess:
Yeah - seems indeed first failure occurs because of scalar indexing (was already the case with CUDA 4). For ROCm, it could be that the issue is related to an not fully working ROCm-aware MPI install on the CI machine

PR #706, opened 2023-01-13
Consistently use `Base.julia_cmd()` in tests

giordano:
This should in principle avoid running in https://github.com/JuliaLang/julia/issues/48039, but locally I'm still seeing the same problem :confused:

PR #783, opened 2023-10-16
Add mpiexecjl for windows

### LINKED ISSUE #545: `mpiexecjl` should use a PowerShell on Windows.

guilhermebodin:
close #545 

This is an attempt to have an mpiexecjl on windows. Currently, you have to call `powershell.exe -File ...` This could be avoided if we did a .bat file calling a .ps1 file but I find this too ugly.

### DIFF
--- a/test/mpiexecjl.jl
+++ b/test/mpiexecjl.jl
@@ -1,6 +1,88 @@
 using Test, Pkg
 using MPI
 
+function my_julia_cmd(julia=joinpath(Sys.BINDIR, Base.julia_exename()); cpu_target::Union{Nothing,String} = nothing)
+    opts = Base.JLOptions()
+    if cpu_target === nothing
+        cpu_target = unsafe_string(opts.cpu_target)
+    end
+    image_file = unsafe_string(opts.image_file)
+    addflags = String[]
+    let compile = if opts.compile_enabled == 0
+                      "no"
+                  elseif opts.compile_enabled == 2
+                      "all"
+                  elseif opts.compile_enabled == 3
+                      "min"
+                  else
+                      "" # default = "yes"
+                  end
+        isempty(compile) || push!(addflags, "--compile=$compile")
+    end
+    let depwarn = if opts.depwarn == 1
+                      "yes"
+                  elseif opts.depwarn == 2
+                      "error"
+                  else
+                      "" # default = "no"
+                  end
+        isempty(depwarn) || push!(addflags, "--depwarn=$depwarn")
+    end
+    let check_bounds = if opts.check_bounds == 1
+                      "yes" # on
+                  elseif opts.check_bounds == 2
+                      "no" # off
+                  else
+                      "" # default = "auto"
+                  end
+        isempty(check_bounds) || push!(addflags, "--check-bounds=$check_bounds")
+    end
+    opts.can_inline == 0 && push!(addflags, "--inline=no")
+    opts.use_compiled_modules == 0 && push!(addflags, "--compiled-modules=no")
+    opts.opt_level == 2 || push!(addflags, "-O$(opts.opt_level)")
+    opts.opt_level_min == 0 || push!(addflags, "--min-optlevel=$(opts.opt_level_min)")
+    push!(addflags, "-g$(opts.debug_level)")
+    if opts.code_coverage != 0
+        # Forward the code-coverage flag only if applicable (if the filename is pid-dependent)
+        coverage_file = (opts.output_code_coverage != C_NULL) ?  unsafe_string(opts.output_code_coverage) : ""
+        if isempty(coverage_file) || occursin("%p", coverage_file)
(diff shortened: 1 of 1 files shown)


giordano:
Can you please rebase on `master`?  CI situation should be much better now.

guilhermebodin:
@giordano it is currently failing silently with some very weird error

```
could not load library "D:\a\MPI.jl\MPI.jl\test\C"
The specified module could not be found. 
ERROR: failed process: Process(`'C:\Users\runneradmin\.julia\artifacts\4c2053ce1fd1b18af0517bdac687a718f567cdae\bin\mpiexec.exe' -n 4 'C:\hostedtoolcache\windows\julia\1.6.7\x64\bin\julia.exe' -Cnative -JC '\hostedtoolcache\windows\julia\1.6.7\x64\lib\julia\sys.dll' --depwarn=yes --check-bounds=yes -g1 --code-coverage=user --color=yes --startup-file=no --startup-file=no -q 'D:\a\MPI.jl\MPI.jl\test\..\docs\examples\01-hello.jl'`, ProcessExited(1)) [1]
```

It is misparsing the sysimage part `-JC` should be `-JC:\hostedtoolcache....` I have tried many things locally but could not solve the problem. Everytime I try to parse the command following command it comes out correctly.
```
run(`$(additional_initial_parts_cmd) $(mpiexecjl) -n $(nprocs) --project=$(dir) $(julia) --startup-file=no -q $(example)`)
```
I could not identify where it gets lost in the middle. Honestly, the $(julia) variable is used only for testing, we could simply write `julia` directly without betting it with `Base.julia_cmd()` and it would work correctly.

giordano (commented):
> it is currently failing silently with some very weird error

Why is the error silent and not loud?  Sounds like that should be fixed.

> It is misparsing the sysimage part

Where is this parsing happening?

> Honestly, the $(julia) variable is used only for testing, we could simply write julia directly without betting it with Base.julia_cmd() and it would work correctly.

No, it wouldn't:  #706

test/mpiexecjl.jl
```diff
@@ -15,24 +15,37 @@ using MPI
...
+        else
+            String["powershell.exe", "-ExecutionPolicy", "Bypass", "-File"]
+        end
+
         # `Base.julia_cmd()` ensures keeping consistent flags when running subprocesses.
         julia = Base.julia_cmd()
         example = joinpath(@__DIR__, "..", "docs", "examples", "01-hello.jl")
         env = ["JULIA_BINDIR" => Sys.BINDIR]
         p = withenv(env...) do
-            run(`$(mpiexecjl) -n $(nprocs) --project=$(dir) $(julia) --startup-file=no -q $(example)`)
```
giordano, line 24: Why was `--startup-file=no` removed from here?  That'd create problems.

guilhermebodin:
@giordano I fixed many things in the comments and got the Powershell script to not fail silently. It will rethrow the exit code of the underlying julia process.

From my perspective the only problem left now is that powershell is splitting the command line switch:
`-JC\hostedtoolcache\windows\julia\1.10.2\x64\lib\julia\sys.dll` 
into two separate ones
`-JC` and `\hostedtoolcache\windows\julia\1.10.2\x64\lib\julia\sys.dll`.

This happens in the PowerShell argument parsing. I am investigating how to avoid this behavior.

guilhermebodin:
Found some material on it https://github.com/PowerShell/PowerShell/issues/16432.

guilhermebodin:
@giordano The only solution I found with Powershell was to replace the `-J` for `--sysimage=` in the `julia_cmd`. I don't think that this is reasonable but it was the only thing that worked.

test/mpiexecjl.jl
```diff
@@ -1,6 +1,88 @@
...
+    if opts.use_sysimage_native_code == 0
+        push!(addflags, "--sysimage-native-code=no")
+    end
+    if opts.use_pkgimages == 0
+        push!(addflags, "--pkgimages=no")
+    else
+        # If pkgimage is set, malloc_log and code_coverage should not
+        @assert opts.malloc_log == 0 && opts.code_coverage == 0
+    end
+    return `$julia -C$cpu_target --sysimage=$image_file $addflags`
```
guilhermebodin, line 83: I only edited this part of the `Base.julia_cmd()` function

juwens:
@guilhermebodin you can workaround this nasty pwsh quirk by using `--%` the "[stop parsing token](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_parsing?view=powershell-7.5#the-stop-parsing-token)" 

```
run(`--% $(additional_initial_parts_cmd) $(mpiexecjl) -n $(nprocs) --project=$(dir) $(julia) --startup-file=no -q $(example)`)
    ^^^
```

I guess it was inspired by bash's [`--`](https://unix.stackexchange.com/a/11382/13534)

PR #788, opened 2023-11-03
[CI] Use Spack OCI build cache for MPI packages

giordano:
Fix #744.

PR #789, opened 2023-11-03
Add compat bounds for more packages

giordano:


Project.toml
```diff
@@ -19,13 +19,20 @@ Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
...
 [compat]
+Distributed = "1"
 AMDGPU = "0.5.7, 0.6"
 CUDA = "3, 4, 5"
 DocStringExtensions = "0.8, 0.9"
+Libdl = "1"
+MPICH_jll = "4"
 MPIPreferences = "0.1.8"
+MPItrampoline_jll = "5"
+OpenMPI_jll = "4"
```
giordano, line 30: Note that we have version 5 as well, but at the moment there's something very broken, so it doesn't work:
```
julia> Base.BinaryPlatforms.HostPlatform()
Linux x86_64 {cxxstring_abi=cxx11, julia_version=1.9.3, libc=glibc, libgfortran_version=5.0.0, libstdcxx_version=3.4.30}

julia> OpenMPI_jll.is_available()
false
```

simonbyrne: are you using the OpenMPI binary?

giordano: I'm not using MPI, I just installed the `OpenMPI_jll` package.

simonbyrne: I think we only load OpenMPI_jll if the appropriate preferences are set.

giordano: Ok, but this was failing in #788 to import OpenMPI_jll v5, no problem with v4.  I honestly don't have the time to track that one down.

PR #792, opened 2023-11-03
[CI] Always run Julia with `--check-bounds=yes`

giordano:
Let's see if this reduces the amount of recompilation between different Julia sessions within the same job.

### DIFF
--- a/.github/workflows/UnitTests.yml
+++ b/.github/workflows/UnitTests.yml
@@ -28,524 +28,524 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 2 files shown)


giordano:
Nevermind, this doesn't seem to work and I don't have many more ideas for the time being.

For the record, arguments in Julia v1.9 are something like
```
-Cnative -J/path/to/lib/julia/sys.so --depwarn=yes --check-bounds=yes -g1 --code-coverage=@/path/to/project --color=yes --startup-file=no --pkgimages=no
```

PR #800, opened 2023-11-14
Create 10-job_scheduler_rma_acc.jl

elbert5770:
An additional example that combines a job scheduler with passive RMA and MPI.Accumulate!


elbert5770:
The MPI.Accumulate! could also be done in one shot: MPI.Accumulate!(matches, root,0,MPI.SUM, win_accumulate).

PR #803, opened 2023-11-24
AMDGPU compat

luraess:
Bump compat to latest 0.8 release.


simonbyrne (approved):
If you bump the patch version, we can tag a release.

PR #812, opened 2024-02-01
RMA improvements

jeffhammond:
Adds:
- win_allocate
- win_lock_all
- win_unlock_all
- win_flush_local

Does not add:
- win_flush_local_all
- win_flush_all
- rput, raccumulate, rget, rget_accumulate

I don't know how to compile or test this so I am relying on your CI to do that for me.

### DIFF
--- a/docs/src/reference/onesided.md
+++ b/docs/src/reference/onesided.md
@@ -3,9 +3,13 @@
 ```@docs
 MPI.Win_create
 MPI.Win_create_dynamic
+MPI.Win_allocate
 MPI.Win_allocate_shared
 MPI.Win_shared_query
 MPI.Win_flush
+MPI.Win_flush_local
+MPI.Win_lock_all
+MPI.Win_unlock_all
 MPI.Win_lock
 MPI.Win_unlock
 MPI.Get!
--- a/src/onesided.jl
+++ b/src/onesided.jl
@@ -70,6 +70,42 @@ function Win_create(base::SubArray{T}, comm::Comm; infokws...) where T
 end
 
 """
+    win, array = MPI.Win_allocate(Array{T}, dims, comm::Comm; infokws...)
+
+Create and allocate a window for objects of type `T` of dimension
+`dims` (either an integer or tuple of integers), returning a `Win` and the
+`Array{T}` attached to the local process.
+
+This is a collective call over `comm`, but `dims` can differ for each call (and
+can be zero).
+
+If the underlying implementation supports MPI 4.1, 
+use [`MPI.Win_shared_query`](@ref) to obtain the `Array` attached to a different
+process in the same shared memory space.
+
+`infokws` are info keys providing optimization hints.
+
+[`MPI.free`](@ref) should be called on the `Win` object once operations have
+been completed.
+"""
+function Win_allocate(::Type{Ptr{T}}, len::Integer, comm::Comm; kwargs...) where T
+    win = Win()
+    out_baseptr = Ref{Ptr{T}}()
+    # int MPI_Win_allocate(MPI_Aint size, int disp_unit, MPI_Info info,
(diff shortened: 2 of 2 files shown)


jeffhammond:
I am not going to finish this until MPI.jl supports ARM.  I do not get paid to write code for Intel anymore.

https://github.com/JuliaParallel/MPI.jl/issues/404

giordano:
@jeffhammond I'm sorry to read you closed the PR for this reason, especially since a few of us, me included, are enthusiastic about ARM chips.  @vchuravy and I published a [paper](https://doi.org/10.1109/CLUSTER51413.2022.00072) about using Julia (and `MPI.jl`) on A64FX in the [Embracing Arm for HPC Workshop](https://arm-hpc-user-group.github.io/eahpc-2022/), and more recently I've been playing with [Julia on NVIDIA Grace Hopper](https://github.com/giordano/julia-on-gh200/), so we really want to support ARM as good as possible!

`MPI.jl` generally works well on ARM (as any other architectures where Julia runs), the issue you linked to refers only to custom MPI reductions, which make use of LLVM trampolines which are just not supported upstream in LLVM!  This issue also affected PowerPC, but then someone fixed the issue for that architecture in LLVM and then we automatically benefited from it.  If your company has LLVM engineers to spare, getting someone to fix this issue would benefit all compiler frontends based on LLVM targeting ARM architectures, not just Julia and `MPI.jl`!

jeffhammond:
I use custom MPI reductions quite a bit in my work, so they aren't something I can ignore.  They are an MPI 1.0 feature AFAIK, so it doesn't make sense for me to work on RMA things if they aren't available (for the systems I'm using).

I will ask if anyone is willing to fix this, but I'm still disappointed that there is no workaround besides fixing LLVM.

giordano:
> I will ask if anyone is willing to fix this

Thanks!  If that's of any help, this should be the upstream ticket: https://github.com/llvm/llvm-project/issues/4100

> but I'm still disappointed that there is no workaround besides fixing LLVM.

Julia uses LLVM quite a lot, so that's not totally unsurprising.

simonbyrne:
> I am not going to finish this until MPI.jl supports ARM. I do not get paid to write code for Intel anymore.

Unfortunately, no one else is paid to maintain or develop this package either, beyond what they immediately need for their own work. We're happy to discuss paths to fix the issue, but this sort of comment isn't really helpful.

jeffhammond:
I'm sorry. I was exasperated because I could not build test my changes on my laptop (Apple M2), which makes development harder. I can try to contribute to this on the weekend. 

I know this is an LLVM issue and I'm trying to help with that.

simonbyrne:
I understand: the test suite is kind of clunky, and could also do with some more improvement, but I thought we disable those tests on non-supported platforms. Feel free to ask on Slack for any questions/frustrations.

Regarding custom operators, one particular challenge that is specific to MPI is that the API doesn't allow you to pass through a "thunk" pointer to the reduction function, which is how we usually avoid this problem for callbacks in other libraries. Do you think there would be any appetite adding an additional API that would support this to the MPI standard? Or have custom operators gone out of style since they're not supported for RDMA operations?

jeffhammond:
Custom reductions aren't optimized but the use cases tolerate that. RDMA doesn't do most MPI ops anyways, certainly not in bulk. 

I don't know what thunk means here but in Mukautuva, I have a callback trampoline that uses datatype attributes to carry state from the caller to the callee, so I can invoke a different function. Does that sound relevant?

giordano:
@jeffhammond now that we have a workaround for custom reduction operations on aarch64 in the form of #871 (closures can't still be used as operators, but regular functions defined at the top-level of the program can now be used), I was wondering if you had interest/time to resume this PR.

PR #816, opened 2024-02-04
Add high-level API `Scatter` for splitting 1D array

ykkan:
Following the idea of `Gather`, a high-level API `Scatter` for `Scatter!` is provided. With this API, users can easily scatter a 1D array without handling `VBuffer` or `UBuffer`.
For example: 
```
MPI.Init()
comm = MPI.COMM_WORLD
rank = MPI.Comm_rank(comm)
nprocs = MPI.Comm_size(comm)


arr = nothing
if rank == 0
  arr = [1:10...]
end

local_arr = MPI.Scatter(data, comm)

print("Hello world, I am rank $(rank) of $(nprocs), $(local_arr)\n")
```  
Output (`nprocs=3`):
```
Hello world, I am rank 0 of 3, [1, 2, 3, 4]
Hello world, I am rank 2 of 3, [8, 9, 10]
Hello world, I am rank 1 of 3, [5, 6, 7]
```

### DIFF
--- a/src/collective.jl
+++ b/src/collective.jl
@@ -168,6 +168,38 @@ Scatter(sendbuf, ::Type{T}, root::Integer, comm::Comm) where {T} =
     Scatter!(sendbuf, Ref{T}(), root, comm)[]
 
 """
+    Scatter(arr::AbstractVector, comm::Comm; root::Cint=0)
+
+Splits a 1D array `arr` with elements of the same type in the `root` process into `nprocs=Comm_size(comm)` smaller 1D arrays. 
+The array `arr` is splitted in rank order, and the number of elements `n=length(arr)` can be not divisible by `nprocs`.
+Each process with the rank `j` returns a smaller array with the number of elements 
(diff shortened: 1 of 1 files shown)


src/collective.jl
```diff
@@ -167,6 +167,38 @@ Scatter(sendbuf, T, comm; root::Integer=Cint(0)) =
...
+  nprocs = MPI.Comm_size(comm)
+
+  arr_len = 0
+  elm_t = nothing
+  if rank == root
+    arr_len = length(arr)
+    elm_t = eltype(arr)
+  end
+  arr_len = MPI.Bcast(arr_len, root, comm)
+  elm_t = MPI.bcast(elm_t, root, comm)
```
vchuravy, line 189: Why is one of these `Bcast` and the other `bcast`?

ykkan: If I remember correctly, `elm_t` is not a `isbits` type and can only be broadcasted with `bcast`. My idea was to use `Bcast` whenever possible, as its overhead should be less than `bcast`(?). Of course, using `bcast` for both `elm_t` and `arr_len` should work. I think it is a matter of style. What do you think?

PR #821, opened 2024-02-25
Add `has_rocm` for OpenMPI

avik-pal:
See https://docs.open-mpi.org/en/main/man-openmpi/man3/MPIX_Query_rocm_support.3.html#mpix-query-rocm-support

Seems to be available from v5 of OpenMPI

### DIFF
--- a/src/environment.jl
+++ b/src/environment.jl
@@ -327,6 +327,8 @@ or `false`.
 
 !!! note
     For OpenMPI or OpenMPI-based implementations you first need to call [Init()](@ref).
+
+See also [`MPI.has_rocm`](@ref) for ROCm support.
 """
 function has_cuda()
     flag = get(ENV, "JULIA_MPI_HAS_CUDA", nothing)
@@ -354,6 +356,8 @@ provides a mechanism to check, so it will return `false` with other implementati
 
 This can be overriden by setting the `JULIA_MPI_HAS_ROCM` environment variable to `true`
 or `false`.
+
+See also [`MPI.has_cuda`](@ref) for CUDA support.
 """
 function has_rocm()
     flag = get(ENV, "JULIA_MPI_HAS_ROCM", nothing)
@@ -369,3 +373,17 @@ function has_rocm()
         return parse(Bool, flag)
     end
 end
+
+"""
+    MPI.has_gpu()
+
+Checks if the MPI implementation is known to have GPU support. Currently this checks for the
+following GPUs:
+
+1. CUDA: via [`MPI.has_cuda`](@ref)
+2. ROCm: via [`MPI.has_rocm`](@ref)
+
+See also [`MPI.has_cuda`](@ref) and [`MPI.has_rocm`](@ref) for more fine-grained
+checks.
+"""
+has_gpu() = has_cuda() || has_rocm()
--- a/src/environment.jl
+++ b/src/environment.jl
@@ -320,7 +320,7 @@ Wtime() = API.MPI_Wtime()
 
 Check if the MPI implementation is known to have CUDA support. Currently only Open MPI
 provides a mechanism to check, so it will return `false` with other implementations
-(unless overriden).
+(unless overriden). For "IBMSpectrumMPI" it will return `true`.
 
 This can be overriden by setting the `JULIA_MPI_HAS_CUDA` environment variable to `true`
 or `false`.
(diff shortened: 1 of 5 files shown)


src/environment.jl
```diff
@@ -341,3 +341,28 @@ function has_cuda()
...
+    MPI.has_rocm()
+
+Check if the MPI implementation is known to have ROCm support. Currently only Open MPI
+provides a mechanism to check, so it will return `false` with other implementations
+(unless overriden).
+
+This can be overriden by setting the `JULIA_MPI_HAS_ROCM` environment variable to `true`
+or `false`.
+"""
+function has_rocm()
```
vchuravy, line 355: Maybe we can add a third function `has_gpu()` that calls `has_cuda() || has_rocm()`

src/environment.jl
```diff
@@ -341,3 +341,28 @@ function has_cuda()
...
+    flag = get(ENV, "JULIA_MPI_HAS_ROCM", nothing)
+    if flag === nothing
+        # Only Open MPI provides a function to check ROCm support
+        @static if MPI_LIBRARY == "OpenMPI"
```
vchuravy, line 359: We probably need to use `MPI_LIBRARY_VERSION` to check that this is indeed `v5`

avik-pal: The current compat is on 4. Should we hold it off till newer JLLs are supported?

giordano: Someone will have to debug what's the problem with OpenMPI 5: https://github.com/JuliaParallel/MPI.jl/pull/789#discussion_r1382022041

vchuravy: Someone could already run with a local version of OpenMPI 5 before we get around to jll

giordano: If that helps for testing, here's a docker container with openmpi v5.0.2: https://github.com/JuliaParallel/github-actions-buildcache/pkgs/container/github-actions-buildcache/183303325?tag=openmpi-5.0.2-wufrzwmzj3h642rou7vsrphzpysokjui.spack

giordano: For the record, `MPI.jl` tests pass in that container, I'll open a PR to add that to the matrix so that we test OpenMPI v5.  We still need to figure out what was the problem with the JLL.

giordano: @avik-pal with #826 now merged we actually have coverage for OpenMPI v5 on CI (although not with an AMD GPU available...) so you can do a test like
```suggestion
        @static if MPI_LIBRARY == "OpenMPI" && MPI_LIBRARY_VERSION >= v"5"
```
as suggested above by @vchuravy and we can at least ensure the ccall is working correctly.

src/environment.jl
```diff
@@ -341,3 +341,28 @@ function has_cuda()
...
+            # int MPIX_Query_rocm_support(void)
+            return 0 != ccall((:MPIX_Query_rocm_support, libmpi), Cint, ())
```
giordano, line 361: Why not
```suggestion
            return ccall((:MPIX_Query_rocm_support, libmpi), Bool, ())
```
or even
```suggestion
            return @ccall libmpi.MPIX_Query_rocm_support()::Bool
```
?

giordano:
Can also document `has_rocm` in https://juliaparallel.org/MPI.jl/stable/knownissues/#Hints-to-ensure-ROCm-aware-MPI-to-be-functional, like `has_cuda` is documented in the CUDA section.

giordano:
@avik-pal are you interested in moving this forward?

avik-pal:
Completely fell off my radar, I will finish this over the weekend

JBlaschke:
Once this is merged, I can implement `has_gpu`; `has_cuda` and `has_rocm` for Cray MPICH

JBlaschke (approved):

luraess:
Maybe one could add a sentence similar to CUDA (https://github.com/JuliaParallel/MPI.jl/blob/master/docs/src/usage.md) in the ROCm-aware section of the docs?

src/environment.jl
avik-pal: Tested locally that ccall works by bumping the jll compat to 5

```julia
julia> using MPI

julia> MPI.has_cuda()
false

julia> MPI.has_rocm()
false
```

avik-pal:
> Maybe one could add a sentence similar to CUDA ([`master`/docs/src/usage.md](https://github.com/JuliaParallel/MPI.jl/blob/master/docs/src/usage.md?rgh-link-date=2024-06-21T21%3A20%3A49Z))
> 
> ```
> If using OpenMPI, the status of CUDA support can be checked via the [MPI.has_cuda()](https://github.com/JuliaParallel/MPI.jl/blob/master/docs/src/@ref) function.
> ```
> 
> in the ROCm-aware section of the docs?

done

giordano:
Thanks!

giordano:
> Can also document `has_rocm` in https://juliaparallel.org/MPI.jl/stable/knownissues/#Hints-to-ensure-ROCm-aware-MPI-to-be-functional, like `has_cuda` is documented in the CUDA section.

Last documentation request and the we should be good to go ☝️

avik-pal:
Updated the docs

vchuravy:
Can you erase on master?

avik-pal:
> Can you erase on master?

What do you want me to erase?

giordano:
I presume that was meant to be "rebase".

Issue #828, opened 2024-04-26
MPI.jl fails to find homebrew MPI in macos Sonoma runner


miguelbiron:
Hi folks -- as the title says, we're getting the following error in our CI workflow 

https://github.com/Julia-Tempering/Pigeons.jl/actions/runs/8849574516/job/24301884110#step:7:1236

The same happens for openmpi. Not sure if this is an issue on your end or upstream (julia's Libdl).

simonbyrne:
Why not just use the default MPI?

miguelbiron:
Because at some point we did just that, and we realized that many of our users were finding issues that were not being picked up by our CI tests.

miguelbiron:
BTW we basically copied the workflow from you, so we expect that if you rerun your CI you should see the same issues (your last run was 3 days ago and [used the old runners](https://github.com/JuliaParallel/MPI.jl/actions/runs/8807615431/job/24175073871#step:1:4)).

giordano:
Is this because macos-latest now is aarch64-darwin?

sloede:
> Is this because macos-latest now is aarch64-darwin?

I don't know if that's the cause, but yes, they are arm now

giordano:
Would setting macos-13 instead of macos-latest help?

sloede:
That or macos-12 worked for us (on a similar issue, but unrelated to MPI.jl)

miguelbiron:
Testing this right now https://github.com/Julia-Tempering/Pigeons.jl/pull/230

giordano:
We probably still want to understand what's happening, but currently I personally don't have the time to do that, but at least using an older macos image should help with getting a working environment.

miguelbiron:
Ok confirmed that using macos-12 [still works](https://github.com/Julia-Tempering/Pigeons.jl/actions/runs/8851596941/job/24308493240?pr=230#step:7:1236). Per @giordano I'll leave the issue open until a more permanent solution arises.

simonbyrne:
It's failing here as well:
https://github.com/JuliaParallel/MPI.jl/actions/runs/8807615431/job/24305646282

I don't really have the time to look into it at the moment though.

giordano:
[...]

PR #829, opened 2024-04-27
Hotfix: Run failing `macos-latest` tests on `macos-13`

sloede:
Based on reports in #828

### DIFF
--- a/.github/workflows/UnitTests.yml
+++ b/.github/workflows/UnitTests.yml
@@ -140,10 +140,14 @@ jobs:
           - openmpi
         julia_version:
           - "1"
+        os:
+          - macos-latest
+          - macos-13
+          - macos-12
 
       fail-fast: false
 
-    runs-on: macos-latest
+    runs-on: ${{ matrix.os }}
     env:
       JULIA_MPI_TEST_BINARY: system
       ZES_ENABLE_SYSMAN: 1  # https://github.com/open-mpi/ompi/issues/10142
--- a/.github/workflows/UnitTests.yml
+++ b/.github/workflows/UnitTests.yml
@@ -344,6 +344,13 @@ jobs:
         exclude:
           - os: macos-latest
             julia_arch: x86
+        include:
+          - os: macos-latest
+            julia_arch: aarch64
+          - os: macos-13
+            julia_arch: x64
+          - os: macos-12
+            julia_arch: x64
 
       fail-fast: false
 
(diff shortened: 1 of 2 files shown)


sloede:
The two system/Homebrew tests on `macos-latest` fail as well:
* [`test-system-brew (mpich, 1, macos-latest)`](https://github.com/JuliaParallel/MPI.jl/actions/runs/8857338317/job/24324703125?pr=829)
* [`test-system-brew (openmpi, 1, macos-latest)`](https://github.com/JuliaParallel/MPI.jl/actions/runs/8857338317/job/24324703957?pr=829)

Switching them to `macos-13` or `macos-12` seems like a suitable workaround.

Running some more tests to see if [`test-mpitrampoline-jll (macos-latest, mpitrampoline, 1.6, x64)`](https://github.com/JuliaParallel/MPI.jl/actions/runs/8857338317/job/24324700762?pr=829) can be fixed by either going to an older macOS image or by providing the correct architecture.

sloede:
@giordano @simonbyrne This PR provides a hotfix for the failing CI tests. Whether to merge this or not, I'll leave up to you. Unfortunately, I do not have the resources to tackle this issue at the root, so this is only a bandaid.

giordano:
I think the problem is relatively simple: julia just can't find the MPI library because it isn't anywhere accessible.  If I manually set `DYLD_FALLBACK_LIBRARY_PATH` then I can find `libmpi`:
```
$ DYLD_FALLBACK_LIBRARY_PATH=/opt/homebrew/Cellar/mpich/4.2.1/lib julia -q
julia> using Libdl

julia> find_library("libmpi")
"libmpi"
```
I have no clue of how this magically works in older versions of macOS

giordano:
Ah, I think I understand now: up to the `macos-13` image, the `brew` commands used to install packages in `/usr/local` (which is probably a default search path for the dynamic loader), but in `macos-14` they're now installed under `/opt/homebrew`.  I don't know whether this is a change in Homebrew, or macOS, or whatever, but the net result is that `libmpi` isn't anymore in a default search path for the dynamic loader, and so we have to do something #831 to reliably make loading libmpi.  An alternative could be to tell Homebrew to install the packages again in `/usr/local`, but I think setting `DYLD_FALLBACK_LIBRARY_PATH` is slightly more general: we don't have to change the installation directory.

sloede:
> I don't know whether this is a change in Homebrew, or macOS, or whatever

https://apple.stackexchange.com/a/437622

giordano:
Thanks for finding the reference!  That confirms there isn't anything broken in `MPI.jl`, it was only an infrastructure change in the environment :slightly_smiling_face:

PR #836, opened 2024-06-13
[CI] Update MPIWrapper to v2.11.0

giordano:
Ref: https://github.com/JuliaParallel/MPI.jl/pull/834#issuecomment-2165801463


giordano:
Finally all green, I'm going to merge it.  Thanks @eschnett for the tip!

PR #839, opened 2024-06-19
AMDGPU compat

ahojukka5:
Bump compat to latest 0.9 release.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,7 @@ Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
 
 [compat]
 Distributed = "1"
-AMDGPU = "0.5.7, 0.6, 0.7, 0.8"
+AMDGPU = "0.5.7, 0.6, 0.7, 0.8, 0.9"
 CUDA = "3, 4, 5"
 DocStringExtensions = "0.8, 0.9"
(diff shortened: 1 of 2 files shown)


luraess:
Thanks. Could you also bump the compat in the test ˋProject.toml`?

luraess:
Thanks

giordano:
Tests seem broken in Julia v1.6 now.

luraess:
Seems all good now, except the GPU-aware tests which fails (unrelated to this PR as reported by @giordano). Shall we bump and patch release and take care of the GPU CI in another PR?

vchuravy (approved):

vchuravy:
Did someone confirm that this is fine locally?

luraess:
> Did someone confirm that this is fine locally?

I was in the process but getting stuck as on a cluster under SLURM I did not manage to allocate resources that could be spawned by running MPI test suite from within the test suite... This is a recurring issue one is hitting and I did not find a work-around yet. It would be different on a multi-GPU server node where one could just run `mpiexec` without SLURM with all devices visible, etc...

vchuravy:
@ahojukka5 thank you for your PR. I will close this in favor of #844

Issue #841, opened 2024-06-19
AMDGPU tests are failing in allgather


vchuravy:
Going as far back as https://github.com/JuliaParallel/MPI.jl/pull/764#issuecomment-1708920070
x-ref: https://github.com/JuliaParallel/MPI.jl/pull/839#issuecomment-2179336955

Might be as simple as updating the OpenMPI version we are using for the tests?
@luraess can you take a look?

```
698] signal (11.2): Segmentation fault
in expression starting at /var/lib/buildkite-agent/builds/amdgpu1-luraess-com/julialang/mpi-dot-jl/test/test_allgather.jl:9
unknown function (ip: 0x77ae2d2bb815)
opal_convertor_pack at /var/lib/buildkite-agent/builds/gpuci-2/julialang/mpi-dot-jl/openmpi/lib/libopen-pal.so.40 (unknown line)
mca_btl_vader_sendi at /var/lib/buildkite-agent/builds/gpuci-2/julialang/mpi-dot-jl/openmpi/lib/openmpi/mca_btl_vader.so (unknown line)
mca_pml_ob1_send_inline.constprop.0 at /var/lib/buildkite-agent/builds/gpuci-2/julialang/mpi-dot-jl/openmpi/lib/openmpi/mca_pml_ob1.so (unknown line)
mca_pml_ob1_send at /var/lib/buildkite-agent/builds/gpuci-2/julialang/mpi-dot-jl/openmpi/lib/openmpi/mca_pml_ob1.so (unknown line)
ompi_coll_base_sendrecv_actual at /var/lib/buildkite-agent/builds/gpuci-2/julialang/mpi-dot-jl/openmpi/lib/libmpi.so (unknown line)
ompi_coll_base_allgather_intra_two_procs at /var/lib/buildkite-agent/builds/gpuci-2/julialang/mpi-dot-jl/openmpi/lib/libmpi.so (unknown line)
ompi_coll_tuned_allgather_intra_dec_fixed at /var/lib/buildkite-agent/builds/gpuci-2/julialang/mpi-dot-jl/openmpi/lib/openmpi/mca_coll_tuned.so (unknown line)
MPI_Allgather at /var/lib/buildkite-agent/builds/gpuci-2/julialang/mpi-dot-jl/openmpi/lib/libmpi.so (unknown line)
[...]

PR #840, opened 2024-06-19
CUDA CI no longer has version tag

vchuravy:
Since https://github.com/JuliaGPU/buildkite/commit/ed35a7903835c0ee89493898821eca0ec38efefb CUDA agent no longer have a version tag.

PR #844, opened 2024-06-20
Fix ROCm CI

luraess:
Address #841 
Superseed #839 wrt AMDGPU compat

Getting ROCm CI back on track

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -150,6 +150,8 @@
         setup:
           version:
             - "1.10"
+      concurrency: 1
+      concurrency_group: mpi_rocm
       plugins:
         - JuliaCI/julia#v1:
             version: "{{matrix.version}}"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -7,9 +7,9 @@
         queue: "juliagpu"
         cuda: "*"
       env:
-        OPENMPI_VER: "5.0"
-        OPENMPI_VER_FULL: "5.0.3"
-        UCX_VER: "1.17.0"
+        OPENMPI_VER: "4.1"
+        OPENMPI_VER_FULL: "4.1.4"
+        UCX_VER: "1.12.1"
         CCACHE_DIR: "/root/ccache"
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -217,6 +186,7 @@
             '
 
         echo "+++ Run tests"
+        export JULIA_MPI_TEST_EXCLUDE="test_allreduce.jl,test_reduce.jl,test_scan.jl"
```
luraess, line 189: Need to investigate why these tests fail on AMDGPU backend. Maybe @pxl-th has an idea?

luraess:
AMDGPU tests pass now having updated the Pipeline to @vchuravy suggestion in #840 and upon using latest OpenMPI and UCX:
```
OPENMPI_VER: "5.0"
OPENMPI_VER_FULL: "5.0.3"
UCX_VER: "1.17.0"
```

The only failing tests are `test_allreduce.jl,test_reduce.jl,test_scan.jl` which I for now excluded using the ENV var mechanism.

CUDA tests still fail though.

vchuravy (approved):

luraess:
@vchuravy all CUDA tests are failing

luraess:
Also, CUDA Buildkite workflows (builds and compilation during test) are running close to an order of magnitude slower compared to ROCm ones

vchuravy:
They are passing on main https://buildkite.com/julialang/mpi-dot-jl/builds/1451

They slow-down is very weird and it looks like things slow down when things are running parallel on the same machine.

luraess:
With concurrency set to 1 test should run serially?

Why test pass on master and not here as you merged master in this branch?

luraess:
> Why test pass on master and not here as you merged master in this branch?

So, using latest OpenMPI and UCX as for ROCm in CUDA Buildkite CI segfaults. Rolling back to versions as on `master` fixes it.

Tests now pass (with exception of `test_allreduce.jl,test_reduce.jl,test_scan.jl` for ROCm), and codecov seems complaining about `changes` and `project`.

PR #850, opened 2024-08-02
Fix typos

lcw:


src/environment.jl
```diff
@@ -287,7 +287,7 @@ It is unaffected by [`MPI.Finalize`](@ref), and is one of the few functions that
 called before [`MPI.Init`](@ref).
 
 # External links
-$(_doc_external("MPI_Intialized"))
+$(_doc_external("MPI_Initialized"))
```
giordano, line 290: Ugh, maybe we should have a check for the documentation that links aren't broken 😕

lcw: Is there a option in Documenter.jl for that?

giordano: Apparently yes: #856.  But MPICH/OpenMPI websites aren't reliable for actually documenting all functions... So the reality is that we do have a large number of broken links 😞

PR #854, opened 2024-08-06
link updates

spaette:
a replacement pull for #852 (Closed)

### DIFF
--- a/docs/examples/05-job_schedule.jl
+++ b/docs/examples/05-job_schedule.jl
@@ -3,8 +3,10 @@
 # assigns one element to each worker to compute the operation.
 # When the worker is finished, the root sends another element
 # until each element is added 100
-# Inspired on
-# https://www.hpc.ntnu.no/ntnu-hpc-group/vilje/user-guide/software/mpi-and-mpi-io-training-tutorial/basic-mpi/job-queue
+# Inspired by https://www.hpc.ntnu.no/vilje/software/mpi-and-mpi-io-training-tutorial/
+# https://www.hpc.ntnu.no/vilje/software/mpi-and-mpi-io-training-tutorial/basic-mpi/job-queue/
+# the updated job_queue.c file is available in the extracted .tar.gz 04_job_queue subdirectory
+# wget https://www.hpc.ntnu.no/wp-content/uploads/2019/09/mpiexamples.tar.gz
 
 using MPI
 
--- a/docs/examples/05-job_schedule.jl
+++ b/docs/examples/05-job_schedule.jl
@@ -5,8 +5,8 @@
 # until each element is added 100
 # Inspired by https://www.hpc.ntnu.no/vilje/software/mpi-and-mpi-io-training-tutorial/
 # https://www.hpc.ntnu.no/vilje/software/mpi-and-mpi-io-training-tutorial/basic-mpi/job-queue/
-# the updated job_queue.c file is available in the extracted .tar.gz 04_job_queue subdirectory
-# wget https://www.hpc.ntnu.no/wp-content/uploads/2019/09/mpiexamples.tar.gz
+# an updated job_queue.c is available in the basic_mpi/04_job_queue/src subdirectory of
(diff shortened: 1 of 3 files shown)


docs/examples/05-job_schedule.jl
```diff
@@ -3,8 +3,8 @@
 # assigns one element to each worker to compute the operation.
 # When the worker is finished, the root sends another element
 # until each element is added 100
-# Inspired on
-# https://www.hpc.ntnu.no/ntnu-hpc-group/vilje/user-guide/software/mpi-and-mpi-io-training-tutorial/basic-mpi/job-queue
+# Inspired on https://www.hpc.ntnu.no/vilje/software/mpi-and-mpi-io-training-tutorial/
```
giordano, line 6:
```suggestion
# Inspired by https://www.hpc.ntnu.no/vilje/software/mpi-and-mpi-io-training-tutorial/
```

docs/examples/05-job_schedule.jl
```diff
@@ -3,8 +3,8 @@
...
+# wget https://www.hpc.ntnu.no/wp-content/uploads/2019/09/mpiexamples.tar.gz
```
giordano, line 7: What is this for?

spaette: > What is this for?

see text of the closed pull for my initial reasoning and the `tree` command output

the pertinent reference files are not factually viewable online

`https://www.hpc.ntnu.no/vilje/software/mpi-and-mpi-io-training-tutorial/`

under **Basic MPI** the hyperlink to the text  Job Queue points to a page which doesn't  seem to exist

and @rafabench in `#449` clearly wanted to add a reference

giordano: > see text of the closed pull for my initial reasoning and the tree command output

I can't say that was particularly clear to me.  How about linking https://www.hpc.ntnu.no/vilje/software/mpi-and-mpi-io-training-tutorial/basic-mpi/job-queue/ instead?  That seems to be the up-to-date page you actually wanted to link.

spaette: > clear

it might have been indicated to at least identifiy the extracted .tar.gz subdirectory with the pertinent files

thank you for tracking down the current webpage address

the therein rendered `src/job_queue.c` file's `#include` lines don't have headers

and the page's hyperlink on the text code (obtain all the example[s]) is wrong

my preference is to take time with this

webpage has more info than just the code

the .tar.gz extraction has more subdir pertinent files than just the webpage's rendered .c file

spaette: I left the top level webpage in place for contextual reasons.

Is the revision clear at this point or do you suggest any other modifications?

giordano: I like the clarification above, but I still don't like the random wget line: it isn't super clear what's for, and wget isn't even necessarily a good suggestion, since it isn't available by default in many systems. I'd say that continuing the text above with something like 
```suggestion
# of https://www.hpc.ntnu.no/wp-content/uploads/2019/09/mpiexamples.tar.gz
```
would work for me.

spaette: I guess this newest revision is concise.

Incidentally in `.github/workflows/UnitTests.yml` there are two `wget` lines.

giordano: > Incidentally in `.github/workflows/UnitTests.yml` there are two `wget` lines.

I'm not sure how that's any relevant here:

1. it isn't user-facing documentation
2. that's for running code in controlled environment where we do know `wget` is available

spaette:
@sjalander

This PR in docs/examples/05-job_schedule.jl updated a couple of NTNU webpage addresses and a download address.

The `cat` command on the 2nd webpage outputs a file which is not identical to that of the extracted .tar.gz download.

from the expanded resolved comments to this pull

> the therein rendered `src/job_queue.c` file's `#include` lines don't have headers

Running a `diff` on the copied webpage's rendered file and the download's job_queue.c may reveal other changes.

Hope I'm not imposing in sending a notice, and if interested on your end you can forward it to the website maintainer.

PR #481, opened 2021-07-04
[mpiexecjl] Accept `JULIA_BINDIR` as environment variable

giordano:
[`JULIA_BINDIR`](https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_BINDIR)
is a standard Julia environment variable pointing to the directory where the
`julia` binary is present.

Fix #476.  CC: @ViralBShah.

PR #858, opened 2024-08-06
Set `JULIA_BINDIR` when running `mpiexecjl` tests

### LINKED ISSUE #857: `mpiexecjl` exit code test failing
I have a system where the test introduced in #834 is failing:
```
MPIPreferences:
  binary:  OpenMPI_jll
  abi:     OpenMPI

Package versions
  MPI.jl:             0.20.20
  MPIPreferences.jl:  0.1.11
  OpenMPI_jll:        4.1.6+0

Library information:
  libmpi:  /home/cceamgi/.julia/artifacts/58dcf187642cdfbafb3581993ca3d8de565acc78/lib/libmpi.so
  libmpi dlpath:  /home/cceamgi/.julia/artifacts/58dcf187642cdfbafb3581993ca3d8de565acc78/lib/libmpi.so
  MPI version:  3.1.0
  Library version:
    Open MPI v4.1.6, package: Open MPI sabae@amdci7.julia.csail.mit.edu Distribution, ident: 4.1.6, repo rev: v4.1.6, Sep 30, 2023
Hello world, I am rank 3 of 4
Hello world, I am rank 2 of 4
Hello world, I am rank 0 of 4
Hello world, I am rank 1 of 4
mpiexecjl: Test Failed at /home/cceamgi/.julia/packages/MPI/is7GN/test/mpiexecjl.jl:41
  Expression: p.exitcode == exit_code
   Evaluated: 1 == 10
```
I need to investigate what's wrong with this.  For the record, this isn't specific to OpenMPI_jll, I see the same with MPICH_jll.  I wonder if the problem is the shell, here `/bin/sh` is
```console
$ /bin/sh --version
GNU bash, version 5.1.8(1)-release (aarch64-redhat-linux-gnu)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

giordano:
I don't have a good explanation of why this is necessary, but setting `JULIA_BINDIR`, as it was done in #481 for the other run in the same test set, fixes #857 for me.  My only hypothesis is that this somehow helps Julia find its own libraries, perhaps especially if using juliaup, but honestly I can't tell what's failing without setting `JULIA_BINDIR`.  Also, note that that I don't have problems using `mpiexecjl` normally, the issue reported in #857 appeared only when running the tests.

### DIFF
--- a/test/mpiexecjl.jl
+++ b/test/mpiexecjl.jl
@@ -15,25 +15,21 @@ using MPI
         # Test a run of mpiexec
         nprocs_str = get(ENV, "JULIA_MPI_TEST_NPROCS", "")
         nprocs = nprocs_str == "" ? clamp(Sys.CPU_THREADS, 2, 4) : parse(Int, nprocs_str)
-        mpiexecjl = joinpath(dir, "mpiexecjl")
+        env = ["JULIA_BINDIR" => Sys.BINDIR]
+        mpiexecjl = addenv(`$(joinpath(dir, "mpiexecjl"))`, env...)
         # `Base.julia_cmd()` ensures keeping consistent flags when running subprocesses.
         julia = Base.julia_cmd()
         example = joinpath(@__DIR__, "..", "docs", "examples", "01-hello.jl")
-        env = ["JULIA_BINDIR" => Sys.BINDIR]
-        p = withenv(env...) do
-            run(`$(mpiexecjl) -n $(nprocs) --project=$(dir) $(julia) --startup-file=no -q $(example)`)
-        end
+        p = run(`$(mpiexecjl) -n $(nprocs) --project=$(dir) $(julia) --startup-file=no -q $(example)`)
         @test success(p)
         # Test help messages
         for help_flag in ("-h", "--help")
-            help_message = withenv(env...) do
-                read(`$(mpiexecjl) --project=$(dir) --help`, String)
-            end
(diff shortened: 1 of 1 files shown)


test/mpiexecjl.jl
```diff
@@ -15,25 +15,21 @@ using MPI
         # Test a run of mpiexec
         nprocs_str = get(ENV, "JULIA_MPI_TEST_NPROCS", "")
         nprocs = nprocs_str == "" ? clamp(Sys.CPU_THREADS, 2, 4) : parse(Int, nprocs_str)
-        mpiexecjl = joinpath(dir, "mpiexecjl")
+        env = ["JULIA_BINDIR" => Sys.BINDIR]
+        mpiexecjl = addenv(`$(joinpath(dir, "mpiexecjl"))`, env...)
```
giordano, line 19: I'm still not sure about why `JULIA_BINDIR` is _needed_, but I think that ensuring `mpiexecjl` is consistently called with the same environment in all these tests is a good change: it's less error-prone as it avoids missing something in possible future tests.  As a side effect, we can slightly simplify the code in these tests whenever we call `mpiexecjl`.

PR #856, opened 2024-08-06
[docs] Build with `linkcheck=true`

giordano:

PR #859, opened 2024-08-06
Fix autogenerated links to OpenMPI/MPICH docs

giordano:
OpenMPI docs are hosted on a new website, and the docs for all the `*_c` functions are not available for that this MPI library.  Also, MPICH docstrings are available at the moment for most functions only in the v4.0 website, not v4.1+.  Extracted out of #856.

### DIFF
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -19,9 +19,27 @@ end
 
 
 function _doc_external(fname)
-"""
-- `$fname` man page: [OpenMPI](https://www.open-mpi.org/doc/current/man3/$fname.3.php), [MPICH](https://www.mpich.org/static/docs/latest/www3/$fname.html)
-"""
+    # Ideally we'd like to use the "latest" version of the docs, but MPICH v4.0
+    # at the moment seems to be the last version with all the docstrings.
+    mpich = "[MPICH](https://www.mpich.org/static/docs/v4.0/www3/$(fname).html)"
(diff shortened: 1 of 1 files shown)


src/MPI.jl
```diff
@@ -19,9 +19,27 @@ end
...
-"""
+    # Ideally we'd like to use the "latest" version of the docs, but MPICH v4.0
+    # at the moment seems to be the last version with all the docstrings.
+    mpich = "[MPICH](https://www.mpich.org/static/docs/v4.0/www3/$(fname).html)"
+    # All the *_c functions are undocumented in OpenMPI website
+    if !endswith(string(fname), "_c")
+        openmpi = "[OpenMPI](https://docs.open-mpi.org/en/main/man-openmpi/man3/$(fname).3.html)"
+        # Few docstrings aren't available for MPICH
+        if fname in (:MPI_Status_c2f, :MPI_Status_f2c) || startswith(string(fname), "MPI_Type_create_f90_")
+            return """
```
giordano, line 30: I'm not a fan of all these super ad-hoc checks, but the alternative is to have broken links 🤷

giordano:
I sampled the links for a few docstrings in the [preview of this PR](https://juliaparallel.org/MPI.jl/previews/PR859/reference/api/) and they are all working as expected. I'm going to merge this so that at least we have no more broken links, but if someone knows better and more reliable links to use we can always fix them afterwards.

PR #896, opened 2025-03-31
add note about CUDA.jl own use of threads

vchuravy:


docs/src/knownissues.md
```diff
@@ -145,6 +145,26 @@ before calling `mpiexec`.
 
 ## CUDA-aware MPI
 
+[CUDA.jl](https://github.com/JuliaGPU/CUDA.jl) starts its own internal threads, thus the notes above apply.
```
sloede, line 148:
```suggestion
[CUDA.jl](https://github.com/JuliaGPU/CUDA.jl) starts its own internal threads, thus the notes in the section about [Multi-threading and signal handling](@ref) apply.
```

PR #901, opened 2025-04-29
Allow CI to pass even if the `processcoverage` step fails

DilumAluthge:


.github/workflows/UnitTests.yml
```diff
@@ -71,6 +71,8 @@ jobs:
 
     - uses: julia-actions/julia-runtest@v1
     - uses: julia-actions/julia-processcoverage@v1
+      # Allow CI to pass even if the `processcoverage` step fails:
+      continue-on-error: true
     - uses: codecov/codecov-action@v5
```
giordano, line 76: Also here

DilumAluthge: IIRC, the default behavior for `codecov-action` is to not fail CI on a Codecov error: https://github.com/codecov/codecov-action/blob/ad3126e916f78f00edff4ed0317cf185271ccc2d/action.yml#L48-L51

So I don't think we need `continue-on-error` for the `codecov-action` step.

giordano: I'd prefer not to rely on defaults, but whatever.

giordano (approved):

PR #899, opened 2025-04-29
CI: Update unit tests from `ubuntu-20.04` to `ubuntu-24.04`

DilumAluthge:
GitHub retired their Ubuntu 20 image (for GitHub-hosted runners) on April 15, 2025. Therefore, we are forced to update.

https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/

PR #907, opened 2025-04-29
Add a test checking number of ranks is correct

### LINKED ISSUE #905: Test idea: check number of ranks
We should probably have a test to check that the number of ranks of a job corresponds to the expected number.  

A possible way to implement the test is:

* gather the comm sizes and ranks and do some consistency checks (e.g. length of these vectors is same as comm size and all ranks are there)
* in https://github.com/JuliaParallel/MPI.jl/blob/042ddc2b96c858d7a83cc0fd6b0fcf230b2d0548/test/runtests.jl#L81 set an env var with the number of expected processes that we can check against in the collected ranks and sizes

This came up in #899 where a broken mpich build was causing MPI jobs to have the wrong number or ranks.

giordano:
Fix #905.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -78,7 +78,11 @@ istest(f) = endswith(f, ".jl") && startswith(f, "test_") && !in(f, excludefiles)
 testfiles = sort(filter(istest, readdir(testdir)))
 
 @testset "$f" for f in testfiles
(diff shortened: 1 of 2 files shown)


giordano:
This wouldn't have caught the issues in #899 because a previous test was failing before this one, but good to have a consistency test anyway, I guess.

PR #908, opened 2025-05-07
GitHub Actions CI: Increase the timeouts (from 20 minutes to 40 minutes) for the `test-mpitrampoline-jll` and `test-system-brew` jobs

DilumAluthge:
These jobs are currently timing out on master. See e.g. #898.

Let's see if they pass once we increase the timeout from 20 minutes to 40 minutes.

### DIFF
--- a/.github/workflows/UnitTests.yml
+++ b/.github/workflows/UnitTests.yml
@@ -135,7 +135,7 @@ jobs:
     - uses: julia-actions/julia-runtest@v1
 
   test-system-brew:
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Hmmm. Why isn't GHA CI running on this PR?

DilumAluthge:
Ah, there it is.

giordano (changes requested):
The jobs you indicated are ***hanging***, increasing the timeout has the only result of wasting more time.

PR #914, opened 2025-09-09
oneAPI-aware MPI

michel2323:
I can only test this on Aurora at ANL. I could provide an MPICH_jll with support. How should I proceed?

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -305,6 +305,11 @@
         MPIR_CVAR_GPU_FAST_COPY_MAX_SIZE: 0
         MPIR_CVAR_GPU_FAST_COPY_MAX_SIZE_H2D: 0
         MPIR_CVAR_GPU_FAST_COPY_MAX_SIZE_D2H: 0
+        # MPICH defaults to legacy command queues for its ZE copies, a path
+        # recent compute runtimes barely exercise (SYCL and oneAPI.jl submit
+        # via immediate command lists); the legacy path intermittently corrupts
+        # small host<->device copies on this device
+        MPIR_CVAR_GPU_USE_IMMEDIATE_COMMAND_LIST: 1
       commands: |
         echo "--- Install packages"
         # libdrm-dev pulls libpciaccess into the build container, where MPICH's
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -266,7 +266,9 @@
 
     - wait
 
-    - label: "Tests -- Julia {{matrix.version}}"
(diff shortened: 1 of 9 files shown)


michel2323:
I have no idea what is failing here. Anyhow, before I add `MPI.has_oneapi()`, I want to leave all this to the user. It works on Aurora, so there's that. Can we merge this?

michel2323:
I've now tried building MPICH with `ze` support, and I'm running into all sorts of issues I shouldn't have to.

I've added a `has_oneapi()` method, which returns true only if the environment variable is explicitly set. At this point, I'd expect the user to explicitly want to be in for a ride. I'll discuss an MPICH build with the MPICH team here at ANL.

michel2323:
@lcw, builds are currently failing on hardware without native FP64 support, such as the integrated GPU used in our Buildkite oneAPI runner. The MPICH OpenCL implementation has extensive FP64 dependencies; when ocloc attempts to compile these for non-FP64 devices, it triggers a segfault. This persists even when applying ocloc emulation flags. 

I don't have a Intel GPU with FP64 support that I could add to buildkite. Given these hardware constraints, I suggest we merge this as an experimental feature for now.

michel2323:
@lcw I fiddled with the buildkite runner. The oneAPI.jl tests seem to pass now. Can you take a look? Would this be sufficient to get this merged?

lcw:
@michel2323 Can you please point me where to look to see them passing? I still see failing tests for `buildkite/mpi-dot-jl/oneapi-tests-julia-1-dot-10` and `buildkite/mpi-dot-jl/oneapi-tests-julia-1-dot-12`.

michel2323:
@lcw Sorry about that. The single tests showed as passed intermittently before then flipping to fail.

I compile MPICH 5.x on the oneAPI Buildkite runner, and this revealed a bug for that particular driver (i915). More recent Intel GPUs use the Xe driver, where I cannot reproduce this issue. Most tests pass, however, the issue is non-deterministic and can affect any test. So I marked the oneAPI.jl tests as `softfail`.

On Aurora, the tests pass, and it would be great to have this in a released MPI.jl. I could set up a runner that uses the ALCF CI at Argonne through GitLab, as I did for oneAPI.jl.

What do you recommend?

lcw:
Thanks for all your effort! I'll merge this and make a release.

michel2323:
@lcw Oh, no worries. As Claude would say, "I liked your pushback" with the runners, etc. 

Just to be clear, the tests are not hot. They fail silently until the issue with MPICH and the i915 on the Buildkite runner is fixed.

Let me know if you want anything changed.

PR #939, opened 2026-03-18
Outline error path

vchuravy:
Manually outline the error path, otherwise on Julia 1.10 we cause an unecessary GC-frame for every MPI function.

### DIFF
--- a/src/api/api.jl
+++ b/src/api/api.jl
@@ -122,6 +122,11 @@ function Base.show(io::IO, err::FeatureLevelError)
     print(io, "FeatureLevelError($(err.function_name)): Minimum MPI version is $(err.min_version)")
 end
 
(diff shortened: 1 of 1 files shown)


giordano (approved):

src/api/api.jl
```diff
@@ -134,7 +139,13 @@ macro mpichk(expr, min_version=nothing)
 
     expr = macroexpand(@__MODULE__, :(@mpicall($expr)))
     # MPI_SUCCESS is defined to be 0
-    :((errcode = $(esc(expr))) == 0 || throw(MPIError(errcode)))
+    quote
+        errcode = $(esc(expr))
```
giordano, line 143: Maybe this should be
```suggestion
        local errcode = $(esc(expr))
```
?

PR #941, opened 2026-03-28
Refactor MPIPtr cconvert to support Julia 1.14

nhz2:
Alternative to #935

CC @eschnett @giordano @vchuravy


eschnett:
One of the main features of `ccall` and its automatic conversion to pointers is that it's GC safe, i.e. that it holds on to the respective Julia objects while a pointer to the object is in use. I didn't follow the logic that you're introducing, but since this kind of error (losing a GC root) is very difficult to diagnose, I want to recommend to add respective comments to the code, stating which part of the machinery is holding on to objects, which functions are called before objects are converted to pointers, and which are called on pointers.

nhz2:
Bump

KristofferC:
Bump, would be nice to have this fixed (+ new release if possible).

giordano:
I believe we'd need some help with the review, none of us is particularly familiar with what's the intended way to go about it. I definitely prefer this kind of approach over #935 because that felt too ad hoc, but the need for CConvWrapper here is also odd.

lcw:
We need to support OpenAPI as well.

nhz2:
> We need to support OpenAPI as well.

What do you mean by this?

lcw:
> > We need to support OpenAPI as well.
> 
> What do you mean by this?

I think we might need to change these lines as well <https://github.com/JuliaParallel/MPI.jl/blob/61cc82efd7735b0b290fb406a3498fed28e6ac9e/ext/OneAPIExt.jl#L7-L21>

nhz2:
I've updated OneAPIExt.jl and added an intermediate pointer type trait function `mpi_ptr_type`.

eschnett:
There is an error in the documentation. Can you add `MPI.mpi_ptr_type` to the `@docs` block in `buffers.md`? This should make the docs build again.

Alternatively, if this function should be private and not be mentioned in the docs, then convert its docstring to a plain comment.

After this change I think this is good to go. I'm really looking forward to see `nightly` pass CI again!

nhz2:
I've made the function private.

vchuravy (approved):

vchuravy:
@nhz2 thank you for all your work!

PR #945, opened 2026-04-14
Correct MPIABI constants (don't use @const_ref)

eschnett:


eschnett:
This corrects an error with the new MPIABI ABI; I used `@const_ref` instead of `const` when declaring certain compile-time constants. I want to merge this although CI fails all nightly builds. I think all nightly builds fail all the time anyway.

PR #949, opened 2026-04-30
Update GPU-aware MPI docs

abussy:
Suggestion to update the GPU-aware MPI docs to include explicit device synchronization. According to this [slack thread](https://julialang.slack.com/archives/C689Y34LE/p1777468536637869), this is a known issue, but it is not yet documented.

### DIFF
--- a/docs/src/usage.md
+++ b/docs/src/usage.md
@@ -115,6 +115,20 @@ your ROCm-aware MPI implementation to use multiple AMD GPUs (one GPU per rank).
 If using OpenMPI, the status of ROCm support can be checked via the
 [`MPI.has_rocm()`](@ref) function.
 
+### Safe MPI communication
+When using GPU-aware MPI (CUDA or ROCm), it is recommended to synchronize the device before initiating MPI communication. Because GPU operations are asynchronous with respect to the host, GPU buffers may not be fully updated when a MPI call is issued. This can lead to race conditions and incorrect results.
+
+With CUDA:
+```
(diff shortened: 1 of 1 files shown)


docs/src/usage.md
```diff
@@ -115,6 +115,20 @@ your ROCm-aware MPI implementation to use multiple AMD GPUs (one GPU per rank).
 If using OpenMPI, the status of ROCm support can be checked via the
 [`MPI.has_rocm()`](@ref) function.
 
+### Safe MPI communication
+When using GPU-aware MPI (CUDA or ROCm), it is recommended to synchronize the device before initiating MPI communication. Because GPU operations are asynchronous with respect to the host, GPU buffers may not be fully updated when a MPI call is issued. This can lead to race conditions and incorrect results.
```
vchuravy:
```suggestion
When using GPU-aware MPI (CUDA or ROCm), it is required to synchronize the (task-local) stream, before initiating MPI communication. Because GPU operations are asynchronous with respect to the host, GPU buffers may not be fully updated when a MPI call is issued. This can lead to race conditions and incorrect results.
```

vchuravy:
Fixes: #443
 x-ref https://github.com/JuliaParallel/MPI.jl/issues/443#issuecomment-1141690517

Looking at some of the old discussions on this, was a reference I used before.

https://ieeexplore.ieee.org/abstract/document/6650987 

Also: https://github.com/open-mpi/ompi/issues/7733#issuecomment-648343243

PR #955, opened 2026-07-17
Add `gc_safe` option to `@mpicall` and `@mpichk` macros

### LINKED ISSUE #954: Use `@ccall gc_safe=true` for blocking calls
We should use https://github.com/JuliaGPU/GPUToolbox.jl/blob/f971bcdeb2cd8096333549e6563cf03a76459ae2/src/ccalls.jl#L110 for Wait and allreduce

giordano:
Note: the diff is quite large because the file with the generated API changed a lot.  I split the changes into different logically consistent commits, for easier review.  I also got the bot to always generate from now on the file with a stable sorting, instead of relying on the order of the functions in the upstream header file (which isn't really stable).

@vchuravy for the time being I marked only `MPI_Wait` and `MPI_Allreduce`, do we need to support more functions?

Fix #954.

### DIFF
--- a/docs/src/knownissues.md
+++ b/docs/src/knownissues.md
@@ -143,6 +143,21 @@ export UCX_ERROR_SIGNALS="SIGILL,SIGBUS,SIGFPE"
 ```
 before calling `mpiexec`.
 
+## Multi-threading and garbage collection
+
+On Julia v1.12 and later, blocking MPI calls (e.g. `MPI_Wait`, `MPI_Barrier`, blocking point-to-point communications and collectives) are marked as *GC-safe*: the Julia garbage collector running on one thread does not have to wait for other threads which are inside one of these MPI calls.
+This prevents the whole process from stalling — or deadlocking outright — when a thread blocks in MPI while waiting on the progress of other ranks and another thread needs to run a garbage collection.
+
+A consequence of this is that in a multi-threaded program the garbage collector can now run *concurrently* with a blocking MPI call, and its finalizers can free MPI handles (communicators, requests, datatypes, operators, etc.) by calling functions like `MPI_Comm_free` while another thread is still inside MPI.
+Concurrent MPI calls from different threads are only permitted when MPI is initialized with the `MPI.THREAD_MULTIPLE` [`MPI.ThreadLevel`](@ref), whereas [`MPI.Init`](@ref) defaults to `:serialized`.
+Multi-threaded programs should therefore initialize MPI with
+
+```julia
+MPI.Init(; threadlevel=:multiple)
+```
+
+Single-threaded programs are unaffected.
+
 ## CUDA-aware MPI
 
 ### Memory pool
--- a/gen/src/MPIgenerator.jl
+++ b/gen/src/MPIgenerator.jl
@@ -58,13 +58,11 @@ module MPIgenerator
         # through `@cfunction` callbacks (e.g. custom reduction operators), which
         # are safe in `gc_safe` regions
         gc_safe = (
-            # request completion and probes
+            # request completion
             :MPI_Wait,
             :MPI_Waitall,
             :MPI_Waitany,
             :MPI_Waitsome,
-            :MPI_Probe,
-            :MPI_Mprobe,
             :MPI_Mrecv,
             # blocking point-to-point
             :MPI_Send,
--- a/gen/src/MPIgenerator.jl
+++ b/gen/src/MPIgenerator.jl
@@ -18,7 +18,9 @@ module MPIgenerator
         mkpath(out)
 
         options = load_options(joinpath(@__DIR__, "generator.toml"))  # wrapper generator options
-        options["general"]["callback_documentation"] = node -> [string('$', "(_doc_external(:", node.id, "))")]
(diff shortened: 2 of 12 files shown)


gen/src/MPIgenerator.jl
```diff
@@ -48,6 +50,13 @@ module MPIgenerator
             :MPI_Wtick,
         )
 
+        # these methods are called with `gc_safe=true`, to allow the garbage
+        # collector to run concurrently with the (potentially blocking) MPI call
+        gc_safe = (
+            :MPI_Wait,
+            :MPI_Allreduce,
+        )
```
giordano, line 97: > @vchuravy for the time being I marked only `MPI_Wait` and `MPI_Allreduce`, do we need to support more functions?

The bot suggests
```julia
gc_safe = (
    # request completion and probes
    :MPI_Wait, :MPI_Waitall, :MPI_Waitany, :MPI_Waitsome,
    :MPI_Probe, :MPI_Mprobe, :MPI_Mrecv,
    # blocking point-to-point
    :MPI_Send, :MPI_Ssend, :MPI_Recv, :MPI_Sendrecv, :MPI_Sendrecv_replace,
    # blocking collectives
    :MPI_Barrier, :MPI_Bcast,
    :MPI_Gather, :MPI_Gatherv, :MPI_Scatter, :MPI_Scatterv,
    :MPI_Allgather, :MPI_Allgatherv, :MPI_Alltoall, :MPI_Alltoallv, :MPI_Alltoallw,
    :MPI_Reduce, :MPI_Allreduce, :MPI_Reduce_scatter, :MPI_Reduce_scatter_block,
    :MPI_Scan, :MPI_Exscan,
    # blocking neighborhood collectives
    :MPI_Neighbor_allgather, :MPI_Neighbor_allgatherv,
    :MPI_Neighbor_alltoall, :MPI_Neighbor_alltoallv, :MPI_Neighbor_alltoallw,
)
```

vchuravy: The only thing I am worried about is the `MPI_Probe` since they should be very fast.

giordano: The bot really insisted that probes are fine

> - `MPI_Mrecv` — the one I nearly cut. After `MPI_Mprobe` the message is matched, so it "should" be a bounded copy. But matching only guarantees the envelope arrived: for a large rendezvous message the payload transfer happens inside `MPI_Mrecv`, and it needs the sender's MPI library to answer the clear-to-send. MPI implementations don't guarantee asynchronous progress — if the sender rank is off in compute code, that answer waits until it re-enters MPI. Unbounded tail. Keep.
>
> [...]
>
> `MPI_Probe` is one of the strongest candidates on the list — it sits at the opposite end of the spectrum from `MPI_Mrecv`.
>
> On the "worth it" test: its entire job is to block until a matching message envelope arrives, i.e. until a peer rank posts the corresponding send. That's a pure peer-progress wait with no local-work bound at all — there isn't even a fast typical case to speak of, since code reaches for `MPI_Probe` precisely when it doesn't know when (or how big) the message will be. A rank sitting in `MPI_Probe` waiting for a slow peer while holding up GC on every other thread is exactly the motivating pathology.
>
> On the safety test: it's about as inert as an MPI call gets — it doesn't touch user buffers at all, writes only a bits-type `MPI_Status`, receives nothing, and has no callback path (unlike `MPI_Wait`, it doesn't complete other pending operations' reduction ops; it only inspects the matching queue and drives progress internally).
>
> So: keep, with confidence. Same verdict for `MPI_Mprobe`, which is the same wait with a matched-message handle as output.

In any case I don't have a need for marking these functions as gc-safe, so I dropped them.

docs/src/knownissues.md
```diff
@@ -143,6 +143,21 @@ export UCX_ERROR_SIGNALS="SIGILL,SIGBUS,SIGFPE"
...
+A consequence of this is that in a multi-threaded program the garbage collector can now run *concurrently* with a blocking MPI call, and its finalizers can free MPI handles (communicators, requests, datatypes, operators, etc.) by calling functions like `MPI_Comm_free` while another thread is still inside MPI.
+Concurrent MPI calls from different threads are only permitted when MPI is initialized with the `MPI.THREAD_MULTIPLE` [`ThreadLevel`](@ref), whereas [`MPI.Init`](@ref) defaults to `:serialized`.
+Multi-threaded programs should therefore initialize MPI with
+
+```julia
+MPI.Init(; threadlevel=:multiple)
+```
+
+Single-threaded programs are unaffected.
+
```
giordano, line 160: @vchuravy this was entirely the bot's initiative, I'm not quite sure whether it's true that multi-threaded applications need
```julia
MPI.Init(; threadlevel=:multiple)
```
to make the GC run concurrently with the MPI functions.  I'm happy to remove this if it's garbage (pun intended).

vchuravy: Oh that is fun... it is an interesting question what happens when we run `MPI_*_free` from a finalizer...

vchuravy: I think the problem is that Julia these days becomes multi-threaded as soon as you do `using CUDA` so you don't know that you are serial.

Or even today just starting Julia will give me an interactive and a worker thread. So we are already in a world where the user can't reason about being single-threaded

PR #956, opened 2026-07-21
Only use stdcall calling convention on 32-bit Windows

giordano:
Ref: https://github.com/JuliaParallel/MPI.jl/issues/954#issuecomment-5039547238.


giordano:
Ok, 64-bit Windows jobs (apart from those with Julia nightly, waiting for something like #941) are passing, so this should be good to go 👍

simonbyrne (approved):

PR #959, opened 2026-08-16
Project.toml: Allow `MPIABI_jll` 1.0

eschnett:


eschnett:
Only the "nightly" tests fail. Those are unreliable anyway.

We need to merge this and tag a new version to switch to `MPIABI_jll` 1.0.

giordano (approved):

PR #962, opened 2026-08-19
Do not call MPI functions directly from finalizers

giordano:
@vchuravy does this address your concerns in https://github.com/JuliaParallel/MPI.jl/pull/955#discussion_r3630535612?

However I'm not happy at all about the implementation, as it adds a new function call before each MPI call.  Benchmark:
```julia
using MPI
MPI.Init()

const comm = MPI.COMM_WORLD

function bench_rank(n)
    acc = 0
    for _ in 1:n
        acc += MPI.Comm_rank(comm)
    end
    return acc
end

function bench_get_count(n)
    # MPI_Get_count on a fabricated status: another cheap local @mpichk call
    status = MPI.Status(0, 0, 0, 0, 0)
    acc = 0
    for _ in 1:n
        acc += MPI.Get_count(status, UInt8)
    end
    return acc
end

const N = 10_000_000
bench_rank(1); bench_get_count(1)  # compile

for (name, f) in ("Comm_rank" => bench_rank, "Get_count" => bench_get_count)
    best = minimum((@elapsed f(N)) for _ in 1:5)
    println(name, ": ", round(best / N * 1e9, digits=2), " ns/call")
end

# allocations on the hot path must be zero
alloc = @allocated bench_rank(1000)
println("allocations for 1000 calls: ", alloc, " bytes")

MPI.Finalize()
```
on `master`, with 2 ranks:
```
Comm_rank: 3.88 ns/call
Comm_rank: 3.88 ns/call
Get_count: 5.56 ns/call
Get_count: 5.59 ns/call
allocations for 1000 calls: 0 bytes
allocations for 1000 calls: 0 bytes
```
on this branch:
```
Comm_rank: 5.5 ns/call
Comm_rank: 5.5 ns/call
Get_count: 15.7 ns/call
Get_count: 15.77 ns/call
allocations for 1000 calls: 0 bytes
allocations for 1000 calls: 0 bytes
```
The overhead is non-negligible (although of the order of ~nanoseconds)

> Finalizers can run at any point of the program where the garbage
> collector is invoked, including concurrently with an MPI call made by
> another thread (which is permitted only when MPI was initialized with
> `MPI_THREAD_MULTIPLE`), or in the middle of an MPI call which runs a
> Julia callback, such as a reduction with a user-defined operator
> (calling MPI functions inside such callbacks is erroneous).
> 
> Instead of calling the corresponding `MPI_*_free` function directly,
> the finalizers of MPI handle objects (communicators, groups, datatypes,
> operators, infos, requests) now capture the raw handle value in a
> closure and enqueue it, and the queue is drained before the next MPI
> call made with `@mpichk`: the fast path of the drain is a single atomic
> counter check, and enqueueing from a finalizer uses `trylock`, followed
> by re-registration of the finalizer when the lock is contended, as
> finalizers must never block on a lock.  This design is similar to how
> PythonCall.jl defers the freeing of Python objects from finalizers to
> the next time the GIL is held.
> 
> Explicit calls to `MPI.free` keep freeing the handle immediately.  The
> finalizers of `Win` and `FileHandle` objects are unchanged: freeing
> those handles is a synchronizing/collective operation, which deferral
> would not make safe.

### DIFF
--- a/src/api/api.jl
+++ b/src/api/api.jl
@@ -109,6 +109,73 @@ macro mpicall(expr)
     return esc(expr)
 end
 
+# Queue of pending `MPI_*_free` calls enqueued by finalizers of MPI handle objects.
+#
+# Finalizers must not call MPI functions directly: a finalizer can run at any point of
+# the program where the garbage collector is invoked, including concurrently with an
+# MPI call made by another thread (which is permitted only when MPI was initialized
+# with `MPI_THREAD_MULTIPLE`), or in the middle of an MPI call which runs a Julia
+# callback (calling MPI functions inside such callbacks is erroneous).  Instead,
+# finalizers capture the handle to be freed in a closure and enqueue it with
+# `defer_free!`, and the queue is drained by `drain_deferred_frees` before the next
+# MPI call made with `@mpichk`.
+const deferred_frees = Any[]
+const deferred_frees_lock = Base.Threads.SpinLock()
+# Number of enqueued closures, for a race-free fast emptiness check in
+# `drain_deferred_frees` without taking the lock.
+const deferred_frees_count = Base.Threads.Atomic{Int}(0)
+
+"""
+    API.defer_free!(f) -> Bool
+
+Enqueue the closure `f` to be called before the next MPI call made with `@mpichk`.
+Returns whether `f` was enqueued: this function is meant to be called from finalizers,
+which must never block on a lock potentially held by the interrupted task, so when the
+queue lock is contended the closure is not enqueued and the caller should re-register
+its finalizer to retry at the next garbage collection.
+"""
+function defer_free!(@nospecialize(f))
+    trylock(deferred_frees_lock) || return false
+    try
+        push!(deferred_frees, f)
+        Base.Threads.atomic_add!(deferred_frees_count, 1)
+    finally
+        unlock(deferred_frees_lock)
+    end
+    return true
+end
+
+"""
+    API.drain_deferred_frees()
+
+Run all the closures enqueued with [`API.defer_free!`](@ref).  Called before each MPI
+call made with `@mpichk`.
+"""
+function drain_deferred_frees()
+    # fast path, keep it cheap: this runs before every MPI call
+    deferred_frees_count[] == 0 && return nothing
+    # if the lock is contended, another task is already draining the queue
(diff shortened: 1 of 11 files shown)


src/api/api.jl
```diff
@@ -109,6 +109,73 @@ macro mpicall(expr)
...
+    API.defer_free!(f) -> Bool
+
+Enqueue the closure `f` to be called before the next MPI call made with `@mpichk`.
+Returns whether `f` was enqueued: this function is meant to be called from finalizers,
+which must never block on a lock potentially held by the interrupted task, so when the
+queue lock is contended the closure is not enqueued and the caller should re-register
+its finalizer to retry at the next garbage collection.
+"""
+function defer_free!(@nospecialize(f))
+    trylock(deferred_frees_lock) || return false
```
eschnett: If you are worried about performance, and if you think that the lock might be expensive, then I'm sure we could add a lock-free fast path. Per-thread queues could also help. (I understand that what I'm suggesting is not trivial, but such a fast path could speed things up.)

eschnett:
Don't we need to call the `free` functions collectively, in the same order on all processes?

This PR doesn't handle all handle types, it omits files and windows.

You could omit the non-empty-queue check for "usually cheap" MPI functions.

src/api/api.jl
```diff
@@ -109,6 +109,73 @@ macro mpicall(expr)
...
+# with `MPI_THREAD_MULTIPLE`), or in the middle of an MPI call which runs a Julia
+# callback (calling MPI functions inside such callbacks is erroneous).  Instead,
+# finalizers capture the handle to be freed in a closure and enqueue it with
+# `defer_free!`, and the queue is drained by `drain_deferred_frees` before the next
+# MPI call made with `@mpichk`.
+const deferred_frees = Any[]
+const deferred_frees_lock = Base.Threads.SpinLock()
+# Number of enqueued closures, for a race-free fast emptiness check in
+# `drain_deferred_frees` without taking the lock.
+const deferred_frees_count = Base.Threads.Atomic{Int}(0)
```
vchuravy, line 126: I think we could just defer the free to the next finalizer phase by calling `finalizer(free, obj)` since that will requeue the finalizer inside Julia and the object won't be dead yet.

PR #969, opened 2026-09-08
Root window buffers via IdSet, not Set

eschnett:
Two different arrays with the same value are considered equal by `Set`, which is not good. Use `IdSet` instead.

### DIFF
--- a/test/test_onesided.jl
+++ b/test/test_onesided.jl
@@ -128,5 +128,31 @@ MPI.Win_detach!(win, buf)
 MPI.free(win)
 MPI.free(address_win)
 
+# Attached buffers must be rooted by identity, not by value: two distinct
+# arrays with equal contents must each be kept alive on their own.
+let win = MPI.Win_create_dynamic(comm)
+    a = zeros(10)
+    b = zeros(10)                       # equal to `a`, but a different object
+    MPI.Win_attach!(win, a)
+    MPI.Win_attach!(win, b)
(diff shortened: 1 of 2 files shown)


vchuravy (approved):

test/test_onesided.jl
```diff
@@ -128,5 +128,31 @@ MPI.Win_detach!(win, buf)
 MPI.free(win)
 MPI.free(address_win)
 
+# Attached buffers must be rooted by identity, not by value: two distinct
+# arrays with equal contents must each be kept alive on their own.
+let win = MPI.Win_create_dynamic(comm)
+    a = zeros(10)
+    b = zeros(10)                       # equal to `a`, but a different object
```
giordano:
```suggestion
    b = copy(a)
```
?

eschnett: The idea is to have to different (separate) object that have the same content (values), so that `Set` treats them as the same while `IdSet` treats them as different.

I'm not sure this test is even needed.

giordano: > The idea is to have to different (separate) object that have the same content (values)

That's precisely what `copy` does, without having to explain it with a comment

giordano: > I'm not sure this test is even needed.

Does the test fail without the source code change?

eschnett: Yes it does.

giordano (approved):

PR #970, opened 2026-09-08
Dist_graph_create*: Finalize the created communicators

eschnett:


eschnett:
The tests here fail due to unrelated issues. Merging https://github.com/JuliaParallel/MPI.jl/pull/982 should help.

src/topology.jl
```diff
@@ -229,6 +231,7 @@ function Dist_graph_create_adjacent(comm::Comm, sources::Vector{Cint}, destinati
                                        length(sources), sources, source_weights,
                                        length(destinations), destinations, destination_weights,
                                        Info(infokws...), reorder, graph_comm)
+    graph_comm != COMM_NULL && finalizer(free, graph_comm)
```
vchuravy, line 234: When does `graph_comm == COMM_NULL` happen? If that is a case we need to worry about I would rather have the check in `free` since the docs now tells the user to call free.

eschnett: The standard doesn't mention that it could be `COMM_NULL`, I think it cannot. This code just follows `the usual` pattern, and that pattern isn't necessary here.

Should be we be defensive? This way we keep the check here and also add it inside the finalizer.

eschnett: Done.

Issue #921, opened 2025-11-07
`MPI.REQUEST_NULL` only pretends to be a constant


vchuravy:
```
using MPI
MPI.Init()

reqs = Vector{MPI.Request}(undef, 2)
fill!(reqs, MPI.REQUEST_NULL) 

# first foot-gun `reqs[1]` and `reqs[2]` are now aliases

MPI.Irecv!(Ref(1.0), MPI.COMM_WORLD, reqs[1])
```

```
julia> reqs
2-element Vector{MPI.Request}:
 incomplete request
 incomplete request
```

But worse...

```
julia> MPI.REQUEST_NULL
MPI.Request: incomplete request
```

PR #971, opened 2026-09-08
get_errorhandler: Finalize the error handlers

eschnett:


src/errhandler.jl
```diff
@@ -53,22 +59,31 @@ end
 
 Get the current [`Errhandler`](@ref) for the relevant MPI object.
 
+The MPI standard specifies that these behave as if a new error handler object were
+created, so the returned handler is owned by the caller: it is freed at finalization, and
+[`MPI.free`](@ref) may be called on it explicitly once it is no longer needed. If the
+implementation hands back one of the predefined handlers
+(`MPI.ERRORS_ARE_FATAL`, `MPI.ERRORS_RETURN`) then nothing is owned and nothing is freed.
```
vchuravy: This is on the overly verbose side. The user doesn't need to run `free` since you are attaching the finalizer

test/test_errorhandler.jl
```diff
@@ -24,10 +24,41 @@ end
...
+# finalization.  A predefined handler is owned by nobody and must be left alone.
+# COMM_SELF was set to the predefined ERRORS_ARE_FATAL above.  (COMM_WORLD is
+# no longer usable for this: DUMMY_COMM is MPI_COMM_WORLD and just got a custom
+# handler.)
+let errh = MPI.get_errorhandler(MPI.COMM_SELF)
+    @test errh == MPI.ERRORS_ARE_FATAL
+    @test MPI.is_predefined(errh)
+    finalize(errh)
+    @test errh == MPI.ERRORS_ARE_FATAL            # no finalizer, nothing freed
+end
```
vchuravy, line 44: Is this test useful?

eschnett: This tests checks whether the finalizer accidentally frees the built-in constant `ERRORS_ARE_FATAL`. It needs to special-case the built-in error handlers.

(This is similar to https://github.com/JuliaParallel/MPI.jl/issues/921 "MPI.REQUEST_NULL only pretends to be a constant", where we accidentally overwrite a built-in constant.)

src/errhandler.jl
eschnett: Removed.

vchuravy (approved):

PR #974, opened 2026-09-08
Reductions: New function `to_type_raw` which does not allocate

eschnett:
Inside an MPI reduction operation we have an MPI datatype. The previous code path first created a `Datatype` object from this (which allocates memory), and then retrieved its MPI datatype to get its attributes. The new function `to_type_raw` avoids the allocation.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -83,7 +83,11 @@ end
 function (w::OpWrapper{F,T})(_a::Ptr{Cvoid}, _b::Ptr{Cvoid}, _len::Ptr{Cint}, t::Ptr{MPI_Datatype}) where {F,T}
     len = unsafe_load(_len)
     if !isconcretetype(T)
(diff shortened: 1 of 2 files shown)


src/operators.jl
```diff
@@ -83,7 +83,11 @@ end
 function (w::OpWrapper{F,T})(_a::Ptr{Cvoid}, _b::Ptr{Cvoid}, _len::Ptr{Cint}, t::Ptr{MPI_Datatype}) where {F,T}
     len = unsafe_load(_len)
     if !isconcretetype(T)
```
giordano: While we're here, I find it easier to read these `if`s like:
```julia
concrete_T = if ...
    # ...
else
    # ...
fi
```

eschnett: You're thinking of Haskell.

eschnett: Done.

giordano: No, I'm thinking of Lisp

giordano (approved):

PR #975, opened 2026-09-08
MPIPreferences: Detect `mpi_abi_wrapper` library as MPIABI

eschnett:
Also correct the version number of MPICH's MPI ABI build as 1, not 5.

(It's MPI standard 5.0, but MPI ABI standard 1.0, and we care about the ABI standard here.)


giordano (commented):
Looks good, but it'd be nicer to add a test, I guess

PR #976, opened 2026-09-08
Get_count: Return nothing on failure

eschnett:
Return `nothing` instead of the (ABI-dependent) integer value `MPI_UNDEFINED` on failure.

Add explicit checks for `nothing` in the caller. (Previously there was no error handling there.)

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -240,6 +240,11 @@ import Base: @deprecate
     (flag, stat) = Iprobe(src, tag, comm)
     flag || return (false, nothing, nothing)
     count = Get_count(stat, UInt8)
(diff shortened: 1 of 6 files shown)


vchuravy:
This is technically a breaking change

eschnett:
Yes, technically it is breaking. But I want to argue that the previous state was ill-defined: Returning `MPI_UNDEFINED` returns a system-dependent constant on error.

Also, all other similar functions already return `nothing` on error. Thus I'd call it "correcting the code by adding the missing error handling branch".

eschnett:
@vchuravy is your 👍 approval for this PR?

PR #978, opened 2026-09-08
Allow user-defined operators on all architectures via a trampoline pool

eschnett:
Many architectures do not support cfunctions wrapping a closure. This makes it impossible to create user-defined reduction operators at run time.

This PR introduces 128 pre-compiled functions as a trampoline pool. This allows using closures on all architectures.

(I'm using the same approach in the
[mpif](https://github.com/eschnett/mpif) Fortran MPI bindings.)

### DIFF
--- a/docs/examples/03-reduce.jl
+++ b/docs/examples/03-reduce.jl
@@ -31,9 +31,9 @@ function pool(S1::SummaryStat, S2::SummaryStat)
     SummaryStat(m,v,n)
 end
 
-# Register the custom reduction operator.  This is necessary only on platforms
-# where Julia doesn't support closures as cfunctions (e.g. ARM), but can be used
-# on all platforms for consistency.
+# Register the custom reduction operator.  This is optional -- any function can be
+# passed directly to the reduction -- but it builds the C callback ahead of time
+# instead of at run time, which is worth doing for an operator used repeatedly.
 MPI.@RegisterOp(pool, SummaryStat)
 
 X = randn(10,3) .* [1,3,7]'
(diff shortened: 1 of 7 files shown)


vchuravy:
I am not a fan of the added complexity and I was hoping that RegisterOp covered most of the use-cases in particular since it allows for `RegisterOp(F, Any)` and then use the datatype to dispatch.

So the user-defined operator case should only be needed for true closures.

I was wondering if we could smuggle the additional information through a DataType, but I guess it would be a bit awkard.

> Smuggle context through the datatype. The callback does receive MPI_Datatype*. Create a dedicated type per op with MPI_Type_dup or MPI_Type_contiguous(1, T), then attach the context with MPI_Type_set_attr using a keyval from MPI_Type_create_keyval, or just key a global dict on the handle. In the callback, MPI_Type_get_attr recovers it. Cost: the op must be paired with its own datatype, so you cannot reuce with a bare MPI_DOUBLE.

eschnett:
From a user's perspective, having to modify datatypes is even more inconvenient than using a macro to define closures. I'd say: We either keep the current setup (requiring macros until LLVM figures things out), or we use the trampoline pool. Since I'm using a trampoline pool in other MPI-related libraries I didn't think this was too much added complexity.

If you want, we can remove the code that automatically extends the trampoline pool at run time. This would simplify things.

Of course there's also waiting for MPI 5.1 or MPI 6. They know that reduction operators have this problem.

eschnett:
The ROCm CI failure is unrelated (it's about I/O, there are no reduction in that test, and this is x86-64 anyway where the trampoline pool is disabled).

PR #979, opened 2026-09-08
Clean up string return values

eschnett:
Open MPI sometimes include a trailing NUL into string values it returns. C doesn't see this. Remove all trailing NULs explicitly.

### DIFF
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -100,8 +100,7 @@ function _string_from_buffer(buf::Vector{UInt8}, len::Integer)
     # Never grow: a `len` beyond what the library was handed would pad the
     # string with uninitialized bytes. Also tolerates an unset `resultlen`.
     resize!(buf, clamp(len, 0, length(buf)))
-    resize!(buf, something(findlast(!=(0x00), buf), 0))
-    return String(buf)
+    return chopsuffix(String(buf), "\0")
 end
 
 include("implementations.jl")
--- a/src/MPI.jl
+++ b/src/MPI.jl
@@ -86,16 +86,13 @@ function (hook::LoadTimeHookSetVal)()
 end
 call(hook) = hook()
 
-"""
-    _string_from_buffer(buf::Vector{UInt8}, len::Integer) :: String
-
-Build a `String` from the first `len` bytes of `buf`, dropping any trailing NUL
-bytes. Consumes `buf`.
-
-MPI's string-returning functions set `resultlen` to the number of characters
-written, *excluding* the terminating NUL. Open MPI's `MPI_Get_library_version`
-includes it, so trim instead of letting a NUL into the string.
-"""
+# Build a `String` from the first `len` bytes of `buf`, dropping any trailing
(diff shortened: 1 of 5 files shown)


src/MPI.jl
```diff
@@ -86,6 +86,24 @@ function (hook::LoadTimeHookSetVal)()
...
+    _string_from_buffer(buf::Vector{UInt8}, len::Integer) :: String
+
+Build a `String` from the first `len` bytes of `buf`, dropping any trailing NUL
+bytes. Consumes `buf`.
+
+MPI's string-returning functions set `resultlen` to the number of characters
+written, *excluding* the terminating NUL. Open MPI's `MPI_Get_library_version`
+includes it, so trim instead of letting a NUL into the string.
+"""
+function _string_from_buffer(buf::Vector{UInt8}, len::Integer)
```
giordano, line 96: I'm mildly sure Base has something for this.

eschnett: Apparently Base doesn't for `Vector{UInt8}`. We can convert to `String` first, then strip (which gives a `SubString`), then maybe convert to `String` again?

eschnett: We can use `chopsuffix` which would remove at most one NUL. That's probably good enough.

giordano: Doesn't
```julia
GC.@preserve buf unsafe_string(pointer(buf, len))
```
do what you want?  May still need to clamp `len` though.

eschnett: `unsafe_string` would ignore everything after the first `NUL`. I'm looking for a function that only removes a trailing `NUL`, if one is there.

Yes, clamping would be necessary anyway.

vchuravy: I am not sure I understand? When would a `NUL` occur in a valid string?

src/datatypes.jl
```diff
@@ -61,7 +61,7 @@ function get_name(datatype::Datatype)
     buffer = Array{UInt8}(undef, API.MPI_MAX_OBJECT_NAME)
     lenref = Ref{Cint}()
     API.MPI_Type_get_name(datatype, buffer, lenref)
-    return String(resize!(buffer, lenref[]))
+    return _string_from_buffer(buffer, lenref[])
```
vchuravy, line 64: This looks like an upstream bug.

eschnett: It does, doesn't it? But Open MPI is so widely used that we should introduce a work-around, especially since all C/C++ users will never have notices this bug.

src/MPI.jl
eschnett: `unsafe_string` without a length argument would be unsafe because the buffer might not be NUL terminated.

`unsafe_string` with a length argument doesn't stop at NUL, it uses exactly that many bytes.

I think the current approach is good.

src/datatypes.jl
eschnett: I filed an upstream report at https://github.com/open-mpi/ompi/issues/14411.

PR #982, opened 2026-09-08
test_io_shared: Replace sync/skip workarounds with barriers

### LINKED ISSUE #555: Spurios test failures on Windows for `get_position_shared`

### LINKED ISSUE #879: [CI] Intermittent failure in `test_io_shared` test on macOS
On macOS, with any MPI library, we occasionally see a failure in `test_io_shared` ([example](https://github.com/JuliaParallel/MPI.jl/actions/runs/11013552982/job/30582168543#step:7:353)):
```
Test Failed at /Users/runner/work/MPI.jl/MPI.jl/test/test_io_shared.jl:49
  Expression: MPI.File.get_position_shared(fh) == sum(1:sz)
   Evaluated: 0 == 10
```

eschnett:
test_io_shared: Replace sync/skip workarounds with the barriers actually needed

The flakiness (#555 on Windows, #879 on macOS, and the Ubuntu failures seen in #926) was a data race in the test itself, not a library bug.

MPI_File_get_position_shared is not collective, but every other shared file pointer call in this test moves the pointer: write_ordered and read_ordered increment it, seek_shared and set_view reset it. ROMIO and OMPIO both perform the reset on rank 0 (ROMIO's set_view: on every rank that has the hidden .shfp file open) *before* their internal barrier, and write_ordered increments the counter on rank 0 as soon as rank 0 enters the call. So a fast rank entering the next operation clobbers the pointer while a slow rank is still inside `@test get_position_shared(fh) == ...`. That is exactly the `0 == 10` / `0 == 3` reported in the issues.

The fix is a plain MPI.Barrier after each block of position checks and before the next shared-pointer operation. MPI_File_sync only flushes file data (ROMIO keeps the pointer in a side file, OMPIO in shared memory) and plays no role; the old sync/Barrier/sync wrapper only helped through its embedded Barrier. With atomicity enabled the header data needs no sync either, only ordering.

The vendor detection and the Apple/MPICH/Windows `skip=` clauses were masking the same race and are removed. Reproduced locally on macOS with MPICH 4.3.1, MPICH_jll 5.x, OpenMPI_jll 5.0.10 and MPItrampoline_jll: 30-60% failure rate without the barriers, 0 failures in 80 runs with them.

Fixes #555, fixes #879.

(Thumbs pressed that this works!)

### DIFF
--- a/test/test_io.jl
+++ b/test/test_io.jl
@@ -15,12 +15,28 @@ fh = MPI.File.open(comm, filename, read=true, write=true, create=true)
 MPI.File.set_view!(fh, 0, MPI.Datatype(Int64), MPI.Datatype(Int64))
 MPI.File.write_at_all(fh, rank*2, ArrayType([Int64(rank+1) for i = 1:2]))
 
-# The file is in the default non-atomic mode, in which data written by one
-# process becomes visible to another only after the writer has called
-# `File.sync`, the two have synchronized, and the reader has called `File.sync`
-# as well (MPI standard, "File Consistency").  `File.sync` is collective, so all
-# ranks call it on both sides of the barrier.  Without the barrier, rank 0 can
-# read before the other ranks' writes have landed.
+# Synchronizing parallel MPI writes/reads requires a bit of a dance.
+#
+# The file is in the default non-atomic mode, in which data written by
+# one process becomes visible to other processes only after *first*
+# the writer has called `File.sync`, and *then* (afterwards) the
+# reader has called `File.sync`. To ensure the "first" ... "then" bit,
+# i.e. to ensure that the writer's `File.sync` happens before the
+# reader's `File.sync`, the two processes need to use e.g. an MPI
+# barrier (MPI standard, "File Consistency").
+#
+# `File.sync` is a collective operation, so all ranks must call it
+# whenever it is called, in this case on both sides of the barrier.
+# Without the barrier, one rank could read before the other ranks'
+# writes have finished.
+#
+# Terminology isn't really helpful here. There is `File.sync`, which
+# synchronizes a process's view of the file with the actual content of
+# the file, and there is an MPI `Barrier`, which synchronizes the MPI
+# processes with each other. That's two different kinds of
+# synchronization. Tongue-in-cheek: We need to barrier-synchronize
+# between the two file-synchronizations.
+
 MPI.File.sync(fh)
 MPI.Barrier(comm)
 MPI.File.sync(fh)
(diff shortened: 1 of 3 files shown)


vchuravy:
Still seems to have some issues on Windows?

eschnett:
I hope https://github.com/JuliaParallel/MPI.jl/pull/984 will help on Windows.

test/test_io.jl
```diff
@@ -15,6 +15,14 @@ fh = MPI.File.open(comm, filename, read=true, write=true, create=true)
 MPI.File.set_view!(fh, 0, MPI.Datatype(Int64), MPI.Datatype(Int64))
 MPI.File.write_at_all(fh, rank*2, ArrayType([Int64(rank+1) for i = 1:2]))
 
+# The file is in the default non-atomic mode, in which data written by one
+# process becomes visible to another only after the writer has called
+# `File.sync`, the two have synchronized, and the reader has called `File.sync`
+# as well (MPI standard, "File Consistency").  `File.sync` is collective, so all
+# ranks call it on both sides of the barrier.  Without the barrier, rank 0 can
+# read before the other ranks' writes have landed.
```
vchuravy: Could we have some docs on how to use this part of IO?

Previously, we had:

```
# Syncing parallel MPI I/O is a bit involved:
function sync(comm, fh)
    # First ensure that all local changes are flushed ...
    MPI.File.sync(fh)
    # ... then wait for all other process to finish doing that ...
    MPI.Barrier(comm)
    # ... then make sure we see all changes that the other processes made.
    MPI.File.sync(fh)
end
```

Which I find more helpful than:

> `File.sync` is collective, so all ranks call it on both sides of the barrier.

File.sync being collective doesn't help me much, it confuses me more (why do I need the barrier when I just called a collective).

I had to go read `Example 14.6` from the MPI standard to make sense of this again.

eschnett: Note that the "helpful" comments in my function were also not quite correct. Not all of the `File.sync` previous calls are needed.

I have updated the comment, first rewriting it into English, and then adding an explanation for the confusing terminology.

vchuravy (approved):

PR #985, opened 2026-09-10
Call MPI_Abort when a test process fails

eschnett:
Failing self-tests exit via an exception. Depending on the MPI implementation, this may lead to a hang, which leads to a time-out in Github CI, which leads to a `cancelled` classification instead of an error.

This PR calls `MPI_Abort` when a test fails by adding a respective `atexit` hook.

This is done only for the self-tests. It might be useful to do the same for all MPI applications – it doesn't really make sense to just exit the Unix process, we should either call `MPI_Finalize` or `MPI_Abort` in either case, and if we exit without the app calling `MPI_Finalize` the we should probably abort. But that's a different issue.

### DIFF
--- a/docs/src/external.md
+++ b/docs/src/external.md
@@ -23,7 +23,7 @@ ccall((:cfunc2, lib), Cint, (Ptr{MPI.MPI_Comm},), comm)
 
 ## Object finalizers and `MPI.Finalize`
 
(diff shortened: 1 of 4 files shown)


eschnett:
The failing test is probably cured by https://github.com/JuliaParallel/MPI.jl/pull/982.

PR #987, opened 2026-09-18
Add MPI.Win_allocate

jack-dunham:
Wraps `MPI_Win_allocate`, the one window constructor in the standard that MPI.jl did not yet expose. `API.MPI_Win_allocate` is already generated in `src/api/generated_api.jl`; this adds the high-level interface beside its siblings in `src/onesided.jl`.

Full disclosure, I am not an expert in MPI, and this code was generated by Claude by pointing it to the existing `MPI.Win_allocate_shared` implementation, but the output seems reasonable to me. If it is not, consider this a feature request with a first attempt at a solution.


src/onesided.jl
```diff
@@ -69,6 +69,51 @@ function Win_create(base::SubArray{T}, comm::Comm; infokws...) where T
...
+
+Unlike [`MPI.Win_create`](@ref), the window memory is allocated by MPI rather than
+by the caller, which allows the implementation to choose a backing store suited to
+one-sided access. The memory belongs to MPI, so `array` must not be accessed after
+[`MPI.free`](@ref) has been called on `win`.
+
+`disp_unit` is `sizeof(T)`, so the `disp` argument of [`MPI.Get!`](@ref) and
+[`MPI.Put!`](@ref) counts elements.
+
+`infokws` are info keys providing optimization hints.
```
vchuravy, line 90: We would still just call them `kwargs...`

src/onesided.jl
```diff
@@ -69,6 +69,51 @@ function Win_create(base::SubArray{T}, comm::Comm; infokws...) where T
...
+    API.MPI_Win_allocate(len*sizeof(T), sizeof(T), Info(infokws...), comm, out_baseptr, win)
+    finalizer(free, win)
+    return win, out_baseptr[]
+end
+function Win_allocate(::Type{Array{T}}, dims, comm::Comm; infokws...) where T
+    len = prod(dims)
+    win, ptr = Win_allocate(Ptr{T}, len, comm; infokws...)
+    # the ptr may be invalid for a zero-size window, which will cause an error as
+    # unsafe_wrap checks the alignment of ptr, even for length 0
+    array = len > 0 ? unsafe_wrap(Array, ptr, dims) : Array{T}(undef, dims)
```
vchuravy, line 112: That error kind would also be true for Win_allocate_shared

src/onesided.jl
```diff
@@ -69,6 +69,51 @@ function Win_create(base::SubArray{T}, comm::Comm; infokws...) where T
...
+
+`infokws` are info keys providing optimization hints.
+
+[`MPI.free`](@ref) should be called on the `Win` object once operations have
+been completed.
+
+# External links
+$(_doc_external("MPI_Win_allocate"))
+"""
+function Win_allocate(::Type{Ptr{T}}, len::Integer, comm::Comm; infokws...) where T
```
vchuravy, line 98: We should check if `T` is pointerfree and inlinealloc

jack-dunham: I.e. `isbitstype(T)` ?

src/onesided.jl
jack-dunham: There are some existing inconsistencies in this file regarding this. Happy to update all instances to `kwargs...`.

